PR #9, opened 2016-01-21
Fixing CCache and Orca Ref Count protocol

armenatzoglou:
Removing AddRef when orca performs a lookup to CCache and retrieves an object that exists in CCache already.
According to the new protocol, in the case described above, CCache is responsible to increase the ref count and orca simply releases the object after the use.


armenatzoglou:
Pushed. Thanks a lot for the review.

PR #10, opened 2016-01-22
Raghav missing col

vraghavan78:
**I** 

In GPDB/Postgres/HAWQ a column in a particular table is reference by attribute number which is an integer.

1-n (positive integers) represent table columns. This can include dropped columns
0 represents the whole table for instance select foo.* from foo
-1 to -n (negative integers) system columns like segment id, tuple id, etc.

Now, inside orca all columns have positive number. The **relation metadata object** has dropped column information and inside the query we have **table descriptors** that does not have dropped columns.

So if we do not map the columns from the **table descriptors** to **relation metadata object** we will have incorrect mapping causing issues when try to compute stats or when we record missing stats. This fix addresses that issue. 

**II** 
Before this code review the algorithm for fetch column stats and recording columns with no stats we did it in three passes. Now it is a single pass algorithm.

**III** 

Added tests for missing stats recording.

@oarap @hsyuan and @xinzweb please take a look.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ BLD_TOP := $(shell sh -c pwd)
 include make/gpo.mk
 
 LIB_NAME = optimizer
-LIB_VERSION = 1.619
+LIB_VERSION = 1.620
 ## ----------------------------------------------------------------------
 ## top level variables; only used in this makefile
 ## ----------------------------------------------------------------------
--- a/libgpopt/include/gpopt/mdcache/CMDAccessor.h
+++ b/libgpopt/include/gpopt/mdcache/CMDAccessor.h
@@ -28,12 +28,14 @@
 
 #include "gpopt/spinlock.h"
 #include "gpopt/mdcache/CMDKey.h"
+#include "gpopt/engine/CStatisticsConfig.h"
 
 #include "naucrates/md/IMDId.h"
 #include "naucrates/md/IMDProvider.h"
 #include "naucrates/md/IMDType.h"
 #include "naucrates/md/IMDFunction.h"
 #include "naucrates/md/CSystemId.h"
+#include "naucrates/statistics/IStatistics.h"
 
 // fwd declarations
 namespace gpdxl
@@ -249,6 +251,23 @@ namespace gpopt
 			// initialize hash tables
 			void InitHashtables(IMemoryPool *pmp);
 
+			// return the column statistics meta data object for a given column of a table
+			const IMDColStats *Pmdcolstats(IMemoryPool *pmp, IMDId *pmdidRel, ULONG ulPos);
+
+			// record histogram and width information for a given column of a table
+			void RecordColumnStats
+					(
+					IMemoryPool *pmp,
+					IMDId *pmdidRel,
+					ULONG ulColId,
+					ULONG ulPos,
+					BOOL fSystemCol,
+					BOOL fEmptyTable,
+					HMUlHist *phmulhist,
+					HMUlDouble *phmuldoubleWidth,
+					CStatisticsConfig *pstatsconf
+					);
+
 			// construct a stats histogram from an MD column stats object  
 			CHistogram *Phist(IMemoryPool *pmp, IMDId *pmdidType, const IMDColStats *pmdcolstats);
 
@@ -341,10 +360,9 @@ namespace gpopt
 				(
 				IMemoryPool *pmp, 
 				IMDId *pmdidRel,
-				DrgPul *pdrgpulHistPos,	// array of attribute positions in the base relation for detailed stats
-				DrgPul *pdrgpulHistColIds,	// array of column ids of those attributes in the query
-				DrgPul *pdrgpulWidthPos,	// array of attribute positions in the base relation for widths
-				DrgPul *pdrgpulWidthColIds	// array of column ids of those attributes in the query
+				CColRefSet *pcrsHist,  // set of column references for which stats are needed
+				CColRefSet *pcrsWidth, // set of column references for which the widths are needed
+				CStatisticsConfig *pstatsconf = NULL
 				);
 			
 			// calculate space necessary for serializing sysids
--- a/libgpopt/src/mdcache/CMDAccessor.cpp
+++ b/libgpopt/src/mdcache/CMDAccessor.cpp
@@ -24,9 +24,12 @@
 #include "gpos/task/CAutoSuspendAbort.h"
 #include "gpos/error/CAutoTrace.h"
 
+#include "gpopt/base/CColRefSetIter.h"
+#include "gpopt/base/CColRefTable.h"
 #include "gpopt/exception.h"
 #include "gpopt/mdcache/CMDAccessor.h"
 #include "gpopt/mdcache/CMDAccessorUtils.h"
+
 
 #include "naucrates/exception.h"
 #include "naucrates/traceflags/traceflags.h"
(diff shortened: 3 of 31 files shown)


Makefile
```diff
@@ -10,7 +10,7 @@ BLD_TOP := $(shell sh -c pwd)
 include make/gpo.mk
 
 LIB_NAME = optimizer
-LIB_VERSION = 1.618
+LIB_VERSION = 1.619.2
```
xinzweb: You will remove the tailing .2 before final commit, right?

libgpopt/include/gpopt/mdcache/CMDAccessor.h
```diff
@@ -28,10 +28,12 @@
 
 #include "gpopt/spinlock.h"
 #include "gpopt/mdcache/CMDKey.h"
+#include "gpopt/engine/CStatisticsConfig.h"
```
xinzweb: Do we order the "include"?

Makefile
vraghavan78: the .2 was for my internal testing. Publishing the library and running ICG on GPDB to ensure that all the tests pass.

libgpopt/include/gpopt/mdcache/CMDAccessor.h
vraghavan78: will do.

libgpopt/src/mdcache/CMDAccessor.cpp
```diff
@@ -1101,76 +1183,174 @@ CMDAccessor::Pstats
 	(
 	IMemoryPool *pmp,
 	IMDId *pmdidRel,
-	DrgPul *pdrgpulHistPos,
-	DrgPul *pdrgpulHistColIds,
-	DrgPul *pdrgpulWidthPos,
-	DrgPul *pdrgpulWidthColIds
+	CColRefSet *pcrsHist,
+	CColRefSet *pcrsWidth,
```
vraghavan78: assert that the two sets are disjoint

server/src/unittest/gpopt/mdcache/CMDAccessorTest.cpp
```diff
@@ -391,6 +393,113 @@ CMDAccessorTest::EresUnittest_Navigate()
...
+
+	SMissingStatsTestCase rgTestCases[] =
+	{
+		// table has one dropped column and we are requesting columns that have stats
+		{GPOPT_MDCACHE_TEST_OID, 1 /* major */, 13 /* minor version */, 1,2,1,3,0},
+
+		// table has one dropped column and one of the requested column is the dropped column
+		{GPOPT_MDCACHE_TEST_OID, 1 /* major */, 13 /* minor version */, 1,2,1,2,1},
+
+		// table has no dropped column and we are requesting columns that have stats
```
vraghavan78: no stats. wrong comment.

server/include/unittest/gpopt/mdcache/CMDAccessorTest.h
```diff
@@ -70,6 +70,30 @@ namespace gpopt
...
+				// id from GPDB system catalog
+				OID m_oid;
+
+				// major version number
+				ULONG m_ulVersionMajor;
+
+				// minor version number
+				ULONG m_ulVersionMinor;
+
+				// columns for which we want to extract stats object
```
xinzweb: If the intend of this test case is to check # of missing statistics matching what we expected, then maybe I'd like to recommend to rewrite the interface a little bit?

Look at the test case below, these two fields actually the first # of columns you want to extract the stats object, right? In your example below, it always 2. (means it start from 1 to 2).

Then later you have an 'array' of the AttNo, maybe just call the 'selected_att_no_array'.

the end very end, it's the 'm_ulMissingStats' actually the 'expected_number_of_missing_stats'.

This would be better than 'Col1, Col2, Attno1, and Attno2', make sense?

libgpopt/include/gpopt/mdcache/CMDAccessor.h
```diff
@@ -249,6 +251,22 @@ namespace gpopt
...
+
+			// record histogram and width information for a given column of a table
+			void RecordColumnStats
+					(
+					IMemoryPool *pmp,
+					IMDId *pmdidRel,
+					ULONG ulColId,
+					ULONG ulPos,
+					BOOL fEmptyTable,
+					HMUlHist *phmulhist,
```
xinzweb: Yeah, if we can have the Hist and Width fit into same data structure, then we can reduce one more input parameter here.

libgpopt/include/gpopt/mdcache/CMDAccessor.h
```diff
@@ -341,10 +359,21 @@ namespace gpopt
 				(
 				IMemoryPool *pmp, 
 				IMDId *pmdidRel,
-				DrgPul *pdrgpulHistPos,	// array of attribute positions in the base relation for detailed stats
-				DrgPul *pdrgpulHistColIds,	// array of column ids of those attributes in the query
-				DrgPul *pdrgpulWidthPos,	// array of attribute positions in the base relation for widths
-				DrgPul *pdrgpulWidthColIds	// array of column ids of those attributes in the query
+				CColRefSet *pcrsHist,  // set of column references for which stats are needed
```
xinzweb: A clarification question. 

What's the relationship between the set of columns have histograms vs. the set of columns have width?

I assume the columns with width would be a SUPER set of columns of histogram?

vraghavan78: Yes. I agree we should remove it. I have created this as a story in the OSS backlog. #112542551

libgpopt/src/mdcache/CMDAccessor.cpp
```diff
@@ -1101,76 +1183,174 @@ CMDAccessor::Pstats
...
+//---------------------------------------------------------------------------
+//	@function:
+//		CMDAccessor::Pstats
+//
+//	@doc:
+//		Construct a statistics object for the columns of the given relation
+//
+//---------------------------------------------------------------------------
+IStatistics *
+CMDAccessor::Pstats
```
xinzweb: Sorry, do we still need this function? I though the above function is doing the Pstats. Is this one for the old interface?

libnaucrates/include/naucrates/md/CMDRelationCtasGPDB.h
```diff
@@ -90,6 +90,10 @@ namespace gpmd
 			// number of system columns
 			ULONG m_ulSystemColumns;
 			
+			// mapping of attribute number in the system catalog to the positions of
+			// the non dropped column in the metadata object
+			HMIUl *m_phmiulAttno2Pos;
+
```
xinzweb: Yeah, sorry to nag on it, but if we can do the refactoring in this shot, that would be wonderful.

libnaucrates/include/naucrates/statistics/IStatistics.h
```diff
@@ -41,12 +44,30 @@ namespace gpnaucrates
...
 
+	// hash map from column id to a histogram
+	typedef CHashMap<ULONG, CHistogram, gpos::UlHash<ULONG>, gpos::FEqual<ULONG>,
+					CleanupDelete<ULONG>, CleanupDelete<CHistogram> > HMUlHist;
+
+	// iterator
+	typedef CHashMapIter<ULONG, CHistogram, gpos::UlHash<ULONG>, gpos::FEqual<ULONG>,
+					CleanupDelete<ULONG>, CleanupDelete<CHistogram> > HMIterUlHist;
+
+	// hash map from column id to a CDouble for width
```
xinzweb: Try to comment on the intent, since the code already told the data type. Something like:

// hash map from column id to column width

libnaucrates/include/naucrates/statistics/IStatistics.h
```diff
@@ -41,12 +44,30 @@ namespace gpnaucrates
...
 	// fwd declarations
-	class CHistogram;
 	class IStatistics;
 
+	// hash map from column id to a histogram
+	typedef CHashMap<ULONG, CHistogram, gpos::UlHash<ULONG>, gpos::FEqual<ULONG>,
+					CleanupDelete<ULONG>, CleanupDelete<CHistogram> > HMUlHist;
+
+	// iterator
+	typedef CHashMapIter<ULONG, CHistogram, gpos::UlHash<ULONG>, gpos::FEqual<ULONG>,
```
xinzweb: Sorry, for short of knowledge, is that possible for the CHashMap return an Iterator of type CHashMapIter? In that case, we don't have to always define those two in pairs?

libnaucrates/include/naucrates/statistics/IStatistics.h
```diff
@@ -41,12 +44,30 @@ namespace gpnaucrates
...
+					CleanupDelete<ULONG>, CleanupDelete<CDouble> > HMUlDouble;
+
+	// iterator
+	typedef CHashMapIter<ULONG, CDouble, gpos::UlHash<ULONG>, gpos::FEqual<ULONG>,
+					CleanupDelete<ULONG>, CleanupDelete<CDouble> > HMIterUlDouble;
+
 	typedef CHashMap<ULONG, ULONG, gpos::UlHash<ULONG>, gpos::FEqual<ULONG>,
 					CleanupDelete<ULONG>, CleanupDelete<ULONG> > HMUlUl;
 
+	// hash maps mapping INT -> ULONG
```
xinzweb: The intend of this map is to convert the 'position' to 'attno', right? Maybe better comment on the intend here?

libnaucrates/src/md/CMDRelationCtasGPDB.cpp
```diff
@@ -90,6 +91,12 @@ CMDRelationCtasGPDB::CMDRelationCtasGPDB
 		{
 			m_pdrgpulNonDroppedCols->Append(GPOS_NEW(m_pmp) ULONG(ul));
 		}		
+
+		(void) m_phmiulAttno2Pos->FInsert
+									(
```
xinzweb: The indent looks very wrong on github. Please double check.

libnaucrates/src/md/CMDRelationCtasGPDB.cpp
```diff
@@ -203,6 +211,26 @@ CMDRelationCtasGPDB::UlSystemColumns() const
...
+//---------------------------------------------------------------------------
+//	@function:
+//		CMDRelationCtasGPDB::UlPosFromAttno
+//
+//	@doc:
+//		Return the position of a column in the metadata object given the
+//		attribute number in the system catalog
+//---------------------------------------------------------------------------
+ULONG
+CMDRelationCtasGPDB::UlPosFromAttno
```
xinzweb: Is that possible to move this function to IMDRelation.h, which is shared by IMDRelationCtasGPDB, IMDRelationExternalGPDB and IMDRelationGPDB?

server/include/unittest/gpopt/mdcache/CMDAccessorTest.h
```diff
@@ -70,6 +70,30 @@ namespace gpopt
...
+				ULONG m_ulCol1;
```
xinzweb: Is that possible to have test case support more than 2 hard-coded columns? 

From your test case example, seems like you only need a number here, since you always do "1,2". Maybe using 2 for "1,2", and "3" for "1, 2, 3"? 

This is more like 'total number of columns'.

server/include/unittest/gpopt/mdcache/CMDAccessorTest.h
```diff
@@ -70,6 +70,30 @@ namespace gpopt
...
+				ULONG m_ulCol2;
+
+				// attribute number (in the system catalog)
+				INT m_iAttno1;
```
xinzweb: Then, these are the columns mapped to the AttNo. Maybe having an array here?

libnaucrates/src/md/CMDRelationCtasGPDB.cpp
vraghavan78: IMDRelation has pure virtual function so I do not want it to have it there. I have created another story for the refactoring

server/include/unittest/gpopt/mdcache/CMDAccessorTest.h
```diff
@@ -70,6 +70,30 @@ namespace gpopt
...
+				INT m_iAttno2;
+
+				// number of columns for which the stats are missing
+				ULONG m_ulMissingStats;
```
xinzweb: This is the expected # of columns missing statistics, right? Maybe comment it out exactly, so people may know what is this test looking for, like:

// expected number of columns with missing stats.

server/include/unittest/gpopt/mdcache/CMDAccessorTest.h
vraghavan78: Not sure what that will add? Can we not cover it with just two columns?

libnaucrates/include/naucrates/statistics/IStatistics.h
vraghavan78: good catch.

libnaucrates/include/naucrates/statistics/IStatistics.h
vraghavan78: That is a change in GPOS. Please write a story for that and we can prioritize it.

libnaucrates/include/naucrates/statistics/IStatistics.h
vraghavan78: you want me to remove CDouble. Sure!

libgpopt/src/mdcache/CMDAccessor.cpp
vraghavan78: Yes, because it is used in the test. CMDAccessor does not generate column references.

libgpopt/include/gpopt/mdcache/CMDAccessor.h
vraghavan78: +1

server/include/unittest/gpopt/mdcache/CMDAccessorTest.h
vraghavan78: sure

libgpopt/src/operators/CLogical.cpp
```diff
@@ -1196,38 +1198,41 @@ CLogical::PstatsBaseTable
 	IMemoryPool *pmp,
 	CExpressionHandle &exprhdl,
 	CTableDescriptor *ptabdesc,
-	CColRefSet *pcrsStatExtra   // additional columns required for stats, not required by parent
+	CColRefSet *pcrsHistExtra   // additional columns required for stats, not required by parent
```
vraghavan78: I think so.

server/include/unittest/gpopt/mdcache/CMDAccessorTest.h
vraghavan78: paired with @xinzweb and refactored the test

PR #19, opened 2016-03-04
Enable ccache to speedup cmake

xinzweb:
This reduce a 4min clean build to less than 20 seconds (after `make clean`)

```
real	0m14.609s
user	0m17.162s
sys	0m20.762s
```


vraghavan78:
+1

PR #20, opened 2016-03-04
Fix ccache find problem

hsyuan:
This fixed the `Unable to dispatch recipe` build error when finding `ccache` program.


hsyuan:
It turned out to be test infra issue.

PR #33, opened 2016-04-09
Enable unit to take into account number of segments in minidump 

vraghavan78:
Tests with minidumps will now that in the segments mentioned in the minidump for costing. If not specified it will take 2 as default.

@xinzweb @hsyuan @d @oarap please take a look.

[#117109479]


hsyuan:
LGTM.

xinzweb:
+1 Great!

PR #34, opened 2016-04-11
Minidump: Disable TF 105000 in the to avoid ConstantEvaluation [#116803217]

vraghavan78:
In the pull request, I have disabled the constant expression evaluation when there is none provided in the particular unit test. 

There is two tests where the predicate seems to have swapped sides. Not sure if these are non-deterministic tests.

@xinzweb @d @oarap @hsyuan please take a look.

### DIFF
--- a/libgpopt/src/minidump/CMinidumperUtils.cpp
+++ b/libgpopt/src/minidump/CMinidumperUtils.cpp
@@ -421,8 +421,8 @@ CMinidumperUtils::PdxlnExecuteMinidump
 	BOOL fConstantExpressionEvaluator = true;
 	if (NULL == pceeval)
 	{
-		// disable constant expression evaluation when running minidump since we
-		// may now have an executor to compute the scalar expression
+		// disable constant expression evaluation when running minidump since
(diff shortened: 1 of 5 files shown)


hsyuan:
LGTM

oarap:
+1

libgpopt/src/minidump/CMinidumperUtils.cpp
```diff
@@ -418,6 +418,16 @@ CMinidumperUtils::PdxlnExecuteMinidump
 	CBitSet *pbsDisabled = NULL;
 	SetTraceflags(pmp, pdxlmd->Pbs(), &pbsEnabled, &pbsDisabled);
 
+	BOOL fConstantExpressionEvaluator = true;
```
xinzweb: Seems like this code got duplicated.

Maybe we can take this chance to refactor this function by reusing the second function with the new MDAccessor?

vraghavan78: its 3 lines, and I did not want add more functions for that :(

PR #36, opened 2016-04-12
Using INT_MAX instead of ULONG_MAX to avoid Xerces issue

xinzweb:
@vraghavan78 @oarap @d

### DIFF
--- a/libgpopt/include/gpopt/engine/CHint.h
+++ b/libgpopt/include/gpopt/engine/CHint.h
@@ -69,7 +69,7 @@ namespace gpopt
 			static
 			CHint *PhintDefault(IMemoryPool *pmp)
 			{
-				return GPOS_NEW(pmp) CHint(ULONG_MAX /* ulMinNumOfPartsToRequireSortOnInsert */);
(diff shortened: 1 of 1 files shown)


hsyuan:
+1

vraghavan78:
+1

danielgustafsson:
Out of curiosity, what was the Xerces issue?

xinzweb:
Xerces `XMLString` class has a `parseInt()` function, which only returned a signed int. However, in GPORCA code, we use it to return a ULONG. In this case, we set default value of ULONG_MAX out of range of INT. Hence, this fix is to lower the default value to INT_MAX instead of ULONG_MAX. Thanks.

danielgustafsson:
> On 12 Apr 2016, at 20:11, Xin Zhang <notifications@github.com> wrote:
> 
> Xerces XMLString class has a parseInt() function, which only returned a signed int. However, in GPORCA code, we use it to return a ULONG. In this case, we set default value of ULONG_MAX out of range of INT. Hence, this fix is to lower the default value to INT_MAX instead of ULONG_MAX. Thanks.

Thanks a lot for clarifying!

PR #37, opened 2016-04-12
Fix CMinidumperUtils to properly disable 105000 [#116803217]

xinzweb:
Reset metadata cache between tests
Release memory to avoid memory leak


oarap:
+1

PR #39, opened 2016-04-13
Fix the memory leak and removing tpc* tests [#116803217]

xinzweb:
Worked with @vraghavan78 on fix the memory leak issue.

Remove long running TPCH and TPCDS tests from GPORCA unit tests.


libgpopt/src/minidump/CMinidumperUtils.cpp
```diff
@@ -392,8 +392,11 @@ CMinidumperUtils::PdxlnExecuteMinidump
 	CMDProviderMemory *pmdp = GPOS_NEW(pmp) CMDProviderMemory(pmp, szFileName);
 	const DrgPsysid *pdrgpsysid = pdxlmd->Pdrgpsysid();
 	DrgPmdp *pdrgpmdp = GPOS_NEW(pmp) DrgPmdp(pmp);
-	pdrgpmdp->Append(pmdp);
 	
+	// ensure there is at least ONE system id
+	pmdp->AddRef();
```
d: Just ran your code from #37 and hit this exact problem. I'm actually really mad that we are running our tests _not_ under Debug. I've talked to @vraghavan78 about this way back, but let's talk tomorrow at stand up

vraghavan78: I always run the test in RelWithDebugInfo and debug before I submit things.
Now I do not want GPDB to run with debug since it will be slow to run icg
with it.

V

xinzweb:
Yeah Jesse, your instinct was right, and also as you said, we should run both debug and retail. Also, maybe we need to reorganize the concourse pipeline to group centos5 debug and retail together, instead of split the debug into `secondary` group.

PR #40, opened 2016-04-13
Clean GPORCA Concourse Pipeline [#117366937]

hsyuan:
Add new group 'gpdb' to filter pipeline to 'gpdb_icg' related.
Reduce max_in_flight from 2 to 1 to speed-up turn around time.


vraghavan78:
+1

PR #43, opened 2016-04-20
Fix issue with volatile function doesn't work correctly inside the 'WITH' clause

oarap:
Fix predicate propagation into CTE Producer for volatile functions

When predicate has volatile function like random(), the predicate shouldn't be pushed inside CTE producer, since it will change the query semantic.

### DIFF
--- a/libgpopt/src/operators/CPredicateUtils.cpp
+++ b/libgpopt/src/operators/CPredicateUtils.cpp
@@ -3104,6 +3104,74 @@ CPredicateUtils::FCompatibleIndexPredicate
 	return (pmdindex->FCompatible(pmdobjScCmp, ulKeyPos));
 }
 
+//---------------------------------------------------------------------------
+//	@function:
+//		CPredicateUtils::FContainsVolatileFunction
+//
+//	@doc:
+//		Check if given array of expressions contain a volatile function like random().
+//---------------------------------------------------------------------------
+BOOL
+CPredicateUtils::FContainsVolatileFunction
+	(
+	DrgPexpr *pdrgpexprPred
+	)
+{
+	GPOS_ASSERT(NULL != pdrgpexprPred);
+
+	const ULONG ulNumPreds = pdrgpexprPred->UlLength();
+	for (ULONG ul = 0; ul < ulNumPreds; ul++)
+	{
+		CExpression *pexpr = (CExpression *)(*pdrgpexprPred)[ul];
+
+		if (FContainsVolatileFunction(pexpr))
+		{
+			return true;
+		}
+	}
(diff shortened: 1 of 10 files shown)


libgpopt/src/operators/CPredicateUtils.cpp
```diff
@@ -3104,6 +3104,76 @@ CPredicateUtils::FCompatibleIndexPredicate
...
+//		Check if given array of expressions contain a volatile function like random().
+//---------------------------------------------------------------------------
+BOOL
+CPredicateUtils::FContainsVolatileFunction
+	(
+	DrgPexpr *pdrgpexprPred
+	)
+{
+	GPOS_ASSERT(NULL!=pdrgpexprPred);
+
```
vraghavan78: NULL != pdrg...

libgpopt/src/operators/CPredicateUtils.cpp
```diff
@@ -3104,6 +3104,76 @@ CPredicateUtils::FCompatibleIndexPredicate
...
+	{
+		CExpression *pexpr = (CExpression *)(*pdrgpexprPred)[ul];
+
+		if (FContainsVolatileFunction(pexpr))
+		{
+			return true;
+		}
+	}
+
+	// cannot find any, return false
```
vraghavan78: any what? no need for comment I think.

libgpopt/src/operators/CPredicateUtils.cpp
```diff
@@ -3104,6 +3104,76 @@ CPredicateUtils::FCompatibleIndexPredicate
...
+
+
+	COperator *popCurrent = pexpr->Pop();
+	GPOS_ASSERT(NULL!=popCurrent);
+
+	if (COperator::EopScalarFunc == popCurrent->Eopid())
+	{
+		CScalarFunc *pCurrentFunction = CScalarFunc::PopConvert(popCurrent);
+		BOOL isVolatile = (IMDFunction::EfsVolatile == pCurrentFunction->EfsGetFunctionStability());
+		return isVolatile;
```
vraghavan78: return (IMDFunction::EfsVolatile == pCurrentFunction->EfsGetFunctionStability());

libgpopt/src/operators/CPredicateUtils.cpp
```diff
@@ -3104,6 +3104,76 @@ CPredicateUtils::FCompatibleIndexPredicate
...
+	}
+
+	// recursively checking children
```
vraghavan78: check

vraghavan78:
nice work!

PR #45, opened 2016-04-28
Aligning nullable properties after dropping columns 117875199

oarap:
This pull request resolves the misalignment of `nullable` column properties.  It aligns `nullable` property of columns with the metadata of the relation and table descriptor after dropped columns.

### DIFF
--- a/libgpopt/include/gpopt/minidump/CMetadataAccessorFactory.h
+++ b/libgpopt/include/gpopt/minidump/CMetadataAccessorFactory.h
@@ -1,0 +1,31 @@
+//	Greenplum Database
+//	Copyright (C) 2016 Pivotal Software, Inc.
+
+#ifndef GPOPT_CMetadataAccessorFactory_H
+#define GPOPT_CMetadataAccessorFactory_H
+
+#include "gpos/memory/IMemoryPool.h"
+#include "gpos/common/CAutoP.h"
+#include "gpopt/mdcache/CMDAccessor.h"
+#include "gpopt/minidump/CDXLMinidump.h"
+
+namespace gpopt
+{
+	class CMetadataAccessorFactory
+	{
+		public:
+			CMetadataAccessorFactory
+				(
+					IMemoryPool *pmp,
+					CDXLMinidump *pdxlmd,
+					const CHAR *szFileName
+				);
+
+			CMDAccessor *Pmda();
+
+		private:
+			CAutoP<CMDAccessor> m_apmda;
+	};
+
+}
+#endif
--- a/libgpopt/src/minidump/CMetadataAccessorFactory.cpp
+++ b/libgpopt/src/minidump/CMetadataAccessorFactory.cpp
@@ -1,0 +1,41 @@
+//	Greenplum Database
+//	Copyright (C) 2016 Pivotal Software, Inc.
+
+#include "gpopt/mdcache/CMDCache.h"
+#include "gpopt/minidump/CMetadataAccessorFactory.h"
+#include "gpos/common/CAutoRef.h"
+#include "naucrates/md/CMDProviderMemory.h"
+
+namespace gpopt
+{
+	CMetadataAccessorFactory::CMetadataAccessorFactory
+		(
+			IMemoryPool *pmp,
+			CDXLMinidump *pdxlmd,
+			const CHAR *szFileName
+		)
+	{
+
+		// set up MD providers
+		CAutoRef<CMDProviderMemory> apmdp(GPOS_NEW(pmp) CMDProviderMemory(pmp, szFileName));
+		const DrgPsysid *pdrgpsysid = pdxlmd->Pdrgpsysid();
+		CAutoRef<DrgPmdp> apdrgpmdp(GPOS_NEW(pmp) DrgPmdp(pmp));
+
+		// ensure there is at least ONE system id
+		apmdp->AddRef();
+		apdrgpmdp->Append(apmdp.Pt());
+
+		for (ULONG ul = 1; ul < pdrgpsysid->UlLength(); ul++)
+		{
+			apmdp->AddRef();
+			apdrgpmdp->Append(apmdp.Pt());
+		}
+
+		m_apmda = GPOS_NEW(pmp) CMDAccessor(pmp, CMDCache::Pcache(), pdxlmd->Pdrgpsysid(), apdrgpmdp.Pt());
+	}
+
+	CMDAccessor *CMetadataAccessorFactory::Pmda()
+	{
+		return m_apmda.Pt();
+	}
+}
--- a/libgpopt/src/minidump/CMinidumperUtils.cpp
+++ b/libgpopt/src/minidump/CMinidumperUtils.cpp
@@ -18,6 +18,7 @@
 
 #include "gpos/base.h"
 #include "gpos/common/CAutoTimer.h"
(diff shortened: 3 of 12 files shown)


libgpopt/include/gpopt/minidump/CMetadataAccessorFactory.h
```diff
@@ -0,0 +1,28 @@
+#ifndef GPOPT_CMetadataAccessorFactory_H
```
vraghavan78: does not have the header like the other files. can we add that?

xinzweb: Yeah, we should put the license/copyright information there. Is there an easy way to always add those as part of the IDE?

libgpopt/src/minidump/CMetadataAccessorFactory.cpp
```diff
@@ -0,0 +1,40 @@
+#include "gpopt/minidump/CMetadataAccessorFactory.h"
```
xinzweb: Let's order the `#include`. Also, remove the empty lines between them.

libgpopt/include/gpopt/minidump/CMetadataAccessorFactory.h
```diff
@@ -0,0 +1,28 @@
...
+#define GPOPT_CMetadataAccessorFactory_H
+
+#include "gpos/memory/IMemoryPool.h"
+#include "gpos/common/CAutoP.h"
+#include "gpopt/mdcache/CMDAccessor.h"
+#include "gpopt/minidump/CDXLMinidump.h"
+
+namespace gpopt
+{
+	class CMetadataAccessorFactory
```
xinzweb: Some documentation follow doxygen to help generating docs in the future?

Something like create `CMDAccessor` on stack, so that it can be automatically released when out of scope?

libgpopt/src/minidump/CMinidumperUtils.cpp
```diff
@@ -18,6 +18,7 @@
 
 #include "gpos/base.h"
```
xinzweb: Maybe just clean up code one file at a time. Try to order the header files.

libgpopt/src/minidump/CMinidumperUtils.cpp
```diff
@@ -388,28 +390,9 @@ CMinidumperUtils::PdxlnExecuteMinidump
...
-	// ensure there is at least ONE system id
-	pmdp->AddRef();
-	pdrgpmdp->Append(pmdp);
-
-	for (ULONG ul = 1; ul < pdrgpsysid->UlLength(); ul++)
-	{	
-		pmdp->AddRef();
-		pdrgpmdp->Append(pmdp);
-	}
+	CMetadataAccessorFactory factory(pmp, pdxlmd, szFileName);
```
xinzweb: Really like the new code, it's cleaner, and don't have to worry about 'Release()'. :)

libgpopt/src/translate/CTranslatorDXLToExpr.cpp
```diff
@@ -2120,16 +2120,22 @@ CTranslatorDXLToExpr::Ptabdesc
...
-	for (ULONG ul = 0; ul < ulColumns; ul++)
+	CHashMapIter<ULONG, ULONG, gpos::UlHash<ULONG>, gpos::FEqual<ULONG>, CleanupDelete<ULONG>, CleanupDelete<ULONG> > iter(phmululColMapping);
+	while (iter.FAdvance())
 	{
+		ULONG ulColDescPos = *iter.Pt();
+		ULONG ulMdColPos = *iter.Pk();
 		BOOL fNullable = false;
-		if (ul < pmdrel->UlColumns())
+
+		// we are only guarding against our tests
```
xinzweb: Let's fix this comment.

This is guarding the scenario where `IMDRelation` got a lot more columns than `CTableDescriptor`, excluding dropped columns. Usually those columns are `system columns` from host, which doesn't exists if we construct `CTableDescriptor` manually for testing.

server/src/unittest/gpopt/translate/CTranslatorDXLToExprTest.cpp
```diff
@@ -755,4 +786,27 @@ CTranslatorDXLToExprTest::EresUnittest_ScalarSubquery()
...
+					(
+					pmp,
+					factory.Pmda(),
+					NULL,
+					CTestUtils::Pcm(pmp)
+					);
+
+	CAutoRef<CExpression> apExpr(CTranslatorDXLToExprTest::Pexpr(pmp, szQueryDroppedColumn));
+
+	GPOS_ASSERT((*dynamic_cast<CLogicalGet*>(apExpr->Pop())->Ptabdesc()->Pdrgpcoldesc())[0]->FNullable() == false);
```
xinzweb: Just a personal opinion. I love short statements. 

```
CLogicalGet * actualGet = (CLogicalGet *) apExpr -> Pop();
bool actualNullable = (actualGet -> Ptabdesc() -> Pdrgpclodesc()) [0] -> FNullable();
GPOS_ASSERT(actualNullable == false);
```

libgpopt/src/translate/CTranslatorDXLToExpr.cpp
```diff
@@ -2120,16 +2120,22 @@ CTranslatorDXLToExpr::Ptabdesc
 						);
 
 	// populate column information from the dxl table descriptor
-	const ULONG ulColumns = pdxltabdesc->UlArity();
-	for (ULONG ul = 0; ul < ulColumns; ul++)
+	CHashMapIter<ULONG, ULONG, gpos::UlHash<ULONG>, gpos::FEqual<ULONG>, CleanupDelete<ULONG>, CleanupDelete<ULONG> > iter(phmululColMapping);
+	while (iter.FAdvance())
 	{
```
vraghavan78: why can't we iterate in a for loop like before. get the ulposition via:

`ULONG *pulPos = phmiulAttnoColMapping->PtLookup(&iAttNo);
 GPOS_ASSERT(NULL != pulPos); 
const IMDColumn *pmdcolNext = pmdrel->Pmdcol(*pulPos)`

This way you do not need that if statement

server/src/unittest/gpopt/translate/CTranslatorDXLToExprTest.cpp
d: Xin the reason we cannot do that is because `GPOS_ASSERT` will be gone in a `Release` build. The compiler will yell at us for dead variables...

In our ideal world, we always want our assertions in tests compiled and executed. I can't fix that in this patch. YET

libgpopt/src/minidump/CMetadataAccessorFactory.cpp
d: addressed in the new push

libgpopt/include/gpopt/minidump/CMetadataAccessorFactory.h
d: addressed in the new push.

And yes, we can put it in the IDE

server/src/unittest/gpopt/translate/CTranslatorDXLToExprTest.cpp
xinzweb: Why not using a RTL assert? `GPOS_RTL_ASSERT`?

PR #46, opened 2016-05-02
Fallback when we generate invalid GatherMotion in plans

d:
This patch checks the validity of GatherMotion nodes in plans to ensure that we either:
0. Generate a valid Gather node; or
0. Fallback to planner we the plan contains an invalid Gather which has only 1 input segments

This patch should stop the system from crashing in this case.


hsyuan:
@d please change your profile to public in greenplum-db org.

xinzweb:
@hsyuan @oarap @vraghavan78

hsyuan:
+1

vraghavan78:
LGTM

PR #52, opened 2016-05-19
Correct minidump for #46

d:
This patch replaces double dashes with C-style comments in the XML
comments. Otherwise our tests would be passing for the wrong reason: we
expect ORCA to fallback in the form of throwing an exception, but we
were giving it a pass for blowing up during XML parsing.

This corrects a subtle bug introduced in 978a927 as part of #46


vraghavan78:
+1

hsyuan:
+1

PR #53, opened 2016-05-24
[#117367069] Publish GPORCA github release packages through concourse

hsyuan:
This patch publishes orca source tarball, debug and release binaries, as well as
release tags through the concourse pipeline.

@xinzweb @oarap @vraghavan78 
Please take a look.


oarap:
+1 Fly it!

xinzweb:
+1

PR #54, opened 2016-05-25
Remove 'secondary' clutter from concourse pipeline. [#119779501]

hsyuan:
Removed the `secondary` from the group `all` to cut down on the clutter in the `all` view of Concourse. `secondary` has been renamed to `platforms` because its purpose is to build and test on different platforms, but different from the pipeline all, which is used for the ICG. 

Non-centos platform builds and tests were moved to the `platforms` group.

Fixed a source dependency to `build_gpdb_centos6` and also `orca_centos5_release`.

### DIFF
--- a/concourse/pipeline.yml
+++ b/concourse/pipeline.yml
@@ -6,10 +6,8 @@ groups:
   - gp_xerces
   - gpos_centos5_release
   - gpos_centos6_release
-  - gpos_secondary
   - orca_centos5_release
   - orca_centos6_release
-  - orca_secondary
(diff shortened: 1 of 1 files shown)


hsyuan:
check out https://gporca.ci.pivotalci.info/pipelines/dev?groups=all to see how it looks like right now.

hsyuan:
@xinzweb @oarap @vraghavan78 @d 
please take a look.

oarap:
LGTM

xinzweb:
+1

PR #55, opened 2016-05-31
118865569 fix wrong error message when produced a wrong plan with Gather with more than one segments.

xinzweb:
@vraghavan78 @d @oarap 

There are two commits.

First one, fixed the wrong error major/minor number used for this DXL validation error.

Second commit, fixed the unittest framework so that we can verify the negative test cases with expected exceptions. So that, we can capture such missed cases where we failed the DXL translation due to different exception.


vraghavan78:
@xinzweb can you update the objective of this PR. What were you trying to solve here.

xinzweb:
@vraghavan78 updated the description, please recheck it again. Thanks.

vraghavan78:
LGTM! Thanks

PR #58, opened 2016-06-08
[#120984085] Adds GUC and related tests for array expansion.

cramja:
This GUC will be used to control the MEMO size as well as optimization
time for large IN list or large array comparison expressions.

Only the Array with less number of elements than the GUC will be
expanded and participate in constraint derivation.

Tradeoff of using this GUC is loss of potential benefits from the
constraint derivation (e.g. conflict detection, partition elimination)
with shorter optimization time and less memory utilization.

### DIFF
--- a/libgpopt/include/gpopt/engine/CHint.h
+++ b/libgpopt/include/gpopt/engine/CHint.h
@@ -43,6 +43,8 @@ namespace gpopt
 
 			ULONG m_ulJoinArityForAssociativityCommutativity;
 
+			ULONG m_ulArrayExpansionThreshold;
+
 
 			// private copy ctor
 			CHint(const CHint &);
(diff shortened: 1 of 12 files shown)


libgpopt/include/gpopt/engine/CHint.h
```diff
@@ -78,13 +82,23 @@ namespace gpopt
 				return m_ulJoinArityForAssociativityCommutativity;
 			}
 
+			// Maximum number of elements in the scalar comparison with an array which
+			// will be expanded in constraint derivation. The benefits of using a smaller number
```
vraghavan78: in -> during

server/include/unittest/gpopt/minidump/CArrayExpansionTest.h
```diff
@@ -0,0 +1,37 @@
...
+//	@filename:
+//		CArrayExpansionTest.h
+//
+//	@doc:
+//		Test for Array expansion in WHERE clause
+//
+//	@owner:
+//
+//
+//	@test:
```
vraghavan78: @d does not like owner and test. Can we have a chore to delete this from all files?

libgpopt/src/base/CConstraint.cpp
```diff
@@ -106,10 +107,20 @@ CConstraint::PcnstrFromScalarArrayCmp
 
 		// get comparison type
 		IMDType::ECmpType ecmpt = CUtils::Ecmpt(popScArrayCmp->PmdidOp());
-
 		CExpression *pexprArray = (*pexpr)[1];
-		DrgPcnstr *pdrgpcnstr = GPOS_NEW(pmp) DrgPcnstr(pmp);
+
 		const ULONG ulArity = pexprArray->UlArity();
+
+		COptimizerConfig *poconf = COptCtxt::PoctxtFromTLS()->Poconf();
```
vraghavan78: please add a comment

vraghavan78:
+1 nice work @cramja @xinzweb @hsyuan @d

d:
Mark and I addressed the comments in the newly pushed commit 3839f45

cramja:
@d Does this mean I can merge?

oarap:
good job, +1

cramja:
Merged

PR #61, opened 2016-06-10
[#121264185] Removes unused tags in file head comments.

cramja:
The @test and @owner tags are not used in this project, therefore they
were left blank on almost every header and CPP file. We carefully
removed these tags from each file we found them for which they were
unused.

Here's [the script](https://gist.github.com/cramja/b84ecd3a63397b931e9567f6f8d4a49f) we used


cramja:
@vraghavan78 removed `inl` as well, moving to GPOS.

@xinzweb @d @hsyuan

hsyuan:
+1

vraghavan78:
+1

PR #63, opened 2016-06-10
Copy `regression.diffs` after ICG failure [#120858343]

hsyuan:
This, in conjunction with greenplum-db/gpdb#832 , will help us highlight `installcheck-good` failures by providing the `regression.diffs` file for posterity


oarap:
+1

PR #66, opened 2016-06-16
Orca generates wrong results for value scan with self comparison

vraghavan78:
Tracker Story: #121621417

```
vraghavan=# select * from (values (2),(null),(1)) v(k) where k = k;
 k
---
 2

 1
(3 rows)
vraghavan=# set optimizer_print_query = on;
LOG:  statement: set optimizer_print_query = on;
SET
vraghavan=# explain select * from (values (2),(null),(1)) v(k) where k = k;
LOG:  statement: explain select * from (values (2),(null),(1)) v(k) where k = k;
LOG:  2016-06-15 13:36:11:897288 PDT,THD000,TRACE,"
Algebrized query:
+--CLogicalSelect
   |--CLogicalUnionAll Output: ("column1" (0)), Input: [("column1" (0)), ("column1" (1)), ("column1" (2))]
   |  |--CLogicalConstTableGet Columns: ["column1" (0)] Values: [(2)]
   |  |--CLogicalConstTableGet Columns: ["column1" (1)] Values: [(null)]
   |  +--CLogicalConstTableGet Columns: ["column1" (2)] Values: [(1)]
   +--CScalarCmp (=)
      |--CScalarIdent "column1" (0)
      +--CScalarIdent "column1" (0)

+--CLogicalUnionAll Output: ("column1" (0)), Input: [("column1" (0)), ("column1" (1)), ("column1" (2))]
   |--CLogicalConstTableGet Columns: ["column1" (0)] Values: [(2)]
   |--CLogicalConstTableGet Columns: ["column1" (1)] Values: [(null)]
   +--CLogicalConstTableGet Columns: ["column1" (2)] Values: [(1)]
",
2016-06-15 13:36:11:901171 PDT,THD000,TRACE,"
Physical plan:
+--CPhysicalUnionAll   rows:3   width:4  rebinds:1   cost:0.000024   origin: [Grp:3, GrpExpr:1]
   |--CPhysicalConstTableGet   rows:1   width:4  rebinds:1   cost:0.000004   origin: [Grp:0, GrpExpr:1]
   |--CPhysicalConstTableGet   rows:1   width:4  rebinds:1   cost:0.000004   origin: [Grp:1, GrpExpr:1]
   +--CPhysicalConstTableGet   rows:1   width:4  rebinds:1   cost:0.000004   origin: [Grp:2, GrpExpr:1]
",
                   QUERY PLAN
------------------------------------------------
 Append  (cost=0.00..0.00 rows=1 width=4)
   ->  Result  (cost=0.00..0.00 rows=1 width=4)
   ->  Result  (cost=0.00..0.00 rows=1 width=4)
   ->  Result  (cost=0.00..0.00 rows=1 width=4)
 Settings:  optimizer=on
 Optimizer status: PQO version 1.633
(6 rows)
```

As you can see, the predicate **k=k** is missing. Reasoning for this behavior is that the constraint cleanup code thought that the **k** is a non-nullable column. The bug of setting **k** as a nullable column is in **CTranslatorDXLToExpr::PexprLogicalConstTableGet**.

In addition to the above fix I also did the following:
* Refactor the function that is only called by **CTranslatorDXLToExpr::PexprLogicalConstTableGet** into that function.
* Disabled pushing filter into the constant table get -- because with this feature the DXL to PlStmt translator complains. I think this is an orthogonal issue. Will fix that in a separate fix.

### DIFF
--- a/libgpopt/src/translate/CTranslatorDXLToExpr.cpp
+++ b/libgpopt/src/translate/CTranslatorDXLToExpr.cpp
@@ -2381,33 +2381,33 @@ CTranslatorDXLToExpr::PexprLogicalConstTableGet
 	CDXLLogicalConstTable *pdxlopConstTable = CDXLLogicalConstTable::PdxlopConvert(pdxlnConstTable->Pdxlop());
 
 	const DrgPdxlcd *pdrgpdxlcd = pdxlopConstTable->Pdrgpdxlcd();
+
 	// translate the column descriptors
 	DrgPcoldesc *pdrgpcoldesc = GPOS_NEW(m_pmp) DrgPcoldesc(m_pmp);
 	const ULONG ulColumns = pdrgpdxlcd->UlLength();
 
-	for (ULONG ul = 0; ul < ulColumns; ul++)
+	for (ULONG ulColIdx = 0; ulColIdx < ulColumns; ulColIdx++)
 	{
-		CDXLColDescr *pdxlcd = (*pdrgpdxlcd)[ul];
+		CDXLColDescr *pdxlcd = (*pdrgpdxlcd)[ulColIdx];
 		const IMDType *pmdtype = m_pmda->Pmdtype(pdxlcd->PmdidType());
 		CName name(m_pmp, pdxlcd->Pmdname()->Pstr());
 
 		const ULONG ulWidth = pdxlcd->UlWidth();
-
 		CColumnDescriptor *pcoldesc = GPOS_NEW(m_pmp) CColumnDescriptor
 														(
 														m_pmp,
 														pmdtype,
 														name,
-														ul + 1, // iAttno
+														ulColIdx + 1, // iAttno
 														true, // FNullable
 														ulWidth
 														);
-		pdrgpcoldescOutput->Append(pcoldesc);
+		pdrgpcoldesc->Append(pcoldesc);
 	}
 
+	// translate values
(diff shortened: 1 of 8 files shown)


libgpopt/src/translate/CTranslatorExprToDXL.cpp
```diff
@@ -1476,10 +1476,6 @@ CTranslatorExprToDXL::PdxlnResult
 					pdxlprop
 					);
 		}
-		case COperator::EopPhysicalConstTableGet:
-		{
-			return PdxlnResultFromConstTableGet(pexprRelational, pdrgpcr, pexprScalar);
-		}
```
vraghavan78: Disabled pushing filter into the constant table get

libgpopt/src/translate/CTranslatorDXLToExpr.cpp
```diff
@@ -2421,12 +2380,34 @@ CTranslatorDXLToExpr::PexprLogicalConstTableGet
...
+		CName name(m_pmp, pdxlcd->Pmdname()->Pstr());
+
+		const ULONG ulWidth = pdxlcd->UlWidth();
+		CColumnDescriptor *pcoldesc = GPOS_NEW(m_pmp) CColumnDescriptor
+														(
+														m_pmp,
+														pmdtype,
+														name,
+														ulColIdx + 1, // iAttno
+														true, // FNullable
```
vraghavan78: This was **false**

hsyuan:
LGTM

PR #69, opened 2016-07-06
Fall back to planner when optimizing scalar group with non-scalar subgroups [#123212061]

hsyuan:
In very rare case, Orca may generate the plan that a group is a scalar
group, but it has non-scalar sub groups. i.e.:
Group 7 ():  --> pgroupRoot->FScalar() == true
   0: CScalarSubquery["?column?" (19)] [ 6 ]
Group 6 (#GExprs: 2): --> pgroupChild->FScalar() == false
   0: CLogicalProject [ 2 5 ]
   1: CPhysicalComputeScalar [ 2 5 ]
In the above case, because group 7 has scalar expression, Orca skipped
generating optimization context for group 7 and its subgroup group 6,
even the group 6 doesn't have scalar expression and it needs optimization.
Orca doesn't support this feature yet, so falls back to planner.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@ cmake_minimum_required(VERSION 3.0 FATAL_ERROR)
 project(gpopt LANGUAGES CXX C)
 
 set(GPORCA_VERSION_MAJOR 1)
-set(GPORCA_VERSION_MINOR 641)
+set(GPORCA_VERSION_MINOR 642)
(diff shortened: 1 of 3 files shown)


hsyuan:
@d @lpetrov-pivotal @vraghavan78 @cramja Please take a look.

cramja:
+1

lpetrov-pivotal:
+1

vraghavan78:
LGTM

hsyuan:
pushed to master:
https://github.com/greenplum-db/gporca/commit/fc9bd3b28cd3d37f71874b41051f6fb942c38eac

PR #76, opened 2016-07-18
[#125960675] Array Constraints Preprocesssing (Array Pt 3)

cramja:
### Preprocessing stage for Array Conversion

Adds an expression preprocessing stage to convert applicable expressions into array IN or NOT IN expressions. This creates more succinct expression trees and reduces redundancy in final expression output.

Therefore an expression like `x = 1 OR x = 2` should be converted into the equivalent `x IN (1,2)`.

This feature works by doing a recursive walk of the expression tree which looks for multiple AND or OR children which are either array expressions or ScalarComparison children. It then converts them into Constraints. Constraints hold logic to detect redundancy and also to convert OR/AND statements into array statements (Array Pt 1).

We chose to put this method after step 8 of expression preprocessing because step 9 will propagate constraints, and therefore it makes sense to propagate array constraints instead of OR constraints because they are more succinct and also preprocessing will attempt to re-derive constraints from propagated constraints which can lead to redundant expressions being created if array constraints are not passed.

This feature is controlled by the `EnableArrayDerive` traceflag.

@lpetrov-pivotal and I learned a lot about reference counting while implementing the Preprocessing method. We summarized our findings in [this document](https://docs.google.com/a/pivotal.io/document/d/1nndCf-ZreeLYtdXeVdOy6BHx_xgyVbVrv9GSdMkxoRg/edit?usp=sharing).

### DIFF
--- a/libgpopt/include/gpopt/base/CConstraintInterval.h
+++ b/libgpopt/include/gpopt/base/CConstraintInterval.h
@@ -18,9 +18,14 @@
 #include "gpopt/base/CConstraint.h"
 #include "gpopt/base/CRange.h"
 #include "gpopt/operators/CScalarConst.h"
+#include "gpopt/operators/CScalarArrayCmp.h"
 
 namespace gpopt
 {
+
+	// range array
+	typedef CDynamicPtrArray<CRange, CleanupRelease> DrgPrng;
+
 	using namespace gpos;
 	using namespace gpmd;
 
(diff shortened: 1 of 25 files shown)


libgpopt/include/gpopt/base/CConstraintInterval.h
```diff
@@ -204,6 +223,9 @@ namespace gpopt
 			virtual
 			CExpression *PexprScalar(IMemoryPool *pmp);
 
+			// scalar expression  which will be a disjunction
```
vraghavan78: huh? can u elaborate?

cramja:
@vraghavan78 I'm closing this PR, moving to a simpler branch: https://github.com/greenplum-db/gporca/pull/84

PR #79, opened 2016-08-02
Fix bug that Orca Generates Wrong Results

hsyuan:
Fix bug ORCA generates wrong results in rare circumstances that satisfy all of the following criteria:
1. The query contains a the project defined as a scalar subquery with a limit (1 or 0)
2. The subquery contains a count aggregate over a subexpression
3. Said subexpression returns an empty set.

For example:
```sql
insert into foo values (1,1);
create table bar (c int, d int);
select (select count(*) from bar group by c limit 1) from foo;
 ?column?
----------
        0
(1 row)
```
It generated wrong result. The right result is `null` instead of `0`.
The bug has been fixed in this pull request.

### DIFF
--- a/libgpopt/src/xforms/CSubqueryHandler.cpp
+++ b/libgpopt/src/xforms/CSubqueryHandler.cpp
@@ -622,13 +622,17 @@ CSubqueryHandler::FCreateOuterApplyForScalarSubquery
 		return fSuccess;
 	}
 
+	DrgPcr *pdrgpcrGroupingCols = NULL;
+	BOOL fHasLogicalGbAgg = CUtils::FHasLogicalGbAgg((*pexprSubquery)[0], &pdrgpcrGroupingCols);;
+	BOOL fHasGroupCols = fHasLogicalGbAgg && (0 < pdrgpcrGroupingCols->UlLength());
+
 	// add projection for subquery column
 	CExpression *pexprPrj = CUtils::PexprAddProjection(pmp, pexprLeftOuterApply, CUtils::PexprScalarIdent(pmp, pcr));
 	const CColRef *pcrComputed = CScalarProjectElement::PopConvert((*(*pexprPrj)[1])[0]->Pop())->Pcr();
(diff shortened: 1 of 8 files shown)


libgpopt/src/xforms/CSubqueryHandler.cpp
```diff
@@ -622,13 +622,17 @@ CSubqueryHandler::FCreateOuterApplyForScalarSubquery
 		return fSuccess;
 	}
 
+	DrgPcr *pdrgpcrGroupingCols = NULL;
```
vraghavan78: can you please add a comment of why this is added?

libgpopt/src/xforms/CSubqueryHandler.cpp
```diff
@@ -614,7 +614,16 @@ CSubqueryHandler::FCreateOuterApplyForScalarSubquery
...
+
+	DrgPcr *pdrgpcrGroupingCols = NULL;
+	BOOL fHasLogicalGbAgg = CUtils::FHasLogicalGbAgg((*pexprSubquery)[0], &pdrgpcrGroupingCols);;
+	BOOL fHasGroupCols = fHasLogicalGbAgg && (0 < pdrgpcrGroupingCols->UlLength());
+
+	// if the subquery doesn't have count agg, just set outer apply expression and return.
+	// if the subquery has both count and group by aggregation, and the grouping column number > 0,
+	// the count(*) column can return NULL. In this case, we don't need proceed to create coalesce
+	// operator.
+	if (!fHasCountAgg || fHasGroupCols)
```
vraghavan78: we are doing a 2 pass algorithm. Is this necessary?

vraghavan78: could we have tests that do not meet this criteria?

hsyuan: I will try 1 pass.

You want tests that have count agg but don't have group by agg?

PR #80, opened 2016-08-02
Fix bug that orca incorrectly collapses these computed columns into a single compute node [#127534257]

hsyuan:
If the query has a cascade of computed columns with set returning function,
ORCA incorrectly collapses these computed columns into a single compute node
leading to a wrong result. For example:

create table t(a text, b text);
insert into t values('10.20,12.50','ABC;DEF');
select c, regexp_split_to_table(b, ';')
from (select regexp_split_to_table(a, ',') as c, b from t) dumb(c, b);

This issue has been fixed in this patch.

### DIFF
--- a/server/src/unittest/gpopt/minidump/CICGTest.cpp
+++ b/server/src/unittest/gpopt/minidump/CICGTest.cpp
@@ -52,6 +52,8 @@ const CHAR *rgszFileNames[] =
 		"../data/dxl/minidump/CollapseCascadeProjects2of2.mdp",
 		"../data/dxl/minidump/CollapseCascadeProjects2of3.mdp",
 		"../data/dxl/minidump/CannotCollapseCascadeProjects.mdp",
+		"../data/dxl/minidump/CollapseProject-SetReturning.mdp",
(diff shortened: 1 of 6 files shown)


server/src/unittest/gpopt/minidump/CICGTest.cpp
```diff
@@ -52,6 +52,8 @@ const CHAR *rgszFileNames[] =
 		"../data/dxl/minidump/CollapseCascadeProjects2of2.mdp",
 		"../data/dxl/minidump/CollapseCascadeProjects2of3.mdp",
 		"../data/dxl/minidump/CannotCollapseCascadeProjects.mdp",
+		"../data/dxl/minidump/CollapseProject-SetReturning.mdp",
+		"../data/dxl/minidump/CollapseProject-SetReturning-CTE.mdp",
 		"../data/dxl/minidump/RemoveUnusedProjElementsInGbAgg.mdp",
 		"../data/dxl/minidump/RemoveUnusedProjElements.mdp",
```
vraghavan78: from line 51 to 57 we have a new test. May be.

vraghavan78:
LGTM

PR #83, opened 2016-08-03
Extract CDatumSortedSet

cramja:
We extracted into a class the logic of constructing an array of datums to feed into the construction of ranges for an interval constraint.

This removes repeated deduplication logic in the creation of `CRange`s.

### DIFF
--- a/libgpopt/CMakeLists.txt
+++ b/libgpopt/CMakeLists.txt
@@ -890,6 +890,8 @@ add_library(gpopt
             src/operators/CPhysicalParallelUnionAll.cpp
             include/gpopt/base/CDistributionSpecStrictHashed.h
             src/base/CDistributionSpecStrictHashed.cpp
+            include/gpopt/base/CDatumSortedSet.h
+            src/base/CDatumSortedSet.cpp
             )
 
 target_link_libraries(gpopt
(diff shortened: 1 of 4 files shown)


cramja:
@hsyuan @oarap Please take a look. It's a small change, I promise!

vraghavan78:
LGTM

libgpopt/src/base/CDatumSet.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+#include "gpopt/operators/COperator.h"
+#include "gpopt/base/CDatumSet.h"
+#include "gpopt/operators/CScalarConst.h"
+#include "gpopt/base/CUtils.h"
+#include "gpos/common/CAutoRef.h"
+
+using namespace gpopt;
+
+CDatumSet::CDatumSet
```
hsyuan: The name `CDatumSet` is a little bit misleading. `Set` is OK, we knows from `Set` that the elements are unique, but better call it `CDatumSortedSet`, since it is sorted. Thoughts?

cramja: @hsyuan that's a fair point. I will change that

cramja: changed

hsyuan:
＋1

cramja:
Great, merging!

d:
Next time when such renaming happens, don't forget to change the commit message

PR #84, opened 2016-08-05
[#125960675] Preprocessing stage for Array Conversion

cramja:
### Preprocessing stage for Array Conversion

Adds an expression preprocessing stage to convert applicable expressions into array IN or NOT IN expressions. This creates more succinct expression trees and reduces redundancy in final expression output.

Therefore an expression like `x = 1 OR x = 2` should be converted into the equivalent `x IN (1,2)`.

This feature works by doing a recursive walk of the expression tree which looks for multiple AND or OR children which are either array expressions or ScalarComparison children. It then converts them into Constraints. Constraints hold logic to detect redundancy and also to convert OR/AND statements into array statements (Array Pt 1).

We chose to put this method after step 8 of expression preprocessing because step 9 will propagate constraints, and therefore it makes sense to propagate array constraints instead of OR constraints because they are more succinct and also preprocessing will attempt to re-derive constraints from propagated constraints which can lead to redundant expressions being created if array constraints are not passed.

This feature is controlled by the `EnableArrayDerive` traceflag.

@lpetrov-pivotal and I learned a lot about reference counting while implementing the Preprocessing method. We summarized our findings in [this document](https://docs.google.com/a/pivotal.io/document/d/1nndCf-ZreeLYtdXeVdOy6BHx_xgyVbVrv9GSdMkxoRg/edit?usp=sharing).

### DIFF
--- a/libgpopt/include/gpopt/operators/CExpressionPreprocessor.h
+++ b/libgpopt/include/gpopt/operators/CExpressionPreprocessor.h
@@ -177,6 +177,14 @@ namespace gpopt
 				const CColRefSet *pcrsReqd
 				);
 
+			// generate a scalar bool op expression or return the only child expression in array
+			static CExpression *
+			PexprScalarBoolOpConvert2In(IMemoryPool *pmp, CScalarBoolOp::EBoolOperator eboolop, DrgPexpr *pdrgpexpr);
+
+			// determines if the expression is likely convertible to an array expression
+			static BOOL
+			FConvert2InIsConvertable(CExpression *pexpr, CScalarBoolOp::EBoolOperator eboolop);
+
 			// private ctor
 			CExpressionPreprocessor();
 
@@ -200,6 +208,10 @@ namespace gpopt
 			// derive constraints on given expression
 			static
 			CExpression *PexprAddPredicatesFromConstraints(IMemoryPool *pmp, CExpression *pexpr);
+
+			// convert series of AND or OR comparisons into array IN expressions
+			static
+			CExpression *PexprConvert2In(IMemoryPool *pmp, CExpression *pexpr);
 
 	}; // class CExpressionPreprocessor
 }
--- a/libgpopt/src/operators/CExpressionPreprocessor.cpp
+++ b/libgpopt/src/operators/CExpressionPreprocessor.cpp
@@ -16,6 +16,7 @@
 #include "gpopt/base/CColRefTable.h"
 #include "gpopt/base/CConstraintInterval.h"
 #include "gpos/common/CAutoTimer.h"
+#include "gpos/common/CAutoRef.h"
 #include "gpopt/exception.h"
 
 #include "gpopt/operators/CWindowPreprocessor.h"
@@ -531,6 +532,158 @@ CExpressionPreprocessor::PexprRemoveSuperfluousOuterRefs
 	return GPOS_NEW(pmp) CExpression(pmp, pop, pdrgpexprChildren);
 }
 
+//---------------------------------------------------------------------------
+//	@function:
+//		CUtils::PexprScalarBoolOpConvert2In
+//
+//	@doc:
+//		Generate a ScalarBoolOp expression or simply return the only expression
+//		in the array if there is only one.
+//
+//---------------------------------------------------------------------------
+CExpression *
+CExpressionPreprocessor::PexprScalarBoolOpConvert2In
+	(
+	IMemoryPool *pmp,
+	CScalarBoolOp::EBoolOperator eboolop,
+	DrgPexpr *pdrgpexpr
+	)
+{
+	GPOS_ASSERT(NULL != pdrgpexpr);
+	GPOS_ASSERT(0 < pdrgpexpr->UlLength());
+
+	if (1 == pdrgpexpr->UlLength())
+	{
+	    // if there is one child, do not wrap it in a bool op
+	    CExpression *pexpr = (* pdrgpexpr)[0];
+	    pexpr->AddRef();
+	    pdrgpexpr->Release();
+	    return pexpr;
+	}
+
+	return GPOS_NEW(pmp) CExpression
+			(
+			pmp,
+			GPOS_NEW(pmp) CScalarBoolOp(pmp, eboolop),
+			pdrgpexpr
+			);
+}
+
+//---------------------------------------------------------------------------
+//	@function:
+//		CExpressionPreprocessor::FConvert2InIsConvertable
+//
+//	@doc:
+//		Checks if the given expression is likely to be simplified by the constraints
+//		framework during array conversion. eboolop is the CScalarBoolOp type
+//		of the expression which contains the argument expression
+//
+//---------------------------------------------------------------------------
+BOOL
+CExpressionPreprocessor::FConvert2InIsConvertable(CExpression *pexpr, CScalarBoolOp::EBoolOperator eboolopParent)
+{
+	if (CPredicateUtils::FCompareIdentToConst(pexpr))
+	{
+		if (IMDType::EcmptEq == CScalarCmp::PopConvert(pexpr->Pop())->Ecmpt() &&
+				CScalarBoolOp::EboolopOr == eboolopParent)
+		{
+			return true;
+		}
+		else if (IMDType::EcmptNEq == CScalarCmp::PopConvert(pexpr->Pop())->Ecmpt() &&
+				CScalarBoolOp::EboolopAnd == eboolopParent)
+		{
+			return true;
+		}
+	}
+	else if (CPredicateUtils::FCompareIdentToConstArray(pexpr))
+	{
+		return true;
+	}
+	return false;
+}
+
+//---------------------------------------------------------------------------
+//	@function:
+//		CExpressionPreprocessor::PexprConvert2In
+//
+//	@doc:
+//		Converts series of AND or OR comparisons into array IN expressions. For
+//		example, x = 1 OR x = 2 will convert to x IN (1,2). This stage assumes
+//		the expression has been unnested using CExpressionUtils::PexprUnnest.
+//
+//---------------------------------------------------------------------------
+CExpression *
+CExpressionPreprocessor::PexprConvert2In
+	(
+	IMemoryPool *pmp,
+	CExpression *pexpr // does not take ownership
+	)
+{
+	// protect against stack overflow during recursion
+	GPOS_CHECK_STACK_SIZE;
+	GPOS_ASSERT(NULL != pmp);
+	GPOS_ASSERT(NULL != pexpr);
+
+	COperator *pop = pexpr->Pop();
+	if (CPredicateUtils::FOr(pexpr) || CPredicateUtils::FAnd(pexpr) )
+	{
+		// the bool op type of this node
+		CScalarBoolOp::EBoolOperator eboolop = CScalarBoolOp::PopConvert(pop)->Eboolop();
+		// derive constraints on all of the simple scalar children
+		// and add them to a new AND or OR expression
+		DrgPexpr *pdrgpexprCollapse = GPOS_NEW(pmp) DrgPexpr(pmp);
+		DrgPexpr *pdrgpexprRemainder = GPOS_NEW(pmp) DrgPexpr(pmp);
+		const ULONG ulArity = pexpr->UlArity();
+		for (ULONG ul = 0; ul < ulArity; ul++)
+		{
+			CExpression *pexprChild = (*pexpr)[ul];
+
+			if (FConvert2InIsConvertable(pexprChild, eboolop))
+			{
+				pexprChild->AddRef();
+				pdrgpexprCollapse->Append(pexprChild);
+			}
+			else
+			{
+				// recursively convert the remainder and add to the array
+				pdrgpexprRemainder->Append(PexprConvert2In(pmp, pexprChild));
+			}
+		}
+
+		if (0 != pdrgpexprCollapse->UlLength())
+		{
+			// create the constraint, rederive the collapsed expression
+			// add the new derived expr to remainder
+			DrgPcrs *pdrgpcr = NULL;
+			pop->AddRef();
+			CAutoRef<CExpression> apexprPreCollapse(GPOS_NEW(pmp) CExpression(pmp, pop, pdrgpexprCollapse));
+			CAutoRef<CConstraint> apcnst(CConstraint::PcnstrFromScalarExpr(pmp, apexprPreCollapse.Pt(), &pdrgpcr));
+			CExpression *pexprPostCollapse = apcnst->PexprScalar(pmp);
+
+			pexprPostCollapse->AddRef();
+			pdrgpexprRemainder->Append(pexprPostCollapse);
+			CRefCount::SafeRelease(pdrgpcr);
+		}
+		else
+		{
+			pdrgpexprCollapse->Release();
+		}
+
+		GPOS_ASSERT(0 < pdrgpexprRemainder->UlLength());
+		return PexprScalarBoolOpConvert2In(pmp, eboolop, pdrgpexprRemainder);
+	}
+
+	DrgPexpr *pdrgpexpr = GPOS_NEW(pmp) DrgPexpr(pmp);
+	DrgPexpr *pdrgexprChildren = pexpr->PdrgPexpr();
+	for (ULONG ul = 0; ul < pexpr->UlArity(); ul++)
+	{
+		pdrgpexpr->Append(PexprConvert2In(pmp, (*pdrgexprChildren)[ul]));
+	}
+
+	pop->AddRef();
+	return GPOS_NEW(pmp) CExpression(pmp, pop, pdrgpexpr);
+
+}
 
 //---------------------------------------------------------------------------
 //	@function:
@@ -1996,10 +2149,20 @@ CExpressionPreprocessor::PexprPreprocess
 	GPOS_CHECK_ABORT;
 	pexprSubqUnnested->Release();
 
+	CExpression *pexprConvert2In = pexprUnnested;
+
+	if (GPOS_FTRACE(EopttraceEnableArrayDerive))
+	{
+		// (8.5) ensure predicates are array IN or NOT IN where applicable
+		pexprConvert2In = PexprConvert2In(pmp, pexprUnnested);
+		GPOS_CHECK_ABORT;
+		pexprUnnested->Release();
+	}
+
 	// (9) infer predicates from constraints
-	CExpression *pexprInferredPreds = PexprInferPredicates(pmp, pexprUnnested);
+	CExpression *pexprInferredPreds = PexprInferPredicates(pmp, pexprConvert2In);
 	GPOS_CHECK_ABORT;
-	pexprUnnested->Release();
+	pexprConvert2In->Release();
 
 	// (10) eliminate self comparisons
 	CExpression *pexprSelfCompEliminated = PexprEliminateSelfComparison(pmp, pexprInferredPreds);
--- a/server/src/unittest/gpopt/operators/CExpressionPreprocessorTest.cpp
+++ b/server/src/unittest/gpopt/operators/CExpressionPreprocessorTest.cpp
@@ -15,6 +15,7 @@
 #include "gpos/io/COstreamString.h"
 #include "gpos/string/CWStringDynamic.h"
 #include "gpos/task/CAutoTraceFlag.h"
+#include "gpos/common/CAutoRef.h"
 
 #include "gpopt/base/CUtils.h"
 #include "gpopt/eval/CConstExprEvaluatorDefault.h"
@@ -63,7 +64,10 @@ CExpressionPreprocessorTest::EresUnittest()
 		GPOS_UNITTEST_FUNC(EresUnittest_PreProcessOuterJoinMinidumps),
 		GPOS_UNITTEST_FUNC(EresUnittest_PreProcessOrPrefilters),
 		GPOS_UNITTEST_FUNC(EresUnittest_PreProcessOrPrefiltersPartialPush),
-		GPOS_UNITTEST_FUNC(EresUnittest_CollapseInnerJoin)
+		GPOS_UNITTEST_FUNC(EresUnittest_CollapseInnerJoin),
+		GPOS_UNITTEST_FUNC(EresUnittest_PreProcessConvert2InPredicate),
+		GPOS_UNITTEST_FUNC(EresUnittest_PreProcessConvertArrayWithEquals),
+		GPOS_UNITTEST_FUNC(EresUnittest_PreProcessConvert2InPredicateDeepExpressionTree)
 		};
 
 	return CUnittest::EresExecute(rgut, GPOS_ARRAY_SIZE(rgut));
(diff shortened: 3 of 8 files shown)


libgpopt/include/gpopt/operators/CExpressionPreprocessor.h
```diff
@@ -177,6 +177,10 @@ namespace gpopt
 				const CColRefSet *pcrsReqd
 				);
 
+			// determines if the expression is likely convertible to an array expression
+			static BOOL
+			PexprConvert2InIsConvertable(CExpression *pexpr, ULONG parentOpType);
```
cramja: ulParentOpType

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -531,6 +531,144 @@ CExpressionPreprocessor::PexprRemoveSuperfluousOuterRefs
...
+//		CExpressionPreprocessor::PexprConvert2InIsConvertable
+//
+//	@doc:
+//		Checks if the given expression is likely to be simplified by the constraints
+//		framework during array conversion. parentOpType is the CScalarBoolOp type
+//		of the expression which contains the argument expression
+//
+//---------------------------------------------------------------------------
+BOOL
+CExpressionPreprocessor::PexprConvert2InIsConvertable(CExpression *pexpr, ULONG parentOpType)
```
cramja: ulParentType

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -531,6 +531,144 @@ CExpressionPreprocessor::PexprRemoveSuperfluousOuterRefs
...
+	GPOS_CHECK_STACK_SIZE;
+	GPOS_ASSERT(NULL != pmp);
+	GPOS_ASSERT(NULL != pexpr);
+
+	COperator *pop = pexpr->Pop();
+	pop->AddRef();
+	if (CPredicateUtils::FOr(pexpr) || CPredicateUtils::FAnd(pexpr) )
+	{
+		// the bool op type of this node
+		ULONG popType = CScalarBoolOp::PopConvert(pop)->Eboolop();
```
cramja: ul

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -531,6 +531,144 @@ CExpressionPreprocessor::PexprRemoveSuperfluousOuterRefs
...
+		// derive constraints on all of the simple scalar children
+		// and add them to a new AND or OR expression
+		DrgPexpr *pdrgpexprCollapse = GPOS_NEW(pmp) DrgPexpr(pmp);
+		DrgPexpr *pdrgpexprRemainder = GPOS_NEW(pmp) DrgPexpr(pmp);
+		for (ULONG ulChild = 0; ulChild < pexpr->UlArity(); ulChild++)
```
cramja: pexpr->UlArity() outside

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -531,6 +531,144 @@ CExpressionPreprocessor::PexprRemoveSuperfluousOuterRefs
...
+			// add the new derived expr to remainder
+			DrgPcrs *pdrgpcr = NULL;
+			pop->AddRef();
+			CExpression *pexprPreCollapse = GPOS_NEW(pmp) CExpression(pmp, pop, pdrgpexprCollapse);
+			CConstraint *pcnst = CConstraint::PcnstrFromScalarExpr(pmp, pexprPreCollapse, &pdrgpcr);
+			CExpression *pexprPostCollapse = pcnst->PexprScalar(pmp);
+
+			pexprPreCollapse->Release();
+
+			// Release only releases the top expression (does not recurse to children).
```
cramja: comment unnecessary, though the whole release thing is confusing

server/src/unittest/gpopt/operators/CExpressionPreprocessorTest.cpp
```diff
@@ -2174,5 +2177,222 @@ CExpressionPreprocessorTest::EresUnittest_CollapseInnerJoin()
...
+//		Tests that a expression with nested OR statements will convert them into
+//		an array IN statement. The statement we are testing looks is equivalent to
+//		+LogicalGet
+//			+-ScalarBoolOp (Or)
+//				+-ScalarBoolCmp
+//					+-ScalarId
+//					+-ScalarConst
+//				+-ScalarBoolCmp
+//					+-ScalarId
+//					+-ScalarConst
```
cramja: add a comment to say what it should look like afterwards

server/src/unittest/gpopt/operators/CExpressionPreprocessorTest.cpp
```diff
@@ -2174,5 +2177,222 @@ CExpressionPreprocessorTest::EresUnittest_CollapseInnerJoin()
...
+
+	CAutoOptCtxt aoc(pmp, &mda, NULL /*pceeval*/, CTestUtils::Pcm(pmp));
+
+	CExpression *pexprGet = CTestUtils::PexprLogicalGet(pmp);
+	COperator *popGet = pexprGet->Pop();
+	popGet->AddRef();
+
+	// Create a disjunct, add as a child
+	CColRef *pcrLeft = CDrvdPropRelational::Pdprel(pexprGet->PdpDerive())->PcrsOutput()->PcrAny();
+	CScalarBoolOp *pcsclrOp = GPOS_NEW(pmp) CScalarBoolOp(pmp, CScalarBoolOp::EboolopOr);
```
cramja: hungerian

server/src/unittest/gpopt/operators/CExpressionPreprocessorTest.cpp
```diff
@@ -2174,5 +2177,222 @@ CExpressionPreprocessorTest::EresUnittest_CollapseInnerJoin()
...
+	// remove the array child and then make an OR node with two equality comparisons
+	CExpression *pexprArrayComp = (*pexpr->PdrgPexpr())[1];
+	GPOS_ASSERT(CUtils::FScalarArrayCmp(pexprArrayComp));
+
+	DrgPexpr *prngexprDisjChildren = GPOS_NEW(pmp) DrgPexpr(pmp);
+	prngexprDisjChildren->Append(pexprArrayComp);
+	prngexprDisjChildren->Append(CUtils::PexprScalarEqCmp(pmp, pcrLeft, CUtils::PexprScalarConstInt4(pmp, 6 /*iVal*/)));
+	prngexprDisjChildren->Append(CUtils::PexprScalarEqCmp(pmp, pcrLeft, CUtils::PexprScalarConstInt4(pmp, 7 /*iVal*/)));
+
+	CScalarBoolOp *pcsclrOp = GPOS_NEW(pmp) CScalarBoolOp(pmp, CScalarBoolOp::EboolopOr);
```
cramja: hungarian

server/src/unittest/gpopt/operators/CExpressionPreprocessorTest.cpp
```diff
@@ -2174,5 +2177,222 @@ CExpressionPreprocessorTest::EresUnittest_CollapseInnerJoin()
...
+	popGet->AddRef();
+
+	// get a column ref from the outermost Get expression
+	DrgPcr *pdrgpcr = CDrvdPropRelational::Pdprel(pexprGet->PdpDerive())->PcrsOutput()->Pdrgpcr(pmp);
+	GPOS_ASSERT(1 < pdrgpcr->UlLength());
+	CColRef *pcrLeft = (*pdrgpcr)[0];
+	CColRef *pcrRight = (*pdrgpcr)[1];
+
+	// inner most OR
+	CScalarBoolOp *pcsclrOpOrInner = GPOS_NEW(pmp) CScalarBoolOp(pmp, CScalarBoolOp::EboolopOr);
```
cramja: hungarian

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -531,6 +531,144 @@ CExpressionPreprocessor::PexprRemoveSuperfluousOuterRefs
...
+	}
+	else if (CPredicateUtils::FCompareIdentToConstArray(pexpr))
+	{
+		return true;
+	}
+	return false;
+}
+
+CExpression *
+PexprPullUpSingleConjDisj(CExpression *pexpr)
```
cramja: what is this?

cramja:
@vraghavan78 @d @hsyuan @oarap Please take a look

libgpopt/include/gpopt/operators/CExpressionPreprocessor.h
```diff
@@ -177,6 +177,10 @@ namespace gpopt
 				const CColRefSet *pcrsReqd
 				);
 
+			// determines if the expression is likely convertible to an array expression
+			static BOOL
+			FConvert2InIsConvertable(CExpression *pexpr, CScalarBoolOp::EBoolOperator eboolop);
```
hsyuan: Can we come up with a better method name for this one? e.g.
FcanConvertToArrayExpr / FexprToArrayExpr / FconvertableToInArrayExpr ?

cramja: I wanted to keep the prefix `Convert2In` because it is a helper method of `PexprConvert2In`. Would that change your opinion?

hsyuan: Ah, I see. then you can keep it.

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -531,6 +532,135 @@ CExpressionPreprocessor::PexprRemoveSuperfluousOuterRefs
...
+	if (CPredicateUtils::FOr(pexpr) || CPredicateUtils::FAnd(pexpr) )
+	{
+		// the bool op type of this node
+		CScalarBoolOp::EBoolOperator eboolop = CScalarBoolOp::PopConvert(pop)->Eboolop();
+		// derive constraints on all of the simple scalar children
+		// and add them to a new AND or OR expression
+		DrgPexpr *pdrgpexprCollapse = GPOS_NEW(pmp) DrgPexpr(pmp);
+		DrgPexpr *pdrgpexprRemainder = GPOS_NEW(pmp) DrgPexpr(pmp);
+		const ULONG ulChildren = pexpr->UlArity();
+		for (ULONG ulChild = 0; ulChild < ulChildren; ulChild++)
```
hsyuan: ulChild -> ul

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -531,6 +532,135 @@ CExpressionPreprocessor::PexprRemoveSuperfluousOuterRefs
...
+				pdrgpexprRemainder->Append(PexprConvert2In(pmp, pexprChild));
+			}
+		}
+
+		if (0 != pdrgpexprCollapse->UlLength())
+		{
+			// create the constraint, rederive the collapsed expression
+			// add the new derived expr to remainder
+			DrgPcrs *pdrgpcr = NULL;
+			pop->AddRef();
```
hsyuan: I don't quite understand pop was added ref twice. Besides L 591

server/src/unittest/gpopt/operators/CExpressionPreprocessorTest.cpp
```diff
@@ -2174,5 +2178,223 @@ CExpressionPreprocessorTest::EresUnittest_CollapseInnerJoin()
...
 
+//---------------------------------------------------------------------------
+//	@function:
+//		CExpressionPreprocessorTest
+//				::EresUnittest_PreProcessConvert2InPredicate
+//
+//	@doc:
+//		Tests that a expression with nested OR statements will convert them into
+//		an array IN statement. The statement we are testing looks is equivalent to
+//		+LogicalGet
```
hsyuan: 👍 a graph tells more than a long document

libgpopt/src/operators/CExpressionPreprocessor.cpp
cramja: agreed, the amount of ref-ing/deref-ing that object is confusing. Even if it is wrapped in an AutoRef, some areas will still have to ref it again because they are passed to new temporary expressions which, when released, will try to delete the `pop`.

Basically, I'm not sure if it's any clearer to use a smart pointer here so using ref-ing gymnastics might be the way to go.

server/src/unittest/gpopt/operators/CExpressionPreprocessorTest.cpp
cramja: Ty!

libgpopt/include/gpopt/operators/CExpressionPreprocessor.h
```diff
@@ -201,6 +205,10 @@ namespace gpopt
 			static
 			CExpression *PexprAddPredicatesFromConstraints(IMemoryPool *pmp, CExpression *pexpr);
 
+			// convert series of AND or OR comparisons into array IN expressions
+			static
+			CExpression *PexprConvert2In(IMemoryPool *pmp, CExpression *pexpr);
+
```
hsyuan: Do we need to consider the cases like this?
i in (1,2,3) or i = 4; --> i in (1,2,3,4)
i in (1,2,3) or i in (4,5,6); --> i in (1,2,3,4,5,6)

Maybe it is not necessary, which is not a general case.

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -531,6 +532,135 @@ CExpressionPreprocessor::PexprRemoveSuperfluousOuterRefs
...
+			CExpression *pexprSingle = (*pdrgpexprRemainder)[0];
+			pexprSingle->AddRef();
+			pdrgpexprRemainder->Release();
+			pop->Release(); // we use the op in all cases but this
+			return pexprSingle;
+		}
+		else
+		{
+			// otherwise, wrap the children in the original bool op
+			return GPOS_NEW(pmp) CExpression(pmp, pop, pdrgpexprRemainder);
```
hsyuan: if pdrgpexprRemainder->UlLength() is 0, do we still need pdrgpexprRemainder?

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -531,6 +532,135 @@ CExpressionPreprocessor::PexprRemoveSuperfluousOuterRefs
...
+		}
+	}
+
+	DrgPexpr *pdrgpexpr = GPOS_NEW(pmp) DrgPexpr(pmp);
+	DrgPexpr *pdrgexprChildren = pexpr->PdrgPexpr();
+	for (ULONG ulChild = 0; ulChild < pexpr->UlArity(); ulChild++)
```
hsyuan: `ulChild` -> `ul`, I just see all other codes use `ul`, so keep consistent.

server/src/unittest/gpopt/operators/CExpressionPreprocessorTest.cpp
```diff
@@ -2174,5 +2178,223 @@ CExpressionPreprocessorTest::EresUnittest_CollapseInnerJoin()
...
+
+	CAutoOptCtxt aoc(pmp, &mda, NULL /*pceeval*/, CTestUtils::Pcm(pmp));
+
+	CAutoRef<CExpression> apexprGet(CTestUtils::PexprLogicalGet(pmp)); // useful for colref
+	COperator *popGet = apexprGet->Pop();
+	popGet->AddRef();
+
+	// Create a disjunct, add as a child
+	CColRef *pcrLeft = CDrvdPropRelational::Pdprel(apexprGet->PdpDerive())->PcrsOutput()->PcrAny();
+	CScalarBoolOp *pscboolop = GPOS_NEW(pmp) CScalarBoolOp(pmp, CScalarBoolOp::EboolopOr);
```
hsyuan: pscboolop -> pscBoolOp

server/src/unittest/gpopt/operators/CExpressionPreprocessorTest.cpp
```diff
@@ -2174,5 +2178,223 @@ CExpressionPreprocessorTest::EresUnittest_CollapseInnerJoin()
...
+}
+
+//---------------------------------------------------------------------------
+//	@function:
+//		CExpressionPreprocessorTest
+//				::EresUnittest_PreProcessConvertArrayWithEquals
+//
+//	@doc:
+//		Test that an array expression like A IN (1,2,3,4,5) OR A = 6 OR A = 7
+//		converts to A IN (1,2,3,4,5,6,7).
```
hsyuan: Aha, you already have this test case as I mentioned above. Nice.

server/src/unittest/gpopt/operators/CExpressionPreprocessorTest.cpp
```diff
@@ -2174,5 +2178,223 @@ CExpressionPreprocessorTest::EresUnittest_CollapseInnerJoin()
...
+
+	CAutoRef<CExpression> apexpr(CTestUtils::PexprLogicalSelectArrayCmp(pmp));
+	// get a ref to the comparison column
+	CColRef *pcrLeft = CDrvdPropRelational::Pdprel(apexpr->PdpDerive())->PcrsOutput()->PcrAny();
+
+	// remove the array child and then make an OR node with two equality comparisons
+	CExpression *pexprArrayComp = (*apexpr->PdrgPexpr())[1];
+	GPOS_ASSERT(CUtils::FScalarArrayCmp(pexprArrayComp));
+
+	DrgPexpr *prngexprDisjChildren = GPOS_NEW(pmp) DrgPexpr(pmp);
```
hsyuan: what does `rng` stand for?

server/src/unittest/gpopt/operators/CExpressionPreprocessorTest.cpp
```diff
@@ -2174,5 +2178,223 @@ CExpressionPreprocessorTest::EresUnittest_CollapseInnerJoin()
...
+	popGet->AddRef();
+
+	// get a column ref from the outermost Get expression
+	CAutoRef<DrgPcr> apdrgpcr(CDrvdPropRelational::Pdprel(apexprGet->PdpDerive())->PcrsOutput()->Pdrgpcr(pmp));
+	GPOS_ASSERT(1 < apdrgpcr->UlLength());
+	CColRef *pcrLeft = (*apdrgpcr)[0];
+	CColRef *pcrRight = (*apdrgpcr)[1];
+
+	// inner most OR
+	CScalarBoolOp *pcsclrOpOrInner = GPOS_NEW(pmp) CScalarBoolOp(pmp, CScalarBoolOp::EboolopOr);
```
hsyuan: what does `csclr` mean in the var name?

libgpopt/src/operators/CExpressionPreprocessor.cpp
cramja: Okay, makes since. It seems like `ul` is hungarian for `i`, unless we were iterating with an int and then the hungarian would be `i` for `i`

libgpopt/include/gpopt/operators/CExpressionPreprocessor.h
vraghavan78: good point. @cramja does it handle that? Can we have a test for that?

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -531,6 +532,135 @@ CExpressionPreprocessor::PexprRemoveSuperfluousOuterRefs
...
+	return false;
+}
+
+//---------------------------------------------------------------------------
+//	@function:
+//		CExpressionPreprocessor::PexprConvert2In
+//
+//	@doc:
+//		Converts series of AND or OR comparisons into array IN expressions. For
+//		example, x = 1 OR x = 2 will convert to x IN (1,2). This stage assumes
```
vraghavan78: what if u had a nested or/and tree?

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -531,6 +532,135 @@ CExpressionPreprocessor::PexprRemoveSuperfluousOuterRefs
...
+		// derive constraints on all of the simple scalar children
+		// and add them to a new AND or OR expression
+		DrgPexpr *pdrgpexprCollapse = GPOS_NEW(pmp) DrgPexpr(pmp);
+		DrgPexpr *pdrgpexprRemainder = GPOS_NEW(pmp) DrgPexpr(pmp);
+		const ULONG ulArity = pexpr->UlArity();
+		for (ULONG ul = 0; ul < ulArity; ul++)
+		{
+			CExpression *pexprChild = (*pexpr)[ul];
+
+			// if constrainable:
```
vraghavan78: comment not needed

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -531,6 +532,135 @@ CExpressionPreprocessor::PexprRemoveSuperfluousOuterRefs
...
+	CExpression *pexpr // does not take ownership
+	)
+{
+	// protect against stack overflow during recursion
+	GPOS_CHECK_STACK_SIZE;
+	GPOS_ASSERT(NULL != pmp);
+	GPOS_ASSERT(NULL != pexpr);
+
+	COperator *pop = pexpr->Pop();
+	pop->AddRef();
```
vraghavan78: why? Addref before u use it

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -531,6 +532,135 @@ CExpressionPreprocessor::PexprRemoveSuperfluousOuterRefs
...
+			pexprPostCollapse->AddRef();
+			pdrgpexprRemainder->Append(pexprPostCollapse);
+			CRefCount::SafeRelease(pdrgpcr);
+		}
+		else
+		{
+			pdrgpexprCollapse->Release();
+		}
+
+		// logic to create a expression from Remainder
```
vraghavan78: comment not needed.

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -531,6 +532,135 @@ CExpressionPreprocessor::PexprRemoveSuperfluousOuterRefs
...
+		if (1 == pdrgpexprRemainder->UlLength())
```
vraghavan78: huh! can this be 2?

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -531,6 +532,135 @@ CExpressionPreprocessor::PexprRemoveSuperfluousOuterRefs
...
+			pexprSingle->AddRef();
+			pdrgpexprRemainder->Release();
+			pop->Release(); // we use the op in all cases but this
+			return pexprSingle;
+		}
+		else
+		{
+			// otherwise, wrap the children in the original bool op
+			return GPOS_NEW(pmp) CExpression(pmp, pop, pdrgpexprRemainder);
+		}
```
vraghavan78:
```
CUtils::PexprOpExpr
	(
	IMemoryPool *pmp,
        COperator *pop,
   	DrgPexpr *pdrgpexpr
	)
{

GPOS_ASSERT(NULL != pdrgpexpr);
GPOS_ASSERT(0 < pdrgpexpr->UlLength());


if (1 == pdrgpexpr->UlLength())
{
	// if there is one child, do not wrap it in a bool op
	CExpression *pexpr = (* pdrgpexpr)[0];
	pexpr->AddRef();

	return pexpr;
}

// otherwise, wrap the children in the original bool op
pop->AddRef();

return GPOS_NEW(pmp) CExpression(pmp, pop, pdrgpexprRemainder);
}
```

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -531,6 +532,135 @@ CExpressionPreprocessor::PexprRemoveSuperfluousOuterRefs
...
+	}
+
+	DrgPexpr *pdrgpexpr = GPOS_NEW(pmp) DrgPexpr(pmp);
+	DrgPexpr *pdrgexprChildren = pexpr->PdrgPexpr();
+	for (ULONG ulChild = 0; ulChild < pexpr->UlArity(); ulChild++)
+	{
+		pdrgpexpr->Append(PexprConvert2In(pmp, (*pdrgexprChildren)[ulChild]));
+	}
+
```
vraghavan78: add pop->AddRef() here.

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -1996,10 +2126,24 @@ CExpressionPreprocessor::PexprPreprocess
 	GPOS_CHECK_ABORT;
 	pexprSubqUnnested->Release();
 
+	CExpression *pexprOptional = NULL;
```
vraghavan78: CExpression *pexprConvert2In = pexprUnnested;

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -1996,10 +2126,24 @@ CExpressionPreprocessor::PexprPreprocess
...
+
+	if (GPOS_FTRACE(EopttraceEnableArrayDerive))
+	{
+		// (8.5) ensure predicates are array IN or NOT IN where applicable
+		pexprOptional = PexprConvert2In(pmp, pexprUnnested);
```
vraghavan78: pexprUnnested  = PexprConvert2In(pmp, pexprUnnested);

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -1996,10 +2126,24 @@ CExpressionPreprocessor::PexprPreprocess
...
+		GPOS_CHECK_ABORT;
+		pexprUnnested->Release();
+	}
+	else
+	{
+		pexprOptional = pexprUnnested;
+	}
```
vraghavan78: remove else clause

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -1996,10 +2126,24 @@ CExpressionPreprocessor::PexprPreprocess
...
+
 	// (9) infer predicates from constraints
```
vraghavan78: update (9) to (10) and others too.

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -1996,10 +2126,24 @@ CExpressionPreprocessor::PexprPreprocess
...
-	CExpression *pexprInferredPreds = PexprInferPredicates(pmp, pexprUnnested);
+	CExpression *pexprInferredPreds = PexprInferPredicates(pmp, pexprOptional);
```
vraghavan78: pexprConvert2In

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -1996,10 +2126,24 @@ CExpressionPreprocessor::PexprPreprocess
...
 	GPOS_CHECK_ABORT;
-	pexprUnnested->Release();
+	pexprOptional->Release();
```
vraghavan78: pexprConvert2In

server/src/unittest/gpopt/operators/CExpressionPreprocessorTest.cpp
vraghavan78: +LogicalGet -> LogicalGet

server/src/unittest/gpopt/operators/CExpressionPreprocessorTest.cpp
```diff
@@ -2174,5 +2178,223 @@ CExpressionPreprocessorTest::EresUnittest_CollapseInnerJoin()
...
+//			+-ScalarArrayCmp
+//				+-ScalarId
+//				+-ScalarArray
+//					+-ScalarConst
+//					+-ScalarConst
+//			+-ScalarCmp
+//				+-ScalarId
+//				+-ScalarId
+//
+//---------------------------------------------------------------------------
```
vraghavan78: did u try things with NOT or nested AND / OR?

vraghavan78:
Sweet!

libgpopt/src/operators/CExpressionPreprocessor.cpp
cramja: @vraghavan78 I will add an assert: `GPOS_ASSERT(1 < pdrgpexprRemainder->UlLength());` to the else case because there should never be a scenario where no children are present. The remainder array will at this point hold the simplified expressions (ones which have gone through the constraints framework) as well as the expressions which we did not pass through the constraints framework and try to simplify.

If there is a case where the call to `CConstraint::PexprScalar` returns no Expressions, then we will fail the assert and that would be bad

libgpopt/src/operators/CExpressionPreprocessor.cpp
cramja: How about modifying an existing method to do the logic you are suggesting:

```
//	@function:
//		CUtils::PexprScalarBoolOp
//
//	@doc:
//		Generate a ScalarBoolOp expression
//
CExpression *
CUtils::PexprScalarBoolOp
	(
	IMemoryPool *pmp,
	CScalarBoolOp::EBoolOperator eboolop,
	DrgPexpr *pdrgpexpr
	)
{
	GPOS_ASSERT(NULL != pdrgpexpr);
	GPOS_ASSERT(0 < pdrgpexpr->UlLength());

	return GPOS_NEW(pmp) CExpression
			(
			pmp,
			GPOS_NEW(pmp) CScalarBoolOp(pmp, eboolop),
			pdrgpexpr
			);
}
```
To:
```
//	@function:
//		CUtils::PexprScalarBoolOp
//
//	@doc:
//		Generate a ScalarBoolOp expression
//
CExpression *
CUtils::PexprScalarBoolOp
	(
	IMemoryPool *pmp,
	CScalarBoolOp::EBoolOperator eboolop,
	DrgPexpr *pdrgpexpr
	)
{
	GPOS_ASSERT(NULL != pdrgpexpr);
	GPOS_ASSERT(0 < pdrgpexpr->UlLength());

	if (1 == pdrgpexpr->UlLength())
	{
	    // if there is one child, do not wrap it in a bool op
	    CExpression *pexpr = (* pdrgpexpr)[0];
	    pexpr->AddRef();

	    return pexpr;
	}

	return GPOS_NEW(pmp) CExpression
			(
			pmp,
			GPOS_NEW(pmp) CScalarBoolOp(pmp, eboolop),
			pdrgpexpr
			);
}
```

libgpopt/src/operators/CExpressionPreprocessor.cpp
cramja: changed add ref location to make this clearer

libgpopt/src/operators/CExpressionPreprocessor.cpp
cramja: 👍

libgpopt/src/operators/CExpressionPreprocessor.cpp
cramja: @vraghavan78  Actually we (@hsyuan) and I tried the change it broke many unrelated tests, so we will make another function instead.

server/src/unittest/gpopt/operators/CExpressionPreprocessorTest.cpp
cramja: Isn't hungarian rule to lowercase some contractio of the type name and then camelcase a given name? In which case, aren't we both right?

server/src/unittest/gpopt/operators/CExpressionPreprocessorTest.cpp
cramja: oops, should be `drg` for what I think is 'dynamic range'

server/src/unittest/gpopt/operators/CExpressionPreprocessorTest.cpp
cramja: definitely my weak hungarian showing. Should be something like `pscboolopOrInner`

server/src/unittest/gpopt/operators/CExpressionPreprocessorTest.cpp
cramja: Added. Created a private helper method in this test to create an Array In or Array Not In statement depending on an argument flag.

libgpopt/src/operators/CExpressionPreprocessor.cpp
cramja: We have that test: `EresUnittest_PreProcessConvert2InPredicateDeepExpressionTree`

libgpopt/include/gpopt/operators/CExpressionPreprocessor.h
cramja: Yep, there is a test for this `EresUnittest_PreProcessConvertArrayWithEquals`

libgpopt/src/operators/CExpressionPreprocessor.cpp
cramja: actually, we refactored part of this method to use a helper and the Ref-ing code got clearer around this area.

cramja:
@hsyuan @vraghavan78 All comments are addressed, should be GTG.

PR #86, opened 2016-08-08
Changes GPOS_ASSERT to GPOS_RTL_ASSERTS in a Constraint test

cramja:
Recent versions of GCC will error-out with unused variable messages in cases where variables are only used in DEBUG builds. In one problematic test, switching the asserts to build in retail should solve the issue.

@hlinnaka @d This should solve the compiler issue as a bandaid solution.

*Open question*: Shouldn't all asserts in tests compile into retail? I mean, tests are tests and not customer facing.


d:
To test it, use
`cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_COMPILER=g++-5 /path/to/orca` to reproduce the unused variable error.

This is great, bump the version and merge it!

PR #88, opened 2016-08-12
Remove unused class CStatsEquivClass and related test cases

hsyuan:
`CStatsEquivClass` is not used anymore in ORCA. So just remove it and its related test cases.


vraghavan78:
LGTM

PR #89, opened 2016-08-12
Revert "Don't genereate plan using ScalarArrayCmp qual for B-Tree index" [#128092771]

hsyuan:
This reverts commit ec95176 . ORCA will proceed to generate a plan
without assuming the capability (or in this case, in inability) of the
index scan to handle qual of type ScalarArrayOpExpr . A change in
translator will guard against such a plan if the database is unable to
execute an Index Scan with a `ScalarArrayOpExpr` qual.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@ cmake_minimum_required(VERSION 3.0 FATAL_ERROR)
 project(gpopt LANGUAGES CXX C)
 
 set(GPORCA_VERSION_MAJOR 1)
-set(GPORCA_VERSION_MINOR 658)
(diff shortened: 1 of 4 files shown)


hsyuan:
@vraghavan78 @d @oarap @cramja Please take a look.

cramja:
+1

hsyuan:
pushed to master
https://github.com/greenplum-db/gporca/commit/d257502ed510dfd94439ca68b2c8fab18ae8bb02
https://github.com/greenplum-db/gporca/commit/c29e9bbe3b5e49ed429fb9ec9397a75593d7f53b

PR #90, opened 2016-08-12
[#128239337] Fix constraint bug in preprocessor

cramja:
We ran into cases where the preprocessor step 'PexprConvert2In' would
try to create a constraint but fail and get a null value returned to it,
which would cause ORCA to crash. This was because the constraints
framework can only create constraints on certain types. By adding an
additional check for constrainable types we guarantee that a constraint
will be returned.

Also added an ASSERT to check for this condition.

### DIFF
--- a/libgpopt/src/operators/CExpressionPreprocessor.cpp
+++ b/libgpopt/src/operators/CExpressionPreprocessor.cpp
@@ -582,24 +582,32 @@ CExpressionPreprocessor::PexprScalarBoolOpConvert2In
 BOOL
 CExpressionPreprocessor::FConvert2InIsConvertable(CExpression *pexpr, CScalarBoolOp::EBoolOperator eboolopParent)
 {
+	bool fConvertableExpression = false;
 	if (CPredicateUtils::FCompareIdentToConst(pexpr))
 	{
-		if (IMDType::EcmptEq == CScalarCmp::PopConvert(pexpr->Pop())->Ecmpt() &&
-				CScalarBoolOp::EboolopOr == eboolopParent)
-		{
-			return true;
-		}
-		else if (IMDType::EcmptNEq == CScalarCmp::PopConvert(pexpr->Pop())->Ecmpt() &&
-				CScalarBoolOp::EboolopAnd == eboolopParent)
-		{
-			return true;
-		}
+		fConvertableExpression |=
+				IMDType::EcmptEq == CScalarCmp::PopConvert(pexpr->Pop())->Ecmpt() &&
+					CScalarBoolOp::EboolopOr == eboolopParent;
+		fConvertableExpression |=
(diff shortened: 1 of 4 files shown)


cramja:
@vraghavan78 @d @hsyuan please take a look. The essence of this change was that a nullptr was being returned when a non-constrainable type was being passed into the constraints frame work (method to convert Expression -> Constraint).

The function `CUtils::FConstrainableType` defines a constrainable type as:
* `int`
* or a type whose default comparators exist in the metadata accessor

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -582,24 +582,32 @@ CExpressionPreprocessor::PexprScalarBoolOpConvert2In
...
-		return true;
+		fConvertableExpression = true;
 	}
-	return false;
+
+	if (fConvertableExpression)
+	{
+		GPOS_ASSERT(0 < pexpr->UlArity());
+		CScalarIdent *pscid = CScalarIdent::PopConvert((*pexpr)[0]->Pop());
+		if (!CUtils::FConstrainableType(pscid->PmdidType()))
```
cramja: check for constrainable type

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -582,24 +582,32 @@ CExpressionPreprocessor::PexprScalarBoolOpConvert2In
...
-				CScalarBoolOp::EboolopOr == eboolopParent)
-		{
-			return true;
-		}
-		else if (IMDType::EcmptNEq == CScalarCmp::PopConvert(pexpr->Pop())->Ecmpt() &&
-				CScalarBoolOp::EboolopAnd == eboolopParent)
-		{
-			return true;
-		}
+		fConvertableExpression |=
```
cramja: This is the same logic, but refactored

hsyuan:
LGTM

vraghavan78:
First, I do not know what the problem that you were trying to fix. Was there a test that failed? If so which one and if not why not?

Second, as it stands now I cannot understand the reasoning for these constraints? Can we update that. Also indentation makes it damn hard.

cramja:
@vraghavan78 to the first question, yes, there was a test that failed. A minidump test in the CICGTest in the ctest suite failed. To repro:

`server/gporca_test -d ../data/dxl/minidump/UnnestSQJoins.mdp`

just be sure to add the flag `103026` to the traceflags to enable array constraints. Without this fix, the test should segfault around CExpressionPreprocessor.cpp:670:
```c++
CExpression *pexprPostCollapse = apcnst->PexprScalar(pmp);
```

To the second question. Which constraints are you referring to- is it that the need for a check for constrainable types?

vraghavan78:
LGTM

cramja:
bumping version and merging

PR #93, opened 2016-08-17
Don't push IsNull predicate into DPE inner child of Outer Join [#126166283]

oarap:
Orca generates wrong results when the IsNull is a predicate that is
passed to the Dynamic Partition Selector and it is an inner child of the
outer join. The following queries provides an example.

```sql
create table t1(a int, b int) DISTRIBUTED BY (a) PARTITION by RANGE(b) (START (1) END (3) EVERY (1));
create table t2(c int, d int);
insert into t2 values (1,1);
insert into t1 values (1,1);

select * from t2 left outer join t1 on t1.b = t2.d where t1.b is null;
```
The query generates a plan where IS NULL predicate is pushed down.

```
	->  Partition Selector for t1 (dynamic scan id: 1)  (cost=10.00..100.00 rows=34 width=4)
        	Filter: t1.b IS NULL
        	Partitions selected: 0 (out of 2)
        ->  Dynamic Table Scan on t1 (dynamic scan id: 1)  (cost=0.00..431.00 rows=1 width=8)
```

This PR fixes the issue and generates a plan that produces correct results.
```
        ->  Partition Selector for t1 (dynamic scan id: 1)  (cost=10.00..100.00 rows=34 width=4)
                Partitions selected: 2 (out of 2)
        ->  Dynamic Table Scan on t1 (dynamic scan id: 1)  (cost=0.00..431.00 rows=1 width=8)
```


oarap:
@vraghavan78 @d @cramja Please review

libgpopt/src/operators/CPhysical.cpp
```diff
@@ -872,8 +872,11 @@ CPhysical::PppsRequiredPushThruNAry
 		if (ppfmReqd->FContainsScanId(ulPartIndexId))
 		{
 			CExpression *pexpr = ppfmReqd->Pexpr(ulPartIndexId);
-			pexpr->AddRef();
-			ppfmResult->AddPartFilter(pmp, ulPartIndexId, pexpr, NULL /*pstats */);
+			if (1 != ulChildIndex && CUtils::FScalarNullTest(pexpr) && CUtils::FPhysicalOuterJoin(exprhdl.Pop()))
```
cramja: I'll try my best to summarize what @d and I found yesterday in regards to this statement:

This `if` condition seems overly restrictive as it is checking if the seemingly very specific condition of: "are we trying to push a null test on the outer side of a LOJ". A better pattern might be to inverse the condition and instead day "what are the conditions which we should push down this partition filter?". 

We tried to look for a query where a valid part filter could correctly be pushed down the outer side of LOJ via this codepath, but would fail this added condition. We could not find such a query, in part because most of these filter pushdown would have already occurred in preprocessing.

Then we tried running ICG with some hacked in code which would write a file when a potentially pushable query was ignored because of the new code. This did expose at least one query which was affected in the `indexjoin` test.

oarap:
@d @cramja @vraghavan78 Fixed bug in condition check, please take a look.

libgpopt/src/operators/CPhysical.cpp
cramja: Instead of the condition
```c++
if (1 != ulChildIndex && CUtils::FScalarNullTest(pexpr) && CUtils::FPhysicalOuterJoin(exprhdl.Pop()))
```
Given the that bug is only dealing with pushing a null test to the inner child of LOJ, can we use the more intuitive statement:
```c++
if (!(1 == ulChildIndex && CUtils::FScalarNullTest(pexpr) && CUtils::FPhysicalOuterJoin(exprhdl.Pop()))
```

libgpopt/src/base/CUtils.cpp
```diff
@@ -1342,6 +1342,27 @@ CUtils::FPhysicalJoin
...
+BOOL
+CUtils::FPhysicalOuterJoin
+	(
+	COperator *pop
+	)
+{
+	GPOS_ASSERT(NULL != pop);
+
+	return 	COperator::EopPhysicalLeftOuterNLJoin == pop->Eopid() ||
+		COperator::EopPhysicalLeftOuterHashJoin == pop->Eopid() ||
```
cramja: We have an additional test case covering the `EopPhysicalLeftOuterHashJoin ` case. Is it possible to come up with a test case covering the other 2 types of joins which this code affects?

hsyuan:
In this PR, we added a method called `FPhysicalOuterJoin`, which checks if a given operator is a physical outer join. There are 2 new test cases to test whether IsNull predicate is
passed to the dynamic partition selector, one is for `LeftOuterNLJoin`, the other one is for `LeftOuterHashJoin`, but no test case for `CorrelatedLeftOuterNLJoin`. We can only get correlated left outer join through subquery (scalar subquery, existential subquery, quantified subquery) and TVF (table value function), both 2 cases neither have explicit join nor explicit join filter, so as far as we know, we can't generate test case for `CorrelatedLeftOuterNLJoin` in this PR.

PR #94, opened 2016-08-22
Revert "Do not coalesce if count is not generated by quantified subqueries"

oarap:
This reverts commit d4bfae0786d991ccdba70ee6d1663463f13ce5df and commit
6bb3ade .

Turns out `COALESCE` is needed in cases where the subquery _is_ expected
to return 0 but when it is decorrelated into a left outer join, it may
become `NULL`.

Example scenario:
Given DDL:
```
CREATE TABLE foo(a, b) AS VALUES (1,1);
CREATE TABLE bar(c int, d int);
```
The following query should return 1 row of zero:
```
SELECT (
  SELECT count(*)
  FROM bar
  WHERE c = a
) AS k
FROM foo;
```
And because we decorrelated this subquery into an outer join:
```
SELECT count AS k
FROM
  FOO LEFT OUTER JOIN (
    SELECT c, count(*)
    FROM bar
    GROUP BY c
  ) AS bar
ON a = c;
```
Here, we need some way of retaining the count, one way of which is `coalesce`:
```
SELECT COALESCE(count,0) AS k
FROM
  FOO LEFT OUTER JOIN (
    SELECT c, count(*)
    FROM bar
    GROUP BY c
  ) AS bar
ON a = c;
```


vraghavan78:
LGTM

PR #95, opened 2016-08-24
Do not COALESCE the result of COUNT with GROUP BY in a scalar subquery

d:
## Do not COALESCE the result of COUNT with GROUP BY in a scalar subquery

Given DDL
```
CREATE TABLE foo (a, b) AS VALUES (1,1);
CREATE TABLE bar (c int, d int);
```
and a query projecting a scalar sub-query, e.g.
```
SELECT (SELECT avg(d) FROM bar WHERE c = a) FROM foo;
```
ORCA will try to decorrelate the subquery à la
```
SELECT avg
FROM foo LEFT OUTER JOIN (
  SELECT avg(d), c FROM bar
  GROUP BY c
) AS bar
ON c = a;
```
For a query containing `count`, special care needs to be taken because
`count` returns `0` on an empty relation where other queries will return
`NULL`.

For example, `SELECT (SELECT count(d) FROM bar WHERE c = a) FROM foo;`
will be transformed into
```
SELECT COALESCE(count, 0) as count
FROM foo LEFT OUTER JOIN (
  SELECT count(d), c FROM bar
  GROUP BY c
) AS bar
ON c = a;
```
Notice the `COALESCE` here is important because in cases where the
sub-query executes on an empty relation, the outer query expects a zero
instead of a `NULL`.

This treatment has a problem though: we don't expect a zero in all cases
with a scalar subquery that uses `count`: when we had a GROUP BY
clause in the subquery, e.g.
```
SELECT (SELECT COUNT(*) FROM bar GROUP BY c LIMIT 1) FROM foo;
```
ORCA used to produce the incorrect decorrelation:
```
SELECT COALESCE(count, 0) as count
FROM foo LEFT OUTER JOIN (
  SELECT count(*)
  FROM bar
  GROUP BY c
  LIMIT 1
) AS bar
ON true;
```
Notice the `COALESCE` above, it's harmful when `bar` is an empty
relation, and it's extraneous otherwise.

This commit resolves that issue.

This commit also fixes a related, but more nuanced bug where we failed
to correctly detect `COUNT` in scalar subqueries, and we generated plans
without `COALESCE` where it was actually desirable, an example victim
query is as follows:
```
SELECT (
  SELECT jazz.count
  FROM (SELECT count(*) FROM bar GROUP BY c LIMIT 1) AS bar,
  	(SELECT count(*) FROM jazz WHERE e = a) AS jazz
)
FROM foo;
```

With an empty table `jazz` and non-empty tables `bar`, and `foo`,
the expected output should be zeroes, not `NULL`s.

This exposes another bug in ORCA though, when we have empty tables
`bar` and `jazz`, but nonempty table `foo`. We will return zeroes,
whereas the expected output should be `NULL`s.

See also commit deae737 .

### DIFF
--- a/libgpopt/src/base/CUtils.cpp
+++ b/libgpopt/src/base/CUtils.cpp
@@ -2515,6 +2515,55 @@ CUtils::FHasCountAgg
 }
 
 
+static BOOL FCountAggMatchingColumn(CExpression* pexprPrjElem, const CColRef* pcr)
+{
+	CColRef* pcrCount = NULL;
+	return CUtils::FCountAggProjElem(pexprPrjElem, &pcrCount) && pcr == pcrCount;
+}
+
+
+BOOL
+CUtils::FHasCountAggMatchingColumn
+	(
+	const CExpression *pexpr,
+	const CColRef *pcr,
+	const CLogicalGbAgg **ppgbAgg
+	)
+{
+	COperator* pop = pexpr->Pop();
+	// base case, we have a logical agg operator
+	if (COperator::EopLogicalGbAgg == pop->Eopid())
+	{
+		const CExpression* const pexprProjectList = (*pexpr)[1];
+		GPOS_ASSERT(COperator::EopScalarProjectList == pexprProjectList->Pop()->Eopid());
+		const ULONG ulArity = pexprProjectList->UlArity();
+		for (ULONG ul = 0; ul < ulArity; ul++)
+		{
+			CExpression* const pexprPrjElem = (*pexprProjectList)[ul];
+			if (FCountAggMatchingColumn(pexprPrjElem, pcr))
+			{
+				const CLogicalGbAgg* pgbAgg = CLogicalGbAgg::PopConvert(pop);
+				*ppgbAgg = pgbAgg;
+				return true;
+			}
+		}
+	}
+	// recurse
+	else
+	{
+		const ULONG ulArity = pexpr->UlArity();
+		for (ULONG ul = 0; ul < ulArity; ul++)
+		{
+			const CExpression* pexprChild = (*pexpr)[ul];
+			if (FHasCountAggMatchingColumn(pexprChild, pcr, ppgbAgg))
+			{
+				return true;
+			}
+		}
+	}
+	return false;
+}
+
 //---------------------------------------------------------------------------
 //	@function:
 //		CUtils::PexprSum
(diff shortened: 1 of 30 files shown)


libgpopt/src/base/CUtils.cpp
```diff
@@ -2477,6 +2477,54 @@ CUtils::FCountAggProjElem
...
+		if (pdrgpcrGroupingCols->UlLength() > 0)
+		{
+			CExpression* pexprProjectList = (*pexpr)[1];
+			GPOS_ASSERT(COperator::EopScalarProjectList == pexprProjectList->Pop()->Eopid());
+			return FHasCountAgg(pexprProjectList, &pcr);
+		}
+		else
+		{
+			return false;
+		}
```
vraghavan78: no need for else{}

d:
Self -1 here: We found a problem with the original approach and a similar problem with this patch. We are submitting a new fix soon.

d:
Updated the PR, @oarap and @vraghavan78 please take a look

Thanks!
@hsyuan and @d

libgpopt/src/base/CUtils.cpp
```diff
@@ -2515,6 +2515,55 @@ CUtils::FHasCountAgg
...
+	COperator* pop = pexpr->Pop();
+	// base case, we have a logical agg operator
+	if (COperator::EopLogicalGbAgg == pop->Eopid())
+	{
+		const CExpression* const pexprProjectList = (*pexpr)[1];
+		GPOS_ASSERT(COperator::EopScalarProjectList == pexprProjectList->Pop()->Eopid());
+		const ULONG ulArity = pexprProjectList->UlArity();
+		for (ULONG ul = 0; ul < ulArity; ul++)
+		{
+			CExpression* const pexprPrjElem = (*pexprProjectList)[ul];
```
vraghavan78: why not check if this is the pcr we are looking for?

server/src/unittest/gpopt/minidump/CAggTest.cpp
```diff
@@ -28,6 +28,9 @@ ULONG CAggTest::m_ulAggTestCounter = 0;  // start from first test
 // minidump files
 const CHAR *rgszAggFileNames[] =
 {
+	"../data/dxl/minidump/ScalarSubqueryCountStarInJoin.mdp",
+	"../data/dxl/minidump/ScalarCorrelatedSubqueryCountStar.mdp",
+	"../data/dxl/minidump/ScalarSubqueryCountStar.mdp",
```
vraghavan78: one test: SELECT (SELECT a1 FROM (SELECT count(*) FROM jazz group by e union all SELECT count(*) from bar group by c) x(a1) LIMIT 1) FROM foo;

oarap:
closed by cdd832b193a1fb21c8fa43bc5161ea381aa2b621

libgpopt/src/base/CUtils.cpp
oarap: isn't that what `FCountAggMatchingColumn` does?

PR #96, opened 2016-08-25
Do not validate Gather Motion based on optimizer_segments [#129012817]

oarap:
When the number of actual segments is 1 and optimizer_segments is set to something greater than 1, orca fallbacks to planner with no reason. The check that causes the fallback should be corrected by checking if the input segments of the Gather matches the actual number of segments.

The issue has been fixed in this PR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@ cmake_minimum_required(VERSION 3.0 FATAL_ERROR)
 project(gpopt LANGUAGES CXX C)
 
 set(GPORCA_VERSION_MAJOR 1)
-set(GPORCA_VERSION_MINOR 665)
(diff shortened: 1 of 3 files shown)


d:
## TL;DR:
+1
ship it.

## Long version
This diff looks good.

Based on a conversation off-line with @hsyuan and @oarap , it's difficult to write a test for this case for standalone ORCA , because unfortunately the query (DXL message) is not self-contained, specifically it's lacking the number of hosts (segments) in the running system.

## Action Item
0. Let's move ahead with this fix without a test in ORCA *per se*
0. Let's try our best to find an opportunity, say in HAWQ, to test this
0. Please file a story in the ice box to include the number of segments in the DXL message

vraghavan78:
+1 to fix and @d

PR #97, opened 2016-09-02
Don't replace filter child when the correlated outer is sequence while translating expr to dxl

bhuvnesh2703:
When the outer child of correlated nest loop join is a partition table,
and it has predicate on the partition key, Orca generated plan with
dynamic partition selector and scan for the outer child, however while
translating plan expression to DXL ORCA always assumed that the outer child
is a relation instead of physical sequence, which caused Orca to replace
the partition selector with filter.

Here is an example to reproduce the issue:
```sql
create table t2(c int, d int);
create table h(j int, i int) DISTRIBUTED BY (i) PARTITION by RANGE(j) (START (1) END (3) EVERY (1));
select (select h.i from t2) from h where h.j = 1;
```
This issue has been fixed in this patch and relevant tests are added.

### DIFF
--- a/libgpopt/src/translate/CTranslatorExprToDXL.cpp
+++ b/libgpopt/src/translate/CTranslatorExprToDXL.cpp
@@ -3449,10 +3449,19 @@ CTranslatorExprToDXL::PdxlnResultFromNLJoinOuter
 	// create a result node from the input expression
 	CDXLNode *pdxlnResult = PdxlnResult(pexprRelational, pdrgpcr, pdrgpdsBaseTables, pulNonGatherMotions, pfDML, pdxlprop);
 
-	// add the new filter to the result replacing its original
-	// empty filter
-	CDXLNode *pdxlnFilter = PdxlnFilter(pdxlnCond);
-	pdxlnResult->ReplaceChild(EdxltsIndexFilter /*ulPos*/, pdxlnFilter);
+	// In case the OuterChild is a physical sequence, it will already have the filter in the partition selector and
+	// dynamic scan, thus we should not replace the filter.
+	if(EdxlopPhysicalSequence != pdxlnResult->Pdxlop()->Edxlop())
+	{
+		// add the new filter to the result replacing its original
+		// empty filter
+		CDXLNode *pdxlnFilter = PdxlnFilter(pdxlnCond);
+		pdxlnResult->ReplaceChild(EdxltsIndexFilter /*ulPos*/, pdxlnFilter);
+	}
+	else
+	{
+		pdxlnCond->Release();
+	}
 
 	return pdxlnResult;
 }
--- a/libgpopt/src/translate/CTranslatorExprToDXL.cpp
+++ b/libgpopt/src/translate/CTranslatorExprToDXL.cpp
@@ -3503,7 +3503,7 @@ CTranslatorExprToDXL::PdxlnNLJoin
 	(
 	CExpression *pexprInnerNLJ,
 	DrgPcr *pdrgpcr,
-	DrgPds *pdrgpdsBaseTables, 
+	DrgPds *pdrgpdsBaseTables,
 	ULONG *pulNonGatherMotions,
 	BOOL *pfDML
 	)
@@ -3513,7 +3513,7 @@ CTranslatorExprToDXL::PdxlnNLJoin
 
 	// extract components
 	CPhysical *pop = CPhysical::PopConvert(pexprInnerNLJ->Pop());
-	
+
 	CExpression *pexprOuterChild = (*pexprInnerNLJ)[0];
 	CExpression *pexprInnerChild = (*pexprInnerNLJ)[1];
 	CExpression *pexprScalar = (*pexprInnerNLJ)[2];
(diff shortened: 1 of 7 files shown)


libgpopt/src/translate/CTranslatorExprToDXL.cpp
```diff
@@ -3449,10 +3449,19 @@ CTranslatorExprToDXL::PdxlnResultFromNLJoinOuter
...
 	CDXLNode *pdxlnResult = PdxlnResult(pexprRelational, pdrgpcr, pdrgpdsBaseTables, pulNonGatherMotions, pfDML, pdxlprop);
 
-	// add the new filter to the result replacing its original
-	// empty filter
-	CDXLNode *pdxlnFilter = PdxlnFilter(pdxlnCond);
-	pdxlnResult->ReplaceChild(EdxltsIndexFilter /*ulPos*/, pdxlnFilter);
+	// In case the OuterChild is a physical sequence, it will already have the filter in the partition selector and
+	// dynamic scan, thus we should not replace the filter.
+	if(EdxlopPhysicalSequence != pdxlnResult->Pdxlop()->Edxlop())
+	{
```
d: -1

@vraghavan78 keep me honest here, but the existing code is clearly suggesting an assumption / precondition: that it assumes the outer relation is always a `TableScan`, as indicated by the enum used here `EdxltsIndexFilter`.
In the case we are trying to fix, this assumption is not true, hence the bug.

d:
-1

@vraghavan78 keep me honest here, but the existing code is clearly suggesting an assumption / precondition: that it assumes the outer relation is always a `TableScan`, as indicated by the enum used here `EdxltsIndexFilter`.
In the case we are trying to fix, this assumption is not true, hence the bug.

The condition we are adding here apparently suffers from the same flaw, that it still assumes the outer relation mostly comes from a `TableScan`, hence the replacement of the supposed `Filter` child.

Questions:
- What's in the condition passed in from the caller? Is it a better scan condition that we can benefit from?
- Apart from `TableScan`, what other operators can be on the outer side (`Sequence`, that goes without saying, but what else)?

@d and @oarap

hsyuan:
I don't think it always assume the outer child is `TableScan`. It can be a `Join`. As long as it can be viewed as a relation.

d:
> > it assumes the outer relation is always a TableScan, as indicated by the enum used here `EdxltsIndexFilter`

> I don't think it always assume the outer child is TableScan.

I apologize for not be clear the first time, let me try again: the enum `EdxltsIndexFilter`, as its name suggests, is specifically the offset/index of the filter child under a `TableScan`. The existing code, thus, is making a gross assumption that the outer child is always a `TableScan`, hence the bug. In our motivating case, a `Sequence` node appears in the outer side, and things blow up (not immediately, but later).

We cannot just put a band-aid here à la "business as usual if it's not a `Sequence`, because anything that's not a `TableScan` will potentially still cause the same bug.

bhuvnesh2703:
Good point Jesse. We can't say the outer child is `TableScan` only just by the enum name, or that the enum comes from `PhysicalTableScan`. The changed function is called when the outer is CPhysicalFilter. Any operator under CPhysicalFilter is expected to have the filter as its second child (1st one is project list), with one exception that is CPhysicalSequence. I don't think it is a band-aid solution, it is just corner-case that needs special consideration. And, we can't find any other counter example at the moment.
Let's also run this with @vraghavan78 and get his inputs.

d:
@bhuvnesh2703 per ORCA code convention, every DXL node class defines a set of enums for the offset (or `index` as they are named in code) of their child nodes. The original code made a gross assumption a `Filter` as the outer child of a correlated left join always has a `TableScan` relational child.

The proposed patch is a band-aid. In fact, it only fixes one corner case when the child of `Filter` is a `Sequence`.

libgpopt/src/translate/CTranslatorExprToDXL.cpp
vraghavan78: I agree with @d we have some magic going on. I would rather have a condition


if (EdxlopPhysicalSequence == pdxlnResult->Pdxlop()->Edxlop())
{
.... Do/Don't do something 1
}
else if is a table scan
{
.... Do/Don't do something 2
}
else
{
.... Do/Don't do something 3
}

hsyuan: Through the function `CTranslatorExprToDXL::PdxlnResultFromFilter`, we know that for the child of CPhysicalFilter, there are only 2 cases:

1. If the Relational child is a partition selector then calls function `PdxlnPartitionSelectorWithInlinedCondition` to translate partition selector to DXL node,
2. If not, calls `CTranslatorExprToDXLUtils::PdxlnResult` to translate relational child. the second child for the afore-mentioned relational node will always be a filter (1st child -projection list, 2nd child filter, 3rd child onetime filter) (Refer `CTranslatorExprToDXLUtils::PdxlnResult`)

This is also the only two cases for outer child of `PdxlnResultFromNLJoinOuter`. The bug happens when the outer child is a partition selector (DXL node is a sequence).

We removed the if condition to replace the filter, and it passed ctest. However, we are not sure if the test cases to cover all the cases are already existing or not. We are in process of running ICG, and will update the results.

bhuvnesh2703:
@oarap and @hsyuan,
@d and I pushed a new commit, trying to clarify some of the intent in this PR, would love to get an extra pair of eyes.

d:
OMG I was an idiot, will take that out ...

oarap:
Is it possible to separate the actual code changes and tab space kinda changes into separate commits? It is hard to follow what is an actual change with this way. Also it would be easier to track changes in the future.

hsyuan:
agreebwith @oarap, it is good to separate the changes.

oarap:
LGTM now.

PR #101, opened 2016-09-15
Only distribute on hashable columns

d:
Previously, ORCA may generate a (parallel append) plan that
redistributes on (among other columns) a column of an unhashable data
type. This plan is not executable by the executor. This commit fixes
that by adding a check on the hashability of the data type of each
column before generating the distribution spec for the motion.


hsyuan:
＋1

oarap:
+1 best code ever

danielgustafsson:
A question on this; when a new datatype is added to GPDB with hash distribution support, does it need to be added to orca in some way for this to work or is this an internal hashing to orca? (not familiar with orca internals so thought better ask)

hsyuan:
No, it doesn't need to be added to orca. Orca will get the datatype info from gpdb metadata. Whether a datatype is hashable or not is defined in gpdb.
@vraghavan78 Can you confirm?

danielgustafsson:
@hsyuan thanks for clarification, just wanted to doublecheck I wasn't breaking anything when adding datatypes to GPDB.

PR #103, opened 2016-09-20
[#130220363] Parallel Union with Random Motion

oarap:
This pull request introduces a new feature where the parallel union all with
Random Motion if there is no redistributable column exists in the input
columns for the union all. In this case, in order to preserve the
paralllel plan we introduce a Random Motion instead of Hash Redistribute
Motion.

It also checks whether the column type is `Redistributable` instead of `Hashable`. `Hashable` property of the type denotes that if the type is hash joinable. It picks up the value from the `pg_operator` table and reads from the column value `oprcanhash` for equality operator of the given type.

Instead, the `Redistributable` checks hardcoded value for types where it actually signifies if the type is itself hashable type. `isGreenplumDbHashable` function in `cdbhash.c` codebase passes the information for  `Redistributablibility` in the dxl.

This pull request will create the same plan from the explain view but the `optimizer_print_plan` guc  and the generated dxl shows the difference. 

For a query like below where `xmin` columns is `xid` type and it is not redistributable.

```
explain SELECT xmin from tbl1 UNION ALL SELECT xmin from tbl2;
```

```
+--CPhysicalMotionGather(master)   rows:1   width:4  rebinds:1   cost:862.000042   origin: [Grp:2, GrpExpr:2]
   +--CPhysicalParallelUnionAll   rows:1   width:4  rebinds:1   cost:862.000027   origin: [Grp:2, GrpExpr:1]
      |--CPhysicalMotionRandom   rows:1   width:34  rebinds:1   cost:431.000013   origin: [Grp:0, GrpExpr:2]
      |  +--CPhysicalTableScan "tbl1" ("tbl1")   rows:1   width:34  rebinds:1   cost:431.000006   origin: [Grp:0, GrpExpr:1]
      +--CPhysicalMotionRandom   rows:1   width:34  rebinds:1   cost:431.000013   origin: [Grp:1, GrpExpr:2]
         +--CPhysicalTableScan "tbl2" ("tbl2")   rows:1   width:34  rebinds:1   cost:431.000006   origin: [Grp:1, GrpExpr:1]
",
                                           QUERY PLAN
 Gather Motion 3:1  (slice3; segments: 3)  (cost=0.00..862.00 rows=1 width=4)
   ->  Append  (cost=0.00..862.00 rows=1 width=4)
               ->  Table Scan on tbl1  (cost=0.00..431.00 rows=1 width=4)
         ->  Redistribute Motion 3:3  (slice2; segments: 3)  (cost=0.00..431.00 rows=1 width=4)
               ->  Table Scan on tbl2  (cost=0.00..431.00 rows=1 width=4)
 Settings:  optimizer=on
```

The pull request also fixes minor issues with the test file provided from previous commit and renames the mdp file.


oarap:
@d @bhuvnesh2703 @hsyuan Please review...

d:
Let's prioritize this Wednesday morning @vraghavan78

vraghavan78:
+1

PR #104, opened 2016-09-22
Hint to disable CXformExpandNAryJoinDP when join has large number of children [#121181555]

vraghavan78:
CXformExpandNAryJoinDP uses dynamic programming (DP). It is an expensive transformation when the number of join children is more than 10 since computes all the subsets and join orders for
large number of tables.

In this commit we added a hint that sets a hard-limit on how many joins we want to consider in the `CXformExpandNAryJoinDP` rule.

-- Venky and Ivan

### DIFF
--- a/server/src/unittest/gpopt/minidump/CJoinOrderDPTest.cpp
+++ b/server/src/unittest/gpopt/minidump/CJoinOrderDPTest.cpp
@@ -1,0 +1,43 @@
+//---------------------------------------------------------------------------
+//	Greenplum Database
+//	Copyright (C) 2016 Pivotal Software
+//
+//	@filename:
+//		CJoinOrderDPTest.cpp
+//
+//	@doc:
+//		Test for n-ary join order
+//---------------------------------------------------------------------------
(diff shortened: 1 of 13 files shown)


server/src/unittest/gpopt/minidump/CJoinOrderDPTest.cpp
```diff
@@ -0,0 +1,43 @@
...
+#include "unittest/gpopt/CTestUtils.h"
+
+
+
+//---------------------------------------------------------------------------
+//	@function:
+//		CJoinOrderDPTest::EresUnittest
+//
+//	@doc:
+//		Unittest for array expansion in WHERE clause
```
hsyuan: update comment

hsyuan:
DXL Attribute `JoinOrderDynamicProgThreshold ` should be explicitly added to `CHint` element in the testing minidump file to make sure that the config value is parsed correctly. 

If we can have two separate minidump tests with difference join order threshold 12 (w/o DP) and 13 (w/ DP) to see difference plan, that will be good, but not required.

PR #108, opened 2016-10-07
Remove broadcast motion for physical sequence project [#131569023]

hsyuan:
When Window operator has row_number as the window function GPORCA does not generate the
correct distribution request from its child (asking Broadcast motion).

We have had other issues with Window operator requesting Broadcast motion from its child.
To avoid generating the incorrect plan we are disabling such a request till we fully
understand when such a request is guaranteed to be correct.


oarap:
LGTM

vraghavan78:
+1

PR #109, opened 2016-10-07
Fix bug that system id counter is never incremented during serialization [#131856005]

hsyuan:
This causes the incomplete minidump file when the query contains CTAS such as:
```sql
create table foo as
select row_number() over() as row_num, id_col
from bar
distributed by (id_col);
```


oarap:
LGTM

hsyuan:
@vraghavan78 @d please take a look.

oarap:
+1

d:
+1!
Looks good to us. We'll rebase this today.

Jesse and my anonymous pair

PR #111, opened 2016-10-22
Fix statistics computation for bug when nulls are present

vraghavan78:
ORCA computes wrong statistics for the following filter contains when NULLs are present in the histogram: 
* <> 
* NOT IN

The issue happens when NULLS are present.

Both <>, and NOT IN predicates will trigger a CHistogram::PhistNEqual histogram calculation.

Example predicates:

* Predicate is c <> 'J'. If the column has NULL values then those need to be returned.
* Predicate is c <> 'null'. If the column has NULL values then those tuples should not be returned.

@d @oarap @hsyuan please take a look.

### DIFF
--- a/data/dxl/minidump/notequal-predicate-over-data-with-nulls.mdp
+++ b/data/dxl/minidump/notequal-predicate-over-data-with-nulls.mdp
@@ -1,0 +1,656 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+create table test_stat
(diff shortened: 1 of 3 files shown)


hsyuan:
Can you elaborate more in the comment?

vraghavan78:
@hsyuan update the description.

vraghavan78:
Abandoning PR. It is wrong.

PR #112, opened 2016-10-24
Enable both unions in the search space, favor parallel union 130121327

oarap:
Previously, when `optimizer_parallel_union` GUC is set, we were only
generating plans with `parallel union all` and omitting `serial union
all`. In this commit, we generate plans for both `serial union all` and
`parallel union all`.

In some cases, enforcement framework may not let `parallel union all` to
be part of a plan because of newly introduced distribution specs and
motions it generates. It is better to provide the legacy union all
implementation to be part of the alternatives instead of completely
omitting the alternative.

### DIFF
--- a/libgpopt/src/xforms/CXformImplementUnionAll.cpp
+++ b/libgpopt/src/xforms/CXformImplementUnionAll.cpp
@@ -72,8 +72,7 @@ CXformImplementUnionAll::Transform
 
 	// extract components
 	CLogicalUnionAll *popUnionAll = CLogicalUnionAll::PopConvert(pexpr->Pop());
-	CPhysicalUnionAllFactory factory(popUnionAll, GPOS_FTRACE(EopttraceEnableParallelAppend));
-	CPhysicalUnionAll *popPhysicalUnionAll = factory.PopPhysicalUnionAll(pmp);
+	CPhysicalUnionAllFactory factory(popUnionAll);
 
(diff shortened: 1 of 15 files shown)


libgpopt/src/xforms/CXformImplementUnionAll.cpp
```diff
@@ -85,17 +84,40 @@ CXformImplementUnionAll::Transform
 		pdrgpexpr->Append(pexprChild);
 	}
 
-	// assemble physical operator
-	CExpression *pexprUnionAll =
+	BOOL fParallel = GPOS_FTRACE(EopttraceEnableParallelAppend);
+
+	if(fParallel)
```
xinzweb: Maybe comment:

`If enable Parallel Append, generate parallel append as alternative to serial append.`

Also, maybe moving the 'parallel append' block from line 87 to 106 as the end of the function.

libgpopt/src/xforms/CXformImplementUnionAll.cpp
```diff
@@ -85,17 +84,40 @@ CXformImplementUnionAll::Transform
...
+	{
+		CPhysicalUnionAll *popPhysicalParallelUnionAll = factory.PopPhysicalUnionAll(pmp, fParallel);
```
xinzweb: Just use `true` instead of `fParallel`.

oarap:
@d can you also review before we push this in?

vraghavan78:
Pair reviewed with @xinzweb and @oarap.

LGTM

PR #114, opened 2016-10-25
GPORCA computes wrong statistics when filter contains "<> or NOT IN + a IS NULL" condition [#132928535]

vraghavan78:
Histograms have the following components:
* Histogram buckets
* Null frequency 
* Distinct values that are not captured by the buckets in both the histograms

Consider the following scenario:

``` 
create table test_stat(a bigint, b integer, c varchar(1) ) distributed by (a);
insert into test_stat select id, mod(id,100),'J' from generate_series(1,1000000) id;
insert into test_stat select id, mod(id,100), null from generate_series(1,1000000) id;
select * from test_stat where c <> 'J' or c is null;
```
In the above query has a predicates combined by an OR. The predicates are on the same column `c`. GPORCA while computing the histogram of the OR, we do not compute the contribution of the null values and the distinct remain values.

In this patch we fix this issue.

### DIFF
--- a/libnaucrates/src/statistics/CHistogram.cpp
+++ b/libnaucrates/src/statistics/CHistogram.cpp
@@ -1943,7 +1943,7 @@ CHistogram::PhistUnionNormalized
 	// compute the total number of distinct values (NDV) that are not captured by the buckets in both the histograms
 	CDouble dNDVRemain = std::max(m_dDistinctRemain, phist->DDistinctRemain());
 
-	// compute the total number of rows having distinct values not  captured by the buckets in both the histograms
+	// compute the total number of rows having distinct values not captured by the buckets in both the histograms
 	CDouble dNDVRemainRows = this->DFreqRemain() * dRows;
 	dNDVRemainRows = dNDVRemainRows + (phist->DFreqRemain() * dRowsOther);
 
@@ -2027,14 +2027,14 @@ CHistogram::PhistUpdatedFrequency
 	}
 
 	CDouble dNullFreq = dNullRows / *pdRowOutput ;
-	CDouble dDistinctRemainFreq =  dNDVRemainRows / *pdRowOutput ;
+	CDouble dNDVRemainFreq =  dNDVRemainRows / *pdRowOutput ;
 	return GPOS_NEW(pmp) CHistogram
 			(
 			pdrgppbucketNew,
 			true /* fWellDefined */,
 			dNullFreq,
-			dDistinctRemain,
-			dDistinctRemainFreq,
+			dNDVRemain,
+			dNDVRemainFreq,
 			false /* fColStatsMissing */
 			);
 }
--- a/libnaucrates/src/statistics/CHistogram.cpp
+++ b/libnaucrates/src/statistics/CHistogram.cpp
@@ -1936,19 +1936,27 @@ CHistogram::PhistUnionNormalized
 	// add any leftover buckets from this histogram
 	AddBuckets(pmp, m_pdrgppbucket, pdrgppbucket, dRows, pdrgpdoubleRows, ul1, ulBuckets1);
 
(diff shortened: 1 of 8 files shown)


oarap:
@vraghavan78 we (@xinzweb and I) thought through the cases a little bit further to understand what would be an expected `DistinctRemain` and `FreqRemain` after merging two histograms. 

Here is what we found:

GIVEN
```
H1:
total rows r1
buckets  b1
null rows (dist 1, freq x1)
remain (dist y1, freq z1)

H2
total rows r2
buckets b2
null rows (dist 1, freq x2)
remain (dist y2, freq z2)
```

Need to produce H3, which is merge of H1 and H2
```
total rows r3
bucket b3
null rows (dist 1, freq x3)
remain (dist y3, freq z3)
```

When merge H1 and H2 into H3, here are some discussions

```
b3 = Merge (b1, b2)

UNION ALL:
# ASSUME totally disjoint
rows r3 = r1 + r2
x3 = (x1 * r1 + x2 * r2) / r3
y3 = (y1 + y2)
z3 = (z1 * r1 +  z2 * r2) /r3

# ASSUME totally strict subset (prefer: over-estimation of cardinality)
rows r3 = max(r1,  r2)
x3 = max(x1 * r1, x2 * r2)/r3
y3 = max (y1, y2)
z3 = (z1 * r1 +  z2 * r2) / r3

UNION: it’s distinct on top of UNION ALL, which impact the 'freq' values.
```

We would prefer the `strict subset` assumption, because we will `under-estimate` the number of distinct, hence ends up `over-estimation` for the cardinality, which will give us a **safer** plan, e.g. HJ is preferred over NL. 

With that said, the current implementation is a mix of above two approaches, and need further discussion.

libnaucrates/src/statistics/CHistogram.cpp
```diff
@@ -1936,19 +1936,27 @@ CHistogram::PhistUnionNormalized
...
-	}
-	phistResult->m_dNullFreq = dNullFreq;
+	// compute the total number of distinct values (NDV) that are not captured by the buckets in both the histograms
+	CDouble dNDVRemain = std::max(m_dDistinctRemain, phist->DDistinctRemain());
 
-	CDouble dDistinctRemain = std::max(m_dDistinctRemain, phist->m_dDistinctRemain);
-	CDouble dFreqRemain = std::max(m_dFreqRemain, phist->m_dFreqRemain);
-	phistResult->m_dDistinctRemain = dDistinctRemain;
-	phistResult->m_dFreqRemain = dFreqRemain;
+	// compute the total number of rows having distinct values not  captured by the buckets in both the histograms
```
vraghavan78: Done.

vraghavan78:
Thanks for that feedback. Updated the PR.

libnaucrates/src/statistics/CHistogram.cpp
```diff
@@ -1936,19 +1936,25 @@ CHistogram::PhistUnionNormalized
...
+	// compute the total number of distinct values (NDV) that are not captured by the buckets in both the histograms
+	CDouble dNDVRemain = std::max(m_dDistinctRemain, phist->DDistinctRemain());
 
-	CDouble dDistinctRemain = std::max(m_dDistinctRemain, phist->m_dDistinctRemain);
-	CDouble dFreqRemain = std::max(m_dFreqRemain, phist->m_dFreqRemain);
-	phistResult->m_dDistinctRemain = dDistinctRemain;
-	phistResult->m_dFreqRemain = dFreqRemain;
+	// compute the total number of rows having distinct values not captured by the buckets in both the histograms
+	CDouble dNDVRemainRows = std::max( (this->DFreqRemain() * dRows), (phist->DFreqRemain() * dRowsOther));
+
```
xinzweb: maybe rename `phist` to `phistOther`? Thanks.

PR #116, opened 2016-10-31
Fix UNION query falling back to planner when parallel union is enabled

hsyuan:
Orca failed to generate plans for the following query when parallel union is enabled:
```sql
SELECT * FROM foo UNION SELECT * FROM bar;
```

Because returning EpetRequired in the EpetDistribution of CPhysicalParallelUnionAll causes the enforcement framework to falsely introduce motions that can cause failure in orca to generate plans for some queries.

When we introduce unnecessary CPhysicalMotionHashDistribute on top of CPhysicalParallelUnionAll, the system actually detects that it is unnecessary. However, the optimization flow gets corrupted because of that sinceCOptimizationContexts are indeed related to the each other. One optimization context can be a child or a parent of another optimization context in another group.

To resolve the issue, share the same logic with the CPhysicalSerialUnionAll by lifting the CPhysicalSerialUnionAll::EpetRequired code to the parent class CPhysicalUnionAll.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@ cmake_minimum_required(VERSION 3.0 FATAL_ERROR)
 project(gpopt LANGUAGES CXX C)
 
 set(GPORCA_VERSION_MAJOR 1)
-set(GPORCA_VERSION_MINOR 682)
+set(GPORCA_VERSION_MINOR 683)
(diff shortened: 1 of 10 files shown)


vraghavan78:
+1

d:
0. I like separate commits, but db45c18 and f260860 really should be squashed, because db45c18 alone won't compile.

Looks good to me otherwise.

PR #117, opened 2016-11-02
Remove whole banning mechanism from gporca [#133570595]

hsyuan:
gporca has a set of banned API calls which needs to be allowed with the
ALLOW_xxx macro in order for gpopt to compile. But it should be the
library caller(GPDB/Orca)'s resposibility to take care of the function call.

see discussions on greenplum-db/gpdb#1136
and https://groups.google.com/a/greenplum.org/forum/#!topic/gpdb-dev/Mcw6JPav6h4


hsyuan:
@hlinnaka @danielgustafsson Please take a look.

oarap:
LGTM

PR #119, opened 2016-11-08
Filters above limit should not pushed down into a partition selector in subquery with limit clause [#133827909]

vraghavan78:
The bug is in derivation of required partition propagation specification from the child. Before this fix, we were pushing the partition constraints below Limit which will return wrong results.

To illustrate consider the following table schema and data:
```sql
create table h(j int, i int) DISTRIBUTED BY (i) PARTITION by RANGE(j) (START (1) END (3) EVERY (1));
insert into h values (1,1);
insert into h values (2,2);
insert into h values (2,2);
```

SQL in question is:

```
select * from (select * from h order by j limit 2) x where j = 2;
```

With Planner:
** As you can see from the explain the condition `j = 2` is not pushed below the limit
```sql
vraghavan=# set optimizer = off;
SET
vraghavan=# select * from (select * from h order by j limit 2) x where j = 2;
 j | i
---+---
 2 | 2
(1 row)

vraghavan=# explain select * from (select * from h order by j limit 2) x where j = 2;
                                             QUERY PLAN
 Subquery Scan x  (cost=3644.00..3644.07 rows=1 width=8)
   Filter: x.j = 2
   ->  Limit  (cost=3644.00..3644.05 rows=2 width=8)
         ->  Gather Motion 3:1  (slice1; segments: 3)  (cost=3644.00..3644.05 rows=2 width=8)
               Merge Key: public.h.j
               ->  Limit  (cost=3644.00..3644.01 rows=1 width=8)
                     ->  Sort  (cost=3644.00..4074.50 rows=57400 width=8)
                           Sort Key (Limit): public.h.j
                           ->  Append  (cost=0.00..1922.00 rows=57400 width=8)
                                 ->  Seq Scan on h_1_prt_1 h  (cost=0.00..961.00 rows=28700 width=8)
                                 ->  Seq Scan on h_1_prt_2 h  (cost=0.00..961.00 rows=28700 width=8)
 Settings:  optimizer=off
(13 rows)
```

Orca before the fix:
* PROBLEM: We push the `j=2` below the limit.

```
vraghavan=# select * from (select * from h order by j limit 2) x where j = 2;
 j | i
---+---
 2 | 2
 2 | 2
(2 rows)

vraghavan=# explain select * from (select * from h order by j limit 2) x where j = 2;                                                                                                                                                                       QUERY PLAN
----------------------------------------------------------------------------------------------------------------
 Result  (cost=0.00..431.00 rows=1 width=8)
   Filter: j = 2
   ->  Limit  (cost=0.00..431.00 rows=1 width=8)
         ->  Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..431.00 rows=1 width=8)
               Merge Key: j
                     ->  Partition Selector for h (dynamic scan id: 1)  (cost=10.00..100.00 rows=34 width=4)
                           Filter: j = 2
                           Partitions selected: 1 (out of 2)
                     ->  Sort  (cost=0.00..431.00 rows=1 width=8)
                           Sort Key: j
                           ->  Dynamic Table Scan on h (dynamic scan id: 1)  (cost=0.00..431.00 rows=1 width=8)
 Settings:  optimizer=on
 Optimizer status: PQO version 1.684
(14 rows)
```

Orca after the fix:

```
vraghavan=# select * from (select * from h order by j limit 2) x where j = 2;
 j | i
---+---
 2 | 2
(1 row)
vraghavan=# explain select * from (select * from h order by j limit 2) x where j = 2;
                                                      QUERY PLAN
----------------------------------------------------------------------------------------------------------------------
 Result  (cost=0.00..431.00 rows=1 width=8)
   Filter: j = 2
   ->  Limit  (cost=0.00..431.00 rows=1 width=8)
         ->  Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..431.00 rows=2 width=8)
               Merge Key: j
               ->  Limit  (cost=0.00..431.00 rows=1 width=8)
                           ->  Partition Selector for h (dynamic scan id: 1)  (cost=10.00..100.00 rows=34 width=4)
                                 Partitions selected: 2 (out of 2)
                           ->  Sort  (cost=0.00..431.00 rows=1 width=8)
                                 Sort Key: j
                                 ->  Dynamic Table Scan on h (dynamic scan id: 1)  (cost=0.00..431.00 rows=1 width=8)
 Settings:  optimizer=on
 Optimizer status: PQO version 1.684
```
@d @oarap @hsyuan @bhuvnesh2703 @xinzweb please take a look

### DIFF
--- a/libgpopt/src/operators/CPhysical.cpp
+++ b/libgpopt/src/operators/CPhysical.cpp
@@ -934,7 +934,8 @@ CPhysical::PppsRequiredPushThruUnresolvedUnary
 	(
 	IMemoryPool *pmp,
 	CExpressionHandle &exprhdl,
-	CPartitionPropagationSpec *pppsRequired
+	CPartitionPropagationSpec *pppsRequired,
+	EPropogatePartConstraint eppcPropogate
 	)
 {
 	GPOS_ASSERT(NULL != pppsRequired);
@@ -962,7 +963,13 @@ CPhysical::PppsRequiredPushThruUnresolvedUnary
 		{
 			// push requirements to child node
 			ppimResult->AddRequiredPartPropagation(ppimReqd, ulPartIndexId, CPartIndexMap::EppraPreservePropagators);
-			(void) ppfmResult->FCopyPartFilter(pmp, ulPartIndexId, ppfmReqd);
+			if (CPhysical::EppcAllowed == eppcPropogate)
+			{
+				// for some logical operators such as limit while we push the part index map, we cannot push the constraints
+				// since they are NOT semantically equivalent. So only push the constraints when the operator asks this
+				// utility function to do so
+				(void) ppfmResult->FCopyPartFilter(pmp, ulPartIndexId, ppfmReqd);
+			}
 		}
 	}
 	
--- a/data/dxl/minidump/DonotPushPartConstThruLimit.mdp
+++ b/data/dxl/minidump/DonotPushPartConstThruLimit.mdp
@@ -1,0 +1,431 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+create table h(j int, i int) DISTRIBUTED BY (i) PARTITION by RANGE(j) (START (1) END (3) EVERY (1));
+insert into h values (1,1);
+insert into h values (2,2);
+insert into h values (2,2);
+analyze h;
+
+select * from (select * from h order by j limit 2) x where j = 2;
+-->
+
+<dxl:DXLMessage xmlns:dxl="http://greenplum.com/dxl/2010/12/">
+  <dxl:Thread Id="0">
+    <dxl:OptimizerConfig>
+      <dxl:EnumeratorConfig Id="0" PlanSamples="0" CostThreshold="0"/>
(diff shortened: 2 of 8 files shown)


libgpopt/src/operators/CPhysical.cpp
```diff
@@ -962,7 +963,13 @@ CPhysical::PppsRequiredPushThruUnresolvedUnary
 		{
 			// push requirements to child node
 			ppimResult->AddRequiredPartPropagation(ppimReqd, ulPartIndexId, CPartIndexMap::EppraPreservePropagators);
-			(void) ppfmResult->FCopyPartFilter(pmp, ulPartIndexId, ppfmReqd);
+			if (fPushPartConstraint)
+			{
+				// for some logical operators such as limit while we push the part index map, we cannot push the constraints
+				// since they are semantically equivalent. So only push the constraints when the operator is asks this
```
oarap: Do you mean "since they are **not** semantically equivalent"?

vraghavan78: Yes. sorry. Correct

hsyuan:
LGTM

PR #120, opened 2016-11-10
Insert random motion for randomly distributed children for parallel append [#133978027]

oarap:
Parallel Union currently has 2 optimization request. The second request is not satisfied
when the children of the parallel union are randomly distributed.

For the first request, if the output columns are not hashable columns the
`CPhysicalParallelUnion` requires random distribution. Since the children are randomly
distributed the `CPhysicalTableScan` already satisfied the optimization request and no
motion is created on top of `CPhysicalTableScan`.

To overcome this issue and enforce a `CPhysicalMotionRandom`, we introduce a new distribution
spec called `CDistributionSpecRandomStrict`. This distribution spec is requested replaces
regular `CDistributionSpecRandom` in the first optimization request where no output column is
redistributable.

Below is the output after this commit:
```
explain select xmin from foo union all select xmin from bar;

Physical plan:
+--CPhysicalMotionGather(master)   rows:1   width:4  rebinds:1   cost:431.000029   origin: [Grp:2, GrpExpr:3]
   +--CPhysicalParallelUnionAll   rows:1   width:4  rebinds:1   cost:431.000014   origin: [Grp:2, GrpExpr:1]
      |--CPhysicalMotionRandom   rows:1   width:38  rebinds:1   cost:431.000014   origin: [Grp:0, GrpExpr:2]
      |  +--CPhysicalTableScan "foo" ("foo")   rows:1   width:38  rebinds:1   cost:431.000007   origin: [Grp:0, GrpExpr:1]
      +--CPhysicalMotionRandom   rows:1   width:38  rebinds:1   cost:431.000014   origin: [Grp:1, GrpExpr:2]
         +--CPhysicalTableScan "bar" ("bar")   rows:1   width:38  rebinds:1   cost:431.000007   origin: [Grp:1, GrpExpr:1]

                                           QUERY PLAN
 Gather Motion 3:1  (slice3; segments: 3)  (cost=0.00..431.00 rows=1 width=4)
   ->  Append  (cost=0.00..431.00 rows=1 width=4)
         ->  Redistribute Motion 3:3  (slice2; segments: 3)  (cost=0.00..431.00 rows=1 width=4)
               ->  Table Scan on bar  (cost=0.00..431.00 rows=1 width=4)

```

### DIFF
--- a/libgpopt/include/gpopt/base/CDistributionSpec.h
+++ b/libgpopt/include/gpopt/base/CDistributionSpec.h
@@ -45,6 +45,7 @@ namespace gpopt
 				EdtSingleton,		// data is on a single segment or the master
 				EdtStrictSingleton,	// data is on a single segment or the master (derived only, only compatible with other singleton distributions)
 				EdtRandom,			// data is randomly distributed across all segments
+				EdtStrictRandom,    // same as random, used to force multiple slices for parallel union all.
 				EdtRouted,			// data is routed to a segment explicitly specified in the tuple,
 				EdtUniversal,		// data is available everywhere (derived only)
 				EdtNonSingleton,	// data can have any distribution except singleton (required only)
--- a/libgpopt/include/gpopt/base/CDistributionSpecStrictRandom.h
+++ b/libgpopt/include/gpopt/base/CDistributionSpecStrictRandom.h
@@ -1,0 +1,51 @@
+//	Greenplum Database
+//	Copyright (C) 2016 Pivotal Software, Inc.
+
+#ifndef GPOPT_CDistributionSpecStrictRandom_H
+#define GPOPT_CDistributionSpecStrictRandom_H
+
+#include "gpos/base.h"
+
+#include "gpopt/base/CDistributionSpec.h"
+#include "gpopt/base/CDistributionSpecRandom.h"
+
+namespace gpopt
+{
+    using namespace gpos;
+    
+    //---------------------------------------------------------------------------
+    //	@class:
+    //		CDistributionSpecStrictRandom
+    //
+    //	@doc:
+    //		Class for representing forced random distribution.
+    //
+    //---------------------------------------------------------------------------
+    class CDistributionSpecStrictRandom : public CDistributionSpecRandom
+    {
+    public:
+        
+        //ctor
+        CDistributionSpecStrictRandom();
+        
+        // accessor
+        virtual
+        EDistributionType Edt() const
+        {
+            return CDistributionSpec::EdtStrictRandom;
+        }
+        
+        // does this distribution match the given one
+        virtual
+        BOOL FMatch(const CDistributionSpec *pds) const;
+        
+        // does this distribution satisfy the given one
+        virtual
+        BOOL FSatisfies(const CDistributionSpec *pds) const;
+    }; // class CDistributionSpecStrictRandom
+    
+}
+
+#endif // !GPOPT_CDistributionSpecStrictRandom_H
+
+// EOF
--- a/libgpopt/src/base/CDistributionSpecRandom.cpp
+++ b/libgpopt/src/base/CDistributionSpecRandom.cpp
@@ -81,7 +81,7 @@ CDistributionSpecRandom::FSatisfies
 	const CDistributionSpec *pds
 	)
 	const
-{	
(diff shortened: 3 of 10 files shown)


libgpopt/include/gpopt/base/CDistributionSpec.h
```diff
@@ -45,6 +45,7 @@ namespace gpopt
 				EdtSingleton,		// data is on a single segment or the master
 				EdtStrictSingleton,	// data is on a single segment or the master (derived only, only compatible with other singleton distributions)
 				EdtRandom,			// data is randomly distributed across all segments
+                EdtRandomStrict,    // same as random, used to force multiple slices for parallel union all.
```
vraghavan78: Just to conform. EdtStrictRandom

libgpopt/src/base/CDistributionSpecRandom.cpp
```diff
@@ -81,8 +81,13 @@ CDistributionSpecRandom::FSatisfies
...
 	const
-{	
-	if (FMatch(pds))
+{
+    if (CDistributionSpec::EdtRandomStrict == pds->Edt())
+    {
+        return false;
+    }
+
+    if (FMatch(pds))
```
vraghavan78: this seems unclean. Why does not FMatch function handle this by saying false?

hsyuan:
what is it going to do with union on hash distributed foo and randomly distributed bar? previously I saw it is going to hash on all the hashable column for every relation.

libgpopt/include/gpopt/base/CDistributionSpecRandomStrict.h
```diff
@@ -0,0 +1,51 @@
...
+        EDistributionType Edt() const
+        {
+            return CDistributionSpec::EdtRandomStrict;
+        }
+        
+        // does this distribution match the given one
+        virtual
+        BOOL FMatch(const CDistributionSpec *pds) const;
+        
+        // does this distribution match the given one
```
hsyuan: match -> satisfy

server/src/unittest/gpopt/minidump/CPhysicalParallelUnionAllTest.cpp
```diff
@@ -22,6 +22,7 @@ static const CHAR *rgszFileNames[] =
 		"../data/dxl/minidump/CPhysicalParallelUnionAllTest/ParallelUnion-ConstTable.mdp",
 		"../data/dxl/minidump/CPhysicalParallelUnionAllTest/ParallelUnionAllWithNotEqualNumOfDistrColumns.mdp",
 		"../data/dxl/minidump/CPhysicalParallelUnionAllTest/FallBackToSerialUnion.mdp",
+        "../data/dxl/minidump/CPhysicalParallelUnionAllTest/RandomDistributedChildrenUnhashableColumns.mdp",
```
hsyuan: indent

libgpopt/src/base/CDistributionSpecRandomStrict.cpp
```diff
@@ -0,0 +1,23 @@
+//	Greenplum Database
+//	Copyright (C) 2016 Pivotal Software, Inc.
+
+#include "gpopt/base/CDistributionSpecRandomStrict.h"
+#include "gpopt/operators/CPhysicalMotionHashDistribute.h"
```
hsyuan: The 3 header files L5~L7  are not used.

oarap:
@hsyuan mixed distributions are ok. as long as there is an output column that is hashable or distributable, we always to hash distribute regardless of how the child is distributed.

oarap:
Please see the requested changes. @hsyuan @vraghavan78

libgpopt/include/gpopt/base/CDistributionSpecStrictRandom.h
```diff
@@ -0,0 +1,51 @@
...
+        EDistributionType Edt() const
+        {
+            return CDistributionSpec::EdtStrictRandom;
+        }
+        
+        // does this distribution match the given one
+        virtual
+        BOOL FMatch(const CDistributionSpec *pds) const;
+        
+        // does this distribution match the given one
```
hsyuan: match -> satisfy

PR #121, opened 2016-11-10
Fixed outer references in groupby columns [#134091245]

oarap:
In general, it's beneficial to remove outer references from groupby,
because this value is always a constant for subquery. For example:

```
select a from t where c in (select count(s.j) from s group by s.i, t.b)
```

The `t.b` can be removed safely from above SQL statement, because for
every execution of IN subquery, the `t.b` is constant.

However, this is an issue if the outer reference is the only groupby
column and there is NO additional aggregate functions used. For example:

```
select a from t where c in (select distinct t.b from s)
```

The above statement cannot be further simplified because the rewritten
query below after removing outer reference is invalid:

```
select a from t where c in (select distinct ??? from s)
```

Hence, we add additional validation in pre-processing to ensure correct
rewritten is done.

Signed-off-by: Xin Zhang <xzhang@pivotal.io>

### DIFF
--- a/libgpopt/src/operators/CExpressionPreprocessor.cpp
+++ b/libgpopt/src/operators/CExpressionPreprocessor.cpp
@@ -513,28 +513,53 @@ CExpressionPreprocessor::PexprRemoveSuperfluousOuterRefs
 
 			CLogicalGbAgg *popAgg = CLogicalGbAgg::PopConvert(pop);
 			DrgPcr *pdrgpcr = CUtils::PdrgpcrExcludeColumns(pmp, popAgg->Pdrgpcr(), pcrsOuter);
-			DrgPcr *pdrgpcrMinimal = popAgg->PdrgpcrMinimal();
-			if (NULL != pdrgpcrMinimal)
-			{
-				pdrgpcrMinimal = CUtils::PdrgpcrExcludeColumns(pmp, pdrgpcrMinimal, pcrsOuter);
-			}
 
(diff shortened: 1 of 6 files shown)


libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -513,28 +513,52 @@ CExpressionPreprocessor::PexprRemoveSuperfluousOuterRefs
...
+			CExpression *pExprProjList = (*pexpr)[1];
+			
+			// It's only valid to remove the outer reference if:
+			// the projection list is NOT empty
+			// or
+			// the outer references are NOT the ONLY Group By column
+			//
+			// For example:
+			// -- Cannot remove t.b from groupby, because this will produce invalid plan
+			// select a from t where c in (select distinct t.b from s)
```
vraghavan78: The algebrizer converts distinct into a group by. So not sure it that will cause confusion. How about:

`select a from t where c in (select  t.b from s group by t.b)`

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -513,28 +513,52 @@ CExpressionPreprocessor::PexprRemoveSuperfluousOuterRefs
...
+			//
+			// -- remove t.b from groupby is ok, because there is at least one agg function: count()
+			// select a from t where c in (select count(s.j) from s group by t.b)
+			//
+			// -- remove t.b from groupby is ok, because there is other groupby column s.j
```
vraghavan78: wrong explanation. u can remove it because there is an agg function

PR #125, opened 2016-11-21
push filter under limit [#129871531]

xinzweb:
Under situation where there is a subquery under limit and also the
columns used by the predicates inside the subquery are also referenced
outside the subquery, then `InferPredicates` under `PexprPreprocess` will
stop producing predicates under limit.

For example:
```
explain select 1 from (select * from foo where a = 1 and b = a limit 1) x;

                                           QUERY PLAN
 Result  (cost=0.00..431.00 rows=1 width=4)
   ->  Result  (cost=0.00..431.00 rows=1 width=1)
         Filter: a = 1 AND b = 1
         ->  Limit  (cost=0.00..431.00 rows=1 width=8)
               ->  Gather Motion 1:1  (slice1; segments: 1)  (cost=0.00..431.00 rows=1 width=8)
                           Filter: a = 1 AND a = b
 Settings:  optimizer=on
 Optimizer status: PQO version 1.687
(9 rows)
```

As you can see the example above, we expect to also produce `b=1` on the `Table Scan on foo`.

This is due to the reason of `InferPredicates` will NOT generate duplicated predicates if parent
operators already generate the predicates.

The basic assumption is that, all the predicates should be generated as high as possible in the
query plan, and later on predicate pushdown will move those predicates in proper location.

However, such assumption is broken when pushing predicates over limit, because it's NOT semantically
correct to push predicates outside limit to inside limit. (It's totally fine to duplicate predicates
outside limit).

The fix is actually checking the operator when deriving predicates out of constraints. If we see a
limit operator, then always generate all the predicates based on its constraint, since there won't
be any predicates pushed down through limit.

### DIFF
--- a/libgpopt/src/operators/CExpressionPreprocessor.cpp
+++ b/libgpopt/src/operators/CExpressionPreprocessor.cpp
@@ -1644,7 +1644,15 @@ CExpressionPreprocessor::PexprFromConstraints
 
 		CColRefSet *pcrsOutChild = GPOS_NEW(pmp) CColRefSet(pmp);
 		pcrsOutChild->Include(pdprelChild->PcrsOutput());
-		pcrsOutChild->Exclude(pcrsProcessed);
+		// Duplicated predicates will not be generated if parent operators already contains the predicates.
+		// if pexpr is a logical limit operator, the pcrsProcessed may contain columns that
+		// we still need to infer predicates on these columns. so don't exclude these columns.
+		// In other words, if we see a limit operator, then always generate all the predicates based on its constraint, since there won't
+		// be any predicates pushed down through limit.
+		if (COperator::EopLogicalLimit != pexpr->Pop()->Eopid())
+		{
+			pcrsOutChild->Exclude(pcrsProcessed);
+		}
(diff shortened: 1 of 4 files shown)


libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -1644,7 +1644,12 @@ CExpressionPreprocessor::PexprFromConstraints
 
 		CColRefSet *pcrsOutChild = GPOS_NEW(pmp) CColRefSet(pmp);
 		pcrsOutChild->Include(pdprelChild->PcrsOutput());
-		pcrsOutChild->Exclude(pcrsProcessed);
+		// if pexpr is a logical limit operator, the pcrsProcessed may contain columns that
+		// we still need to infer predicates on these columns. so don't exclude these columns.
+		if (COperator::EopLogicalLimit != pexpr->Pop()->Eopid())
```
vraghavan78: Hey can u elaborate on this that we have multi-phase thingy. Constraint Derivation + blah blah + ...

PR #127, opened 2016-12-14
New hint to penalize broadcast of large number of rows

dhanashreek89:
Added a guc, `optimizer_large_table_broadcast` to set the threshold on maximum number of rows to broadcast.

```
                                             QUERY PLAN
 Gather Motion 3:1  (slice2; segments: 3)  (cost=0.00..869.16 rows=1 width=16)
   ->  Hash Join  (cost=0.00..869.16 rows=1 width=16)
         Hash Cond: bar.d = foo.a
         ->  Table Scan on bar  (cost=0.00..431.69 rows=33190 width=8)
         ->  Hash  (cost=431.02..431.02 rows=100 width=8)
               ->  Broadcast Motion 3:3  (slice1; segments: 3)  (cost=0.00..431.02 rows=100 width=8)
                     ->  Table Scan on foo  (cost=0.00..431.00 rows=34 width=8)
 Settings:  optimizer=on
 Optimizer status: PQO version 1.694
```

If we set the guc to a value lower than the row count such `set optimizer_large_table_broadcast to 10;` optimizer will penalize the broadcast motion with huge cost. 

So, the new plan will look like below:
```
                                             QUERY PLAN
 Gather Motion 3:1  (slice2; segments: 3)  (cost=0.00..882.24 rows=1 width=16)
   ->  Hash Join  (cost=0.00..882.24 rows=1 width=16)
         Hash Cond: bar.d = foo.a
         ->  Redistribute Motion 3:3  (slice1; segments: 3)  (cost=0.00..433.44 rows=33190 width=8)
               Hash Key: bar.d
               ->  Table Scan on bar  (cost=0.00..431.69 rows=33190 width=8)
         ->  Hash  (cost=431.00..431.00 rows=34 width=8)
               ->  Table Scan on foo  (cost=0.00..431.00 rows=34 width=8)
 Settings:  optimizer=on; optimizer_large_table_broadcast=10
```

The broadcast plan will still be generated but not picked up.

### DIFF
--- a/libgpdbcost/src/CCostModelGPDB.cpp
+++ b/libgpdbcost/src/CCostModelGPDB.cpp
@@ -1199,7 +1199,8 @@ CCostModelGPDB::CostMotion
 
 		if(dRowsOuter > ulBroadcastThreshold)
 		{
-			costLocal = CCost(100000000000000);
+			DOUBLE ulPenalizationFactor = 100000000000000.0;
+			costLocal = CCost(ulPenalizationFactor);
 		}
 	}
 
--- a/libgpdbcost/src/CCostModelGPDB.cpp
+++ b/libgpdbcost/src/CCostModelGPDB.cpp
@@ -24,6 +24,8 @@
 #include "naucrates/statistics/CStatisticsUtils.h"
 #include "gpopt/operators/CExpression.h"
 #include "gpdbcost/CCostModelGPDB.h"
+#include "gpopt/optimizer/COptimizerConfig.h"
+#include "gpopt/engine/CHint.h"
 
 using namespace gpos;
 using namespace gpdbcost;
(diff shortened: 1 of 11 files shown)


libgpdbcost/src/CCostModelGPDB.cpp
```diff
@@ -1189,6 +1191,19 @@ CCostModelGPDB::CostMotion
...
 
+
+	if(COperator::EopPhysicalMotionBroadcast == eopid)
+	{
+		COptimizerConfig *poconf = COptCtxt::PoctxtFromTLS()->Poconf();
+		ULONG ulBroadcastThreshold = poconf->Phint()->UlBroadcastThreshold();
+
+		if(dRowsOuter > ulBroadcastThreshold)
+		{
+			costLocal = CCost(100000000000000);
```
hsyuan: CCost is double, I guess. Will the default cause overflow if we multiply with other factors?

ivannovick:
cool idea!

ivannovick:
the PR says GUC, but i don't see a change to guc.c to enable it?

hsyuan:
@ivannovick The guc is enabled here: https://github.com/greenplum-db/gpdb/pull/1443

dhanashreek89:
PRs for guc in gpdb master and gpdb4:
https://github.com/greenplum-db/gpdb/pull/1443
https://github.com/greenplum-db/gpdb4/pull/197

libgpdbcost/src/CCostModelGPDB.cpp
dhanashreek89: Good point @hsyuan . We have looked at the cost model and have not seen anywhere the cost is actually multiplied. It looks like the costs are added and accumulated through the root.

hsyuan:
LGTM

PR #128, opened 2016-12-17
[#130928821] Adding CDistributionSpecExternal for external table

dhanashreek89:
This will enable ORCA add RandomMotion or RedistributeMotion on top of the CPhysicalExternalScan.

That extra motion will redistribute rows across all segments hence improve the performance of parallel loading using external table.

This behavior is consistent with Planner behavior.
If it is a Master Only table then there is no extra motion added.


hsyuan:
LGTM!

d:
This patch set has clear signs of copy-pasta and dead code

PR #135, opened 2017-01-16
Orca should skip collapsing project elements that have set returning functions with its child [#137642207]

vraghavan78:
If the project element of a subquery has a set returning function, it should not be merged with its child since it may generate wrong results under the following conditions:

* The child's project list contains a set returning function (this was
fixed in a prior fix)
* The current project list has other set returning functions that may not be merged
with its child since it uses the childs output - see the tests added with the commit.
Splitting these project elements will change the semantic of the query
(fixed by the current fix).

[#137642207]


server/src/unittest/gpopt/minidump/CICGTest.cpp
```diff
@@ -51,14 +51,6 @@ const CHAR *rgszFileNames[] =
...
 		"../data/dxl/minidump/ConvertHashToRandomSelect.mdp",
 		"../data/dxl/minidump/ConvertHashToRandomInsert.mdp",
-		"../data/dxl/minidump/DMLCollapseProject.mdp",
-		"../data/dxl/minidump/CollapseCascadeProjects2of2.mdp",
-		"../data/dxl/minidump/CollapseCascadeProjects2of3.mdp",
-		"../data/dxl/minidump/CannotCollapseCascadeProjects.mdp",
-		"../data/dxl/minidump/CollapseProject-SetReturning.mdp",
-		"../data/dxl/minidump/CollapseProject-SetReturning-CTE.mdp",
-		"../data/dxl/minidump/RemoveUnusedProjElementsInGbAgg.mdp",
-		"../data/dxl/minidump/RemoveUnusedProjElements.mdp",
```
vraghavan78: moved:
* the collapse tests into CCollapseProjectTest
* remove column tests to CPruneColumnsTest

libgpopt/src/base/CUtils.cpp
```diff
@@ -5873,16 +5871,12 @@ CUtils::PexprCollapseProjects
...
-
-		// if both pexprPrE and pexprChildScalar have set returning functions we should not collapse
-		BOOL fBothHasSetReturningFunc = fChildProjElHasSetReturn && fParentProjElHasSetReturn;
+		BOOL fHasSetReturn = CDrvdPropScalar::Pdpscalar(pexprPrE->PdpDerive())->FHasNonScalarFunction();
 
 		pcrsUsed->Intersection(pcrsDefinedChild);
 		ULONG ulIntersect = pcrsUsed->CElements();
 
-		if (0 == ulIntersect && !fBothHasSetReturningFunc)
+		if (0 == ulIntersect && !fHasSetReturn)
```
hsyuan: Does this mean whenever we have set returning function on the current projection list, no matter the child relation has set returning function or not, we are not going to collapse the project element?

vraghavan78:
Abandoning the PR since it is too restrictive

PR #138, opened 2017-01-17
Orca should skip collapsing project elements that have set returning functions with its child columns [#137642207] 

vraghavan78:
If the project element of a subquery has a set returning function, it should not be merged with its child since it may generate wrong results under the following conditions:

* The child's project list contains a set returning function (this was fixed in a prior fix)
* The current project list has other set returning functions that may not be merged with its child since it uses the childs output - see the tests added with the commit. Splitting these project elements will change the semantic of the query (fixed by the current fix).

[#137642207]


vraghavan78:
@d @oarap @bhuvnesh2703 @hsyuan @dhanashreek89 @khannaekta please take a look.

libgpopt/include/gpopt/base/CUtils.h
```diff
@@ -74,6 +74,10 @@ namespace gpopt
 			static
 			BOOL FHasAllDefaultComparisons(const IMDType *pmdtype);
 
+			//	append the expressions in the source array to destination array
+			static
+			void AppendArrayExpr(DrgPexpr *pdrgpexprSrc, DrgPexpr *pdrgpexprDest);
```
oarap: Can we change the order of arguments? Usually in append, dst is the first and src is the second. `char *strcat(char *dst, const char *src)`

If orca follows the other way, that's fine too.

libgpopt/src/base/CUtils.cpp
```diff
@@ -5943,6 +5975,32 @@ CUtils::PexprCollapseProjects
 
 //---------------------------------------------------------------------------
 //	@function:
+//		CUtils::AppendArrayExpr
+//
+//	@doc:
+//		Append the expressions in the source array to destination array
```
oarap: Drop the `the`

vraghavan78:
Thanks @oarap I will do that.

libgpopt/include/gpopt/base/CUtils.h
vraghavan78: In Orca, we have it the otherway, for example:
{code}
static
CColRef *PcrMap(CColRef *pcrSource, DrgPcr *pdrgpcrSource, DrgPcr *pdrgpcrTarget);
{code}

vraghavan78:
@oarap how is the following commit message:
---

[#137642207]

If the project element of a subquery has a set returning function, it should not be merged with its child since it may generate wrong results under the following conditions:

* The child's project list contains a set returning function (this was fixed in a prior fix)
* The current project list has other set returning functions that may not be merged with its child since it uses the childs output - see the example below. Splitting these project elements will change the semantic of the query (fixed by the current fix).


Example SQL:

select generate_series(0,2) rn, unnest(arr) cnt_div from (select ARRAY['0','1','2', '3'] as arr) arra;

The SQL has two project nodes:
* The top project node with two set returning function.
* The lower ARRAY function over a constant table get.

```

+--CLogicalProject
   |--CLogicalProject
   |  |--CLogicalConstTableGet Columns: ["" (0)] Values: [(1)]
   |  +--CScalarProjectList
   |     +--CScalarProjectElement "arr" (1)
   |        +--CScalarArray: {eleMDId: (25,1.0), arrayMDId: (1009,1.0)}
   |           |--CScalarConst (161088044.000)
   |           |--CScalarConst (161096236.000)
   |           +--CScalarConst (161104428.000)
   +--CScalarProjectList
      |--CScalarProjectElement "rn" (2)
      |  +--CScalarFunc (generate_series)
      |     |--CScalarConst (0)
      |     +--CScalarConst (2)
      +--CScalarProjectElement "cnt_div" (3)
         +--CScalarFunc (unnest)
            +--CScalarIdent "arr" (1)
```

The top project node has two set returning functions:
* generate_series - that has no dependence on the child's project list so can be collapsed.
* unnest(arr) - takes as input `arr` which comes from child's project list so cannot be collapsed.

Before this fix, we collapsed the first function and not the second. Like below.


```
+--CLogicalProject
   |--CLogicalProject
   |  |--CLogicalConstTableGet Columns: ["" (0)] Values: [(1)]
   |  +--CScalarProjectList
   |     |--CScalarProjectElement "rn" (2)
   |     |  +--CScalarFunc (generate_series)
   |     |     |--CScalarConst (0)
   |     |     +--CScalarConst (2)
   |     +--CScalarProjectElement "arr" (1)
   |        +--CScalarArray: {eleMDId: (25,1.0), arrayMDId: (1009,1.0)}
   |           |--CScalarConst (161088044.000)
   |           |--CScalarConst (161096236.000)
   |           +--CScalarConst (161104428.000)
   +--CScalarProjectList
      +--CScalarProjectElement "cnt_div" (3)
         +--CScalarFunc (unnest)
            +--CScalarIdent "arr" (1)
```

This causes wrong result since it changes the SQL semantics when we have multiple set returning functions in the a single project list.
It is all or nothing.

oarap:
It looks great. I think it is good to go.

PR #143, opened 2017-02-15
[#139042295] Dedup constraints in linear time

oarap:
GP Orca rearranges constraints list where a constrint only refers a single
column reference as much as possible. In order to do that the previous implementation
traverse the list for every single constraint and that results in O(n^2) (quadratic)
running time.

If the number of constraints in the list is higher than some threashold
(hardcoded as 5 for now) it results in significant performance degredation.

This commit uses hash map to create a map from column referances to constraints where
a constraint only refers a single column. This will reduce the quadratic running time
of deduplication process to a linear running time. However, creating a hashmap for short
list of constraints also results in performnce degradation.

### DIFF
--- a/libgpopt/src/base/CConstraint.cpp
+++ b/libgpopt/src/base/CConstraint.cpp
@@ -10,8 +10,12 @@
 //---------------------------------------------------------------------------
 
 #include "gpos/base.h"
+#include "gpos/common/CAutoRef.h"
 
 #include "gpopt/base/CUtils.h"
+#include "gpopt/base/IColConstraintsMapper.h"
+#include "gpopt/base/CColConstraintsArrayMapper.h"
+#include "gpopt/base/CColConstraintsHashMapper.h"
 #include "gpopt/base/CColRefSetIter.h"
 #include "gpopt/base/CConstraint.h"
 #include "gpopt/base/CConstraintInterval.h"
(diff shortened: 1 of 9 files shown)


oarap:
@d @vraghavan78 @bhuvnesh2703 @hsyuan Please review

libgpopt/src/base/CConstraint.cpp
```diff
@@ -668,8 +668,15 @@ CConstraint::PdrgpcnstrDeduplicate
 	DrgPcnstr *pdrgpcnstrNew = GPOS_NEW(pmp) DrgPcnstr(pmp);
 
 	CColRefSet *pcrsDeduped = GPOS_NEW(pmp) CColRefSet(pmp);
+	HMColConstr *phmcolconstr = NULL;
 
 	const ULONG ulLen = pdrgpcnstr->UlLength();
+
+	if(ulLen >= 5)
```
danielgustafsson: A variable or constant with a descriptive name would be good to avoid these hardcoded checks never go out of sync. A comment on why there is a threshold would also be good.

libgpopt/src/base/CConstraint.cpp
```diff
@@ -691,8 +698,18 @@ CConstraint::PdrgpcnstrDeduplicate
...
+		DrgPcnstr *pdrgpcnstrCol = NULL;
+		if(ulLen < 5)
+		{
+			// get all constraints from the input array that reference this column
+			pdrgpcnstrCol = PdrgpcnstrOnColumn(pmp, pdrgpcnstr, pcr, true /*fExclusive*/);
+		}
+		else
+		{
+			GPOS_ASSERT(phmcolconstr);
+			pdrgpcnstrCol = phmcolconstr->PtLookup(pcr);
```
danielgustafsson: As far as I can see there is no destruction, I assume it will be cleaned up at the end of planning. Is there a potential scenario where keeping this around without releasing the memory would be harmful?

oarap: Totally right, i fixed the issue.

d: The repeated check on `ulLen < 5` is a clear smell to me. We should be able to extract the lookup logic behind a consistent interface, no?

PR #145, opened 2017-02-15
Fix bug that ORCA cannot do DPE for array coerce predicate

hsyuan:
Before this patch, Orca can't extract the array elements inside ScalarArrayCmp if the array is wrapped with a ArrayCoerceExpr. This patch tries to extract array inside, let Orca be able to expand the array into disjunctions and do DPE for predicate with array coerce expression.


d:
What would happen if the array coerce happens in the form of a function call (à la Postgres 8.2)?

hsyuan:
I didn't get it. Orca only support Ident comparing with Constant Array.

d:
To follow up on https://github.com/greenplum-db/gporca/pull/145#issuecomment-280377165 , here's the scenario when peeking under a cast blindly will generate a bad plan:

```SQL
CREATE TYPE jesse as (x int2, y int2);
CREATE FUNCTION jesse2int(jesse) RETURNS int as ' select $1.y::int ; ' LANGUAGE SQL STRICT IMMUTABLE;
CREATE CAST (jesse as int) WITH FUNCTION jesse2int(jesse) AS IMPLICIT;

CREATE TABLE omer (a int, pk int) partition by range (pk) ( end (2), start (2));
EXPLAIN SELECT * FROM omer WHERE pk = ANY (ARRAY['(1,2)'::jesse,'(2,3)'::jesse]);
```

PR #146, opened 2017-02-19
Improve optimization performance for large IN const array

hsyuan:
When there is large number of values in the array or IN list, Orca spent a lot of time on optimization. The reason is that every const item in the array is used to create a group and group expression in the MEMO structure. For scalar const values in the array, it is not necessary to create the MEMO group for it. We create a new scalar operator `CScalarConstArray` to treat the constant values in the children expression as its attribute, which avoids the MEMO group creation for constant values to achieve performance improvement.

Made in SFO->PEK UA888 Boeing 747 aircraft.

Signed-off-by: Omer Arap <oarap@pivotal.io>
Signed-off-by: Xin Zhang <xzhang@pivotal.io>
[#132930893]

### DIFF
--- a/libgpopt/src/base/CConstraint.cpp
+++ b/libgpopt/src/base/CConstraint.cpp
@@ -19,6 +19,7 @@
 #include "gpopt/base/CConstraintDisjunction.h"
 #include "gpopt/base/CConstraintNegation.h"
 #include "gpopt/operators/CScalarIdent.h"
+#include "gpopt/operators/CScalarConstArray.h"
 #include "gpopt/operators/CScalarArrayCmp.h"
 #include "gpopt/optimizer/COptimizerConfig.h"
 #include "gpopt/operators/CPredicateUtils.h"
@@ -102,7 +103,13 @@ CConstraint::PcnstrFromScalarArrayCmp
 		IMDType::ECmpType ecmpt = CUtils::Ecmpt(popScArrayCmp->PmdidOp());
 		CExpression *pexprArray = (*pexpr)[1];
 
-		const ULONG ulArity = pexprArray->UlArity();
+		ULONG ulArity = pexprArray->UlArity();
+		CScalarConstArray *popScArray = NULL;
+		if (CUtils::FScalarConstArray(pexprArray))
+		{
+			popScArray = CScalarConstArray::PopConvert(pexprArray->Pop());
+			ulArity = popScArray->UlSize();
+		}
 
 		// When array size exceeds the threshold, don't expand it into a DNF
 		COptimizerConfig *poconf = COptCtxt::PoctxtFromTLS()->Poconf();
@@ -117,9 +124,12 @@ CConstraint::PcnstrFromScalarArrayCmp
 
 		for (ULONG ul = 0; ul < ulArity; ul++)
 		{
-			GPOS_ASSERT(CUtils::FScalarConst((*pexprArray)[ul]) && "expecting a constant");
+			COperator *pop = popScArray ?
+									popScArray->PopConstAt(ul) :
+									(*pexprArray)[ul]->Pop();
+			GPOS_ASSERT(COperator::EopScalarConst == pop->Eopid() && "expecting a constant");
 
-			CScalarConst *popScConst = CScalarConst::PopConvert((*pexprArray)[ul]->Pop());
+			CScalarConst *popScConst = CScalarConst::PopConvert(pop);
 			CConstraintInterval *pci =  CConstraintInterval::PciIntervalFromColConstCmp(pmp, pcr, ecmpt, popScConst);
 			pdrgpcnstr->Append(pci);
 		}
--- a/libgpopt/src/base/CDatumSortedSet.cpp
+++ b/libgpopt/src/base/CDatumSortedSet.cpp
@@ -4,6 +4,7 @@
 #include "gpopt/operators/COperator.h"
 #include "gpopt/base/CDatumSortedSet.h"
 #include "gpopt/operators/CScalarConst.h"
+#include "gpopt/operators/CScalarConstArray.h"
 #include "gpopt/base/CUtils.h"
 #include "gpos/common/CAutoRef.h"
 
@@ -19,15 +20,38 @@ CDatumSortedSet::CDatumSortedSet
 	DrgPdatum(pmp),
 	m_fIncludesNull(false)
 {
-	GPOS_ASSERT(0 < pexprArray->UlArity());
-	GPOS_ASSERT(COperator::EopScalarArray == pexprArray->Pop()->Eopid());
+	GPOS_ASSERT(CUtils::FScalarArray(pexprArray) ||
+			CUtils::FScalarConstArray(pexprArray) ||
+			CUtils::FScalarArrayCoerce(pexprArray));
 
-	const ULONG ulArrayExprArity = pexprArray->UlArity();
+	if (CUtils::FScalarArrayCoerce(pexprArray))
+	{
+		pexprArray = (*pexprArray)[0];
+	}
+
+	ULONG ulArrayExprArity = pexprArray->UlArity();
+	CScalarConstArray *popScArray = NULL;
+	if (CUtils::FScalarConstArray(pexprArray))
+	{
+		popScArray = CScalarConstArray::PopConvert(pexprArray->Pop());
+		ulArrayExprArity = popScArray->UlSize();
+	}
+
+	GPOS_ASSERT(0 < ulArrayExprArity);
 
 	gpos::CAutoRef<DrgPdatum> aprngdatum(GPOS_NEW(pmp) DrgPdatum(pmp));
 	for (ULONG ul = 0; ul < ulArrayExprArity; ul++)
 	{
-		CScalarConst *popScConst = CScalarConst::PopConvert((*pexprArray)[ul]->Pop());
+		CScalarConst *popScConst = NULL;
+		if (NULL == popScArray)
+		{
+			popScConst = CScalarConst::PopConvert((*pexprArray)[ul]->Pop());
+		}
+		else
+		{
+			popScConst = popScArray->PopConstAt(ul);
+		}
+
 		IDatum *pdatum = popScConst->Pdatum();
 		if (pdatum->FNull())
 		{
--- a/libgpopt/src/base/CUtils.cpp
+++ b/libgpopt/src/base/CUtils.cpp
@@ -3960,6 +3960,24 @@ CUtils::FScalarArray
 
 //---------------------------------------------------------------------------
 //	@function:
+//		CUtils::FScalarConstArray
+//
+//	@doc:
+//		 Check if given expression is a scalar const array
+//
+//---------------------------------------------------------------------------
+BOOL
+CUtils::FScalarConstArray
+	(
+	CExpression *pexpr
+	)
+{
+	return (COperator::EopScalarConstArray == pexpr->Pop()->Eopid());
+}
+
+
+//---------------------------------------------------------------------------
+//	@function:
 //		CUtils::FScalarArrayCoerce
 //
 //	@doc:
--- a/libgpopt/src/operators/CExpressionPreprocessor.cpp
+++ b/libgpopt/src/operators/CExpressionPreprocessor.cpp
@@ -1728,6 +1728,15 @@ CExpressionPreprocessor::PexprPruneEmptySubtrees
 	return GPOS_NEW(pmp) CExpression(pmp, pop, pdrgpexpr);
 }
 
+//---------------------------------------------------------------------------
+//	@function:
+//		CExpressionPreprocessor::PexprCollapseLargeArray
+//
+//	@doc:
+// 		If it's a scalar array of all CScalarConst, collapse it into an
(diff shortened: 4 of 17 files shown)


vraghavan78:
What is the performance impact?

hsyuan:
The optimization time goes down from 2300ms to 330ms for query that has a array with 20K constant values. Planner spent 200ms on that query.

Eulerizeit:
Nice!

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -1728,6 +1728,72 @@ CExpressionPreprocessor::PexprPruneEmptySubtrees
...
 
+//---------------------------------------------------------------------------
+//	@function:
+//		CExpressionPreprocessor::PexprCollapseConstArray
+//
+//	@doc:
+// 		If it's a scalar array of all CScalarConst, collapse it into an
+//		CScalarConstArray to reduce MEMO groups.
+//
+//---------------------------------------------------------------------------
```
vraghavan78: remove this padding, since we talked about not adding fluff

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -1728,6 +1728,72 @@ CExpressionPreprocessor::PexprPruneEmptySubtrees
...
+CExpressionPreprocessor::PexprCollapseConstArray
+(
+	IMemoryPool *pmp,
+	CExpression *pexpr
+)
+{
+	GPOS_ASSERT(NULL != pexpr);
+
+	const ULONG ulArity = pexpr->UlArity();
+
```
vraghavan78:
```
if (!CUtils::FScalarArray(pexpr))
{
      pexpr->AddRef();

      return pexpr;
}
```

That is cleaner, IMHO

hsyuan: but you can't simply return without processing children.

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -1728,6 +1728,72 @@ CExpressionPreprocessor::PexprPruneEmptySubtrees
...
+	if (CUtils::FScalarArray(pexpr))
+	{
+		BOOL fAllConsts = true;
+
+		for (ULONG ul = 0; fAllConsts && ul < ulArity; ul++)
+		{
+			fAllConsts = CUtils::FScalarConst((*pexpr)[ul]);
+		}
```
vraghavan78: This is O(2N), can we squash this?

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -1728,6 +1728,72 @@ CExpressionPreprocessor::PexprPruneEmptySubtrees
...
+			IMDId *pmdidElem = psArray->PmdidElem();
+			IMDId *pmdidArray = psArray->PmdidArray();
+			pmdidElem->AddRef();
+			pmdidArray->AddRef();
+
+			CScalarConstArray *pConstArray = GPOS_NEW(pmp) CScalarConstArray(pmp, pmdidElem, pmdidArray, psArray->FMultiDimensional(), pdrgpconst);
+			return GPOS_NEW(pmp) CExpression(pmp, pConstArray);
+		}
+	}
+
```
vraghavan78: if we come here, can we just not do this:

```
pexpr->AddRef();

return pexpr;
```

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -2228,9 +2294,14 @@ CExpressionPreprocessor::PexprPreprocess
 
 	CAutoTimer at("\n[OPT]: Expression Preprocessing Time", GPOS_FTRACE(EopttracePrintOptStats));
 
+	// (0) collapse CScalarArray with constant values to CScalarConstArray
+	CExpression *pexprConstArray = PexprCollapseConstArray(pmp, pexpr);
+	GPOS_CHECK_ABORT;
```
vraghavan78: in the future no (-1) step please!

libgpopt/src/operators/CExpressionPreprocessor.cpp
oarap: Yes, it is recursive.

libgpopt/src/operators/CPredicateUtils.cpp
```diff
@@ -1544,11 +1544,16 @@ CPredicateUtils::FCompareIdentToConstArray
 
 	if (!CUtils::FScalarArrayCmp(pexpr) ||
 		!CUtils::FScalarIdent((*pexpr)[0]) ||
-		!CUtils::FScalarArray((*pexpr)[1]))
```
vraghavan78: should the comment be fixed. what is the difference between constant array and scalar array

libgpopt/src/operators/CExpressionPreprocessor.cpp
hsyuan: yes, sure. in Computer Science, we always start from 0.

libgpopt/src/operators/CPredicateUtils.cpp
```diff
@@ -1544,11 +1544,16 @@ CPredicateUtils::FCompareIdentToConstArray
...
+		(!CUtils::FScalarArray((*pexpr)[1]) && !CUtils::FScalarConstArray((*pexpr)[1])))
 	{
 		return false;
 	}
 
+	if (CUtils::FScalarConstArray((*pexpr)[1]))
+	{
+		return true;
+	}
+
```
vraghavan78: Below, we iterate over the list to make sure we have all constants. You use this in CExpressionPreprocessor::PexprCollapseConstArray.

Can we refactor this into a utility functions?

libgpopt/src/operators/CExpressionPreprocessor.cpp
oarap: 2 for loops makes it look like it is 2N but actually it is not. If you process all the elements you will do it no matter what. If A and B are single unit of jobs done in both loops, this implementation does it AAAAABBBBB. It won't change if we do it in one single shot and make it ABABABABAB and we will even allocate unnecessary space if the are not all consts.

libgpopt/src/translate/CTranslatorExprToDXL.cpp
```diff
@@ -6705,6 +6706,58 @@ CTranslatorExprToDXL::PdxlnArray
...
+										pmdidArray,
+										pConstArray->FMultiDimensional()
+										)
+						);
+
+	DrgPconst *pConsts = pConstArray->PConsts();
+	const ULONG ulArity = pConsts->UlLength();
+	for (ULONG ul = 0; ul < ulArity; ul++)
+	{
+		CScalarConst *popScConst = (*pConsts)[ul];
```
khannaekta: This [loop body](https://github.com/greenplum-db/gporca/blob/2bc5ac84c4d145a5987abf96e1294715c8ee22bb/libgpopt/src/translate/CTranslatorExprToDXL.cpp#L6747-L6753) looks like copy-pasta. We should not allow this kind of duplication here. Consider extracting the appropriate code.

libgpopt/src/base/CConstraint.cpp
```diff
@@ -117,9 +124,12 @@ CConstraint::PcnstrFromScalarArrayCmp
...
 		for (ULONG ul = 0; ul < ulArity; ul++)
 		{
-			GPOS_ASSERT(CUtils::FScalarConst((*pexprArray)[ul]) && "expecting a constant");
+			COperator *pop = popScArray ?
+									popScArray->PopConstAt(ul) :
+									(*pexprArray)[ul]->Pop();
+			GPOS_ASSERT(COperator::EopScalarConst == pop->Eopid() && "expecting a constant");
 
-			CScalarConst *popScConst = CScalarConst::PopConvert((*pexprArray)[ul]->Pop());
+			CScalarConst *popScConst = CScalarConst::PopConvert(pop);
```
vraghavan78: I see what you're doing here. Little messy

libgpopt/src/operators/CExpressionPreprocessor.cpp
hsyuan: Can't agree more.

libgpopt/src/operators/CExpressionPreprocessor.cpp
vraghavan78: my bad!

libgpopt/src/operators/CScalarConstArray.cpp
```diff
@@ -0,0 +1,155 @@
...
+				if (!popConst1->FMatch(popConst2))
+				{
+					return false;
+				}
+			}
+			return true;
+		}
+	}
+
+	return false;
```
vraghavan78: Do we need to match it arrays? Seems expensive.

khannaekta: That's necessary for correctness. Also, that's the price we pay for having big operators, no? OTOH, this reduces the number of groups significantly, which cuts down search time.

hsyuan: yes, we need. think about this query: 
```sql 
select * from foo where a in ('a','b') or (a = 'c' and b in ('c', 'd'))
```
all the cols are `text` type.

You can see the plan diff with and without comparing the constants.

libgpopt/src/base/CConstraint.cpp
hsyuan: yea, need cleaner way.

libgpopt/src/base/CUtils.cpp
```diff
@@ -3960,6 +3960,24 @@ CUtils::FScalarArray
 
 //---------------------------------------------------------------------------
 //	@function:
+//		CUtils::FScalarConstArray
+//
+//	@doc:
+//		 Check if given expression is a scalar const array
+//
+//---------------------------------------------------------------------------
+BOOL
```
vraghavan78: remove the whole comment formatting

libgpopt/src/operators/CExpressionPreprocessor.cpp
vraghavan78: ok. thanks for explaining

libgpopt/src/operators/CExpressionPreprocessor.cpp
vraghavan78: my bad

libgpopt/src/operators/CScalarConstArray.cpp
vraghavan78: sorry I did not state it clearly, is it okay to say no two arrays match. So match will always give u false. This will avoid big checks.

libgpopt/src/base/CDatumSortedSet.cpp
```diff
@@ -19,15 +20,38 @@ CDatumSortedSet::CDatumSortedSet
 	DrgPdatum(pmp),
 	m_fIncludesNull(false)
 {
-	GPOS_ASSERT(0 < pexprArray->UlArity());
-	GPOS_ASSERT(COperator::EopScalarArray == pexprArray->Pop()->Eopid());
+	GPOS_ASSERT(CUtils::FScalarArray(pexprArray) ||
+			CUtils::FScalarConstArray(pexprArray) ||
+			CUtils::FScalarArrayCoerce(pexprArray));
```
khannaekta: There is no clear justification for including `ArrayCoerce` here. This is speculative coding.

hsyuan: correct

libgpopt/src/operators/CExpressionPreprocessor.cpp
khannaekta: These lines of determining whether we have a `Array` expression with all `ScalarConst` child is a [verbatim copy from `PreciateUtils.cpp`](https://github.com/greenplum-db/gporca/blob/aa466bfecc72731f22366380673272280f0d8268/libgpopt/src/operators/CPredicateUtils.cpp#L1555-L1561). We should extract that logic instead of growing our already sad pile of copy-pasta.

libgpopt/src/base/CConstraint.cpp
```diff
@@ -102,7 +103,13 @@ CConstraint::PcnstrFromScalarArrayCmp
 		IMDType::ECmpType ecmpt = CUtils::Ecmpt(popScArrayCmp->PmdidOp());
 		CExpression *pexprArray = (*pexpr)[1];
 
-		const ULONG ulArity = pexprArray->UlArity();
+		ULONG ulArity = pexprArray->UlArity();
+		CScalarConstArray *popScArray = NULL;
+		if (CUtils::FScalarConstArray(pexprArray))
+		{
```
khannaekta: (jz + ek) The litany of `if` switches and the propagation of assumption throughout the codebase is very concerning. We should take this opportunity to refactor the code. What emerged as a theme seems to be we want to iterate through elements of an array. If that's consistent everywhere, we should be able to put the difference behind an interface.

hsyuan:
Closed this PR, see the new one: https://github.com/greenplum-db/gporca/pull/147

PR #148, opened 2017-03-01
140601033 equiv class intersect optimization

oarap:
This pull request introduces a better algorithm for generating the intersection
of two sets of equivalence classes. The input sets are arrays of column
reference sets which denotes individual equivalence classes.

Since the equivalence classes are disjoint by definition, the implementation
could be reduced from quadratic time to the linear time by utilizing an HashMap.

We build a hash-map using the columns as keys and equivalence classes as the
value.

Below is a sample input and output scenario.

`Classes1`: `[(a,b),(c,d,e),(f,g)]`
`Classes2`: `[(a),(b,c,d),(e),(f,g)]`

The hashmap after first iteration on `Classes1`:
`a->(a,b)`
`b->(a,b)`
`c->(c,d,e)`
`d->(c,d,e)`
`e->(c,d,e)`
`f->(f,g)`
`g->(f,g)`

In the probe iteration, once we detect an intersection, we replace the intersected
columns' entry in the HashMap with an empty set to avoid duplication.

In the probe iteration, first `(a)` will be read first and column `a` will be
probed. `map[a] = (a,b)`. So the intersection of `(a,b)` and `(a)` is `(a)` and it will
be added to result list. `a`'s entry in the map is emptied. Hash map after this step
will look like below:
`a->()`
`b->(a,b)`
`c->(c,d,e)`
`d->(c,d,e)`
`e->(c,d,e)`
`f->(f,g)`
`g->(f,g)`

Then, `(b,c,d)` will be processed. First column `b` is used to probe the hash map and
it will return `(a,b)`. The intersection will be `(b)` and it will be appended to the
result list. `b`'s entry in the map is emptied. Hash Map after this step:
`a->()`
`b->()`
`c->(c,d,e)`
`d->(c,d,e)`
`e->(c,d,e)`
`f->(f,g)`
`g->(f,g)`

Then column `c` will be looked up and hash map will return `(c,d,e)`. The intersaction
of `(b,c,d)` and `(c,d,e)` will be `(c,d)` and it will be added to result list. `c` and
`d`'s entry is also emptied in the Hash Map after this step and it will be like below:
`a->()`
`b->()`
`c->()`
`d->()`
`e->(c,d,e)`
`f->(f,g)`
`g->(f,g)`

So, when we look up for column `d` at this point which is the last column of `(b,c,d)`,
since `d`'s entry is blank in the hash map, it will not produce a duplicate intersection
result of `(c,d)` again.

The final result will be `[(a),(b),(c,d),(e),(f,g)]`


vraghavan78:
What is your performance impact of this change?

libgpopt/src/base/CUtils.cpp
```diff
@@ -3409,8 +3410,8 @@ CUtils::PdrgpcrsMergeEquivClasses
 //		CUtils::PdrgpcrsIntersectEquivClasses
 //
 //	@doc:
-//		Intersect 2 arrays of equivalence classes. This is accomplished by doing
-//		pairwise intersection between elements of the two arrays
+//		Intersect 2 arrays of equivalence classes. This is accomplished by using
+//		a hashmap to find intersects between two arrays
```
vraghavan78: of column references.

libgpopt/src/base/CUtils.cpp
```diff
@@ -6047,4 +6100,34 @@ INT CUtils::IDatumCmp
...
+	)
+{
+	const ULONG ulLen = pdrgpcrs->UlLength();
+
+	// nothing to check
+	if (ulLen == 0)
+	{
+		return true;
+	}
+
```
vraghavan78: can't u implement this also via a hashmap?

server/src/unittest/gpopt/base/CEquivalenceClassesTest.cpp
```diff
@@ -0,0 +1,236 @@
+//---------------------------------------------------------------------------
+//	Greenplum Database
+//	Copyright (C) 2009 Greenplum, Inc.
```
vraghavan78: fix the copyright

server/src/unittest/gpopt/base/CEquivalenceClassesTest.cpp
```diff
@@ -0,0 +1,236 @@
...
+//
+//	@filename:
+//		CEquivalenceClassesTest.cpp
+//
+//	@doc:
+//		Tests for CColRefSet
```
vraghavan78: fix the comment

server/include/unittest/gpopt/base/CEquivalenceClassesTest.h
```diff
@@ -0,0 +1,30 @@
+//---------------------------------------------------------------------------
+//	Pivotal Software, Inc
+//	Copyright (C) 2017 Pivotal Software, Inc
+
```
vraghavan78: why only half of the header?

khannaekta: Let's just drop the top bar

server/include/unittest/gpopt/base/CEquivalenceClassesTest.h
```diff
@@ -0,0 +1,30 @@
...
+#ifndef GPOS_CEquivalenceClassesTest_H
```
khannaekta: wrong macro. by convention it should be `GPOPT_CEquivalenceClassesTest_H`

libgpopt/src/base/CUtils.cpp
oarap: done

PR #149, opened 2017-03-08
Refactor Expr to DXL translator for list partition selector predicates

hsyuan:
Previously when the final plan is translated into DXL, Orca uses range predicates to represent the values of the list partition (range start and end value equal a single list value), which is error-prone and redundant for the query executor.

In this patch, we use the following way to translate the predicates of the list partition selector in DXL (range based partition remain unchanged):
1. PK op Scalar -> Scalar reverse(op) Any(PartListValues)
e.g. 
For partition selector predicate `pk1 < 5`, will be translated to `ArrayComp(5 > Any(PartListValues))`, which means as long as any value of list partition values is less than 5, the partition will be selected.
2. PK is (not) NULL -> PartListNullTest
3. Propagation Expression ->
Const1 = Any(PartListValues) or Const2 = Any(PartListValues) ...


oarap:
Is it possible to make the history cleaner and more reviewable? I would initially suggest to put all test file changes into one commit so that the reviewers can focus on the core development.

hsyuan:
cleaned commit history as requested.

hsyuan:
It is kind of messy PR, I have to admit. IMHO, only 2 files needs your attention: `CTranslatorExprToDXLUtils.cpp` and `CTranslatorExprToDXL.cpp`

libgpopt/src/translate/CTranslatorExprToDXL.cpp
```diff
@@ -4966,37 +5003,84 @@ CTranslatorExprToDXL::PdxlnScCmpPartKey
 	GPOS_ASSERT(NULL != pexprPartKey && NULL != pexprOther);
 	GPOS_ASSERT(IMDType::EcmptOther != ecmpt);
 
+	// For varchar(1043) like columns, GPDB will automatically cast to text(25).
+	// So we need special care to get the actual type from the expression.
+	pmdidTypePartKey = CScalar::PopConvert(pexprPartKey->Pop())->PmdidType();
```
d: No no no!
This is *throwing away* the original type of the partition key, which is not only logically wrong, but also leads to wrong type information in the plan.

Which query motivated this? If we drop this line what problem do you foresee?

libnaucrates/src/CDXLUtils.cpp
```diff
@@ -1973,6 +1973,34 @@ CDXLUtils::PstrSerialize
...
+			// last element: do not print a comma
+			a_pstr->AppendFormat(GPOS_WSZ_LIT("%c"), tValue);
+		}
+		else
+		{
+			a_pstr->AppendFormat(GPOS_WSZ_LIT("%c%ls"), tValue, CDXLTokens::PstrToken(EdxltokenComma)->Wsz());
+		}
+	}
+
+	return a_pstr.PtReset();
```
hsyuan: I didn't get it. Can you give an example?

libnaucrates/include/naucrates/md/CMDRelationGPDB.h
```diff
@@ -131,6 +134,7 @@ namespace gpmd
 				DrgPmdcol *pdrgpmdcol,
 				DrgPul *pdrgpulDistrColumns,
 				DrgPul *pdrgpulPartColumns,
+				DrgPsz *m_pdrgpszPartTypes,
```
hsyuan: nice catch.

libgpopt/src/translate/CTranslatorExprToDXL.cpp
hsyuan: as you can see from the comment, any list part table with varchar part key

hsyuan:
The partition type is enum. even though it's specified by char, but it's viewed as int.

libgpopt/src/translate/CTranslatorExprToDXL.cpp
d: query

libnaucrates/src/CDXLUtils.cpp
d: Meaning: stop using the smart pointer (`AutoP`)

libgpopt/src/translate/CTranslatorExprToDXL.cpp
d: That's because you missed the cast from the partition key side of the comparison. I added comments inline

hsyuan: correct, that's what I want to avoid.

libgpopt/src/translate/CTranslatorExprToDXL.cpp
```diff
@@ -4966,37 +5003,84 @@ CTranslatorExprToDXL::PdxlnScCmpPartKey
...
+								);
 	}
+	else // list partition
+	{
+		ecmpt = CPredicateUtils::EcmptReverse(ecmpt);
 
-	CTranslatorExprToDXLUtils::ExtractCastMdids(pexprNewPartKey->Pop(), &pmdidTypeCastExpr, &pmdidCastFunc);
+		// If the pexprOther is not comparable with pexprPartKey, but can be casted to pexprPartKey,
+		// and not yet casted, then we add a cast on top of pexprOther.
+		if (!CMDAccessorUtils::FCmpExists(m_pmda, pmdidTypeOther, pmdidTypePartKey, ecmpt))
```
d: This is wrong for several reasons:

0. It is casting the "other" side instead of the side of the partition key
0. This combined with the check for cast on line 5055, missed the case when the partition key expression is given in the form of a cast, hence will lead to a double cast.

The fix for this is not to introduce a *fake* `pmdidTypePartKey` in line 5008 above, but instead, stay consistent with the range partition case: i.e. we add cast to the partition key side.

hsyuan: Can you give an example that can have double cast? The plan can't be generated, as far as I know.

d: If we remove the fake `pmdidTypePartKey` assignment from line 5008, wouldn't this cast the right hand side (again)?

d:
The handling of cast here is wrong. I posted a small repro over at https://github.com/greenplum-db/gpdb/pull/2076#pullrequestreview-29344763 .

In both list and range partitions, there are cases where we will generate a plan that wrongly advertises the value of the partition key (type of `PARTBOUNDEXPR` and element type of `PARTLISTEXCPR`, respectively).

In general, instead of rigging the executor to increasingly relax the expectation of our plan's correctness, we should fix it in ORCA:

hsyuan:
Well, if you want to support GPDB 5 only,  I can modify it that way.

Sent from my iPhone

libgpopt/src/translate/CTranslatorExprToDXL.cpp
hsyuan: no, I don't think so.

d:
To recap our discussion offline, array coerce expressions _are portable_ even to GPDB 4. The translator in a database without Array Coerce Expression needs to translate it to a FUNCEXPR with the corresponding `resulttype` and `fnoid`

hsyuan:
Updated code to address comments.

libnaucrates/src/CDXLUtils.cpp
```diff
@@ -1975,6 +1975,41 @@ CDXLUtils::PstrSerialize
 
 //---------------------------------------------------------------------------
 //	@function:
+//		CDXLUtils::PstrSerializeSz
+//
+//	@doc:
+//		Serialize a list of chars into a comma-separated string
+//
+//---------------------------------------------------------------------------
```
bhuvnesh2703: could you please remove this type of comment and have 
```
// serialize a list of chars into a comma-separated string
```

libnaucrates/src/CDXLUtils.cpp
```diff
@@ -1973,6 +1973,34 @@ CDXLUtils::PstrSerialize
...
+CDXLUtils::PstrSerializeSz
+	(
+	IMemoryPool *pmp,
+	const DrgPsz *pdrgsz
+	)
+{
+	CWStringDynamic *pstr = GPOS_NEW(pmp) CWStringDynamic(pmp);
+
+	ULONG ulLength = pdrgsz->UlLength();
+	for (ULONG ul = 0; ul < ulLength; ul++)
```
oarap: you can loop until ulLength-1

and after the loop you append it without comma. this would help you avoid if, else

hsyuan: make sense

libnaucrates/src/CDXLUtils.cpp
hsyuan: you are looking at the old commit

libnaucrates/src/CDXLUtils.cpp
hsyuan: well, if changing it this way, the code going to be:
```cpp
	if (0 == ulLength)
	{
		return pstr;
	}
	ULONG ul = 0;
	for (; ul < ulLength-1; ul++)
	{
		CHAR tValue = *((*pdrgsz)[ul]);
		pstr->AppendFormat(GPOS_WSZ_LIT("%c%ls"), tValue, CDXLTokens::PstrToken(EdxltokenComma)->Wsz());
	}
	CHAR tValue = *((*pdrgsz)[ul]);
	pstr->AppendFormat(GPOS_WSZ_LIT("%c"), tValue);
```

It is as succinct as the original one. I will keep it as it is.

libgpopt/src/translate/CTranslatorExprToDXL.cpp
```diff
@@ -4702,17 +4710,42 @@ CTranslatorExprToDXL::ConstructLevelFiltersPartitionSelectorRange
...
-							m_pmp,
-							m_pmda,
-							pdxlnEq,
-							pmdidTypePartKey,
-							pmdidTypeOther,
-							NULL /*pmdidTypeCastExpr*/,
-							NULL /*pmdidCastFunc*/,
-							ulLevel
-							);
+			if (fRangePart)
```
bhuvnesh2703: The name of the function is `ConstructLevelFiltersPartitionSelectorRange` which does not correlated with the new changes..

bhuvnesh2703:
LGTM (Bhuvnesh & Ekta)

libgpopt/src/translate/CTranslatorExprToDXL.cpp
hsyuan: hmm, how about `ConstructLevelFilters4PartitionSelector`?

hsyuan:
@d @bhuvnesh2703 @oarap @dhanashreek89, I think I addressed all the PR comments. Please let me know if you have any other comments / concerns. It passed all the test and hence I am thinking of pushing this tomorrow Morning. Please let me know if you want more time to review and I am more than happy to wait for it.

hsyuan:
Pushed to master, thanks all for reviewing!

PR #151, opened 2017-03-16
Fix bug that can't generate equality partition filters for multilevel partition table

hsyuan:
The equality parition filter of partition selector works very well for single
level partition table, but if the table has multilevel partitions, e.g. 2
levels with pk1, pk2 as the partition key for level 1 and 2, and there is a
equality predicate in the query, say pk1 = 6, then level 2 equality filter is null,
the function `FEqPartFiltersAllLevels` will return false, causing the equality
predicate put into PartFilters instead of PartEqFilters. This bug has been
fixed in the patch.

[#141826453]


libgpopt/src/translate/CTranslatorExprToDXL.cpp
```diff
@@ -4592,10 +4592,16 @@ CTranslatorExprToDXL::FEqPartFiltersAllLevels
 		CExpression *pexprEqFilter = popSelector->PexprEqFilter(ul);
 		CExpression *pexprFilter = popSelector->PexprFilter(ul);
 
-		if (NULL == pexprEqFilter &&
-			(!fCheckGeneralFilters || NULL == pexprFilter || !CPredicateUtils::FConjunctionOfEqComparisons(m_pmp, pexprFilter)))
+		if (NULL == pexprEqFilter && NULL != pexprFilter)
```
karthijrk: Is this change all about returning true even when `pexprFilter == NULL`?

libgpopt/src/translate/CTranslatorExprToDXL.cpp
```diff
@@ -4592,10 +4592,16 @@ CTranslatorExprToDXL::FEqPartFiltersAllLevels
...
 		{
-			return false;
+			if (!fCheckGeneralFilters)
+			{
+				return false;
+			}
+			else if (!CPredicateUtils::FConjunctionOfEqComparisons(m_pmp, pexprFilter))
```
dhanashreek89: why not just OR the conditions instead of else if?

libgpopt/src/translate/CTranslatorExprToDXL.cpp
hsyuan: when pexprEqFilter == NULL, if the fCheckGeneralFilters is false, it always return false.

libgpopt/src/translate/CTranslatorExprToDXL.cpp
hsyuan: yes, good catch.

dhanashreek89:
If `pexprEqFilter` and `pexprFilter` are both NULL are we supposed to return `true`? A bit confusing? If yes, can you add a  comment to explain it better?

hsyuan:
I think it should return true. why do you think it is confusing?

dhanashreek89:
The comment of the function says: `Check whether the given partition selector has equality filters on all partitioning levels. If fCheckGeneralFilters is true then the function checks the general filters as well. Otherwise, only the EqFilters are considered`

So, if there is no equality filter at all, it will still return `true`. This contradicts with the comment of the function because it says it is checking if it has equality filters on levels.

dhanashreek89:
According to the function comment, we feel like the code should be like below:

```
		if (NULL == pexprEqFilter)
		{
			return false;
		}
		else
		{
			if(fCheckGeneralFilters)
			{
				if(pexprFilter == NULL || !CPredicateUtils::FConjunctionOfEqComparisons(m_pmp, pexprFilter)
				{
					return false;
				}
			}
		}

```

hsyuan:
I see what you mean. The comments needs to be addressed. 
But in the code you posted, if any one of the partition level does not have eq filter, 
e.g.
select * from pt where pk2 = 3; 
but no filter on 1st level partition, it is going to return false without checking the 2nd level. The returning result stored in var fPassThrough is false, then the equality filter will be put into general filter in the DXL, which is not what we want. And the bug is still there.

hsyuan:
The comment should be: `Check whether the given partition selector has equality filters or no filters on all partitioning levels.`

libgpopt/src/translate/CTranslatorExprToDXL.cpp
karthijrk: Ok, looks like from other comments, we want to return `true` when `pexprEqFilter == NULL` and `pexprFilter == NULL`.

hsyuan: not necessarily return true in this case, but leave it to the next partition level to determine.

hsyuan:
Updated comments.

dhanashreek89:
We feel like `pexprFilter ` should be only checked if`fCheckGeneralFilters` is `true`. So, the current code makes it really complicated to understand the intent of this function. We would request to restructure to logic and comment on every condition. 

Dhanashree && ( 1 || 0 ) &&  Omer

hsyuan:
You can use the example query in the story to verify what you feels like is wrong or not. I think `fCheckGeneralFilters` is used to indicate whether check the content of general filter is conjunction of equality filter or not. If it is false, we always view the general filter as non-equality filter if the `pexprFilter` is not null.

libgpopt/src/translate/CTranslatorExprToDXL.cpp
```diff
@@ -4570,8 +4570,9 @@ CTranslatorExprToDXL::PdxlnPartitionSelectorFilter
 //		CTranslatorExprToDXL::FEqPartFiltersAllLevels
 //
 //	@doc:
-//		Check whether the given partition selector has equality filters on
-//		all partitioning levels. If fCheckGeneralFilters is true then the function
+//		Check whether the given partition selector only has equality filters
```
foyzur: Why do we need both a doc and a header comment?

hsyuan: I don't know who made this decision. It's a legacy issue.

dhanashreek89:
It makes sense now but this should be documented that way. Because `fCheckGeneralFilters` totally implies that it is a precondition as it is passed as a function parameter.

hsyuan:
In production, the fCheckGeneralFilters is always false, it is true only in
debug mode. I didn't see any point in the assert part in debug mode. And I
can't figure out an example to make general filters have conjunction of
equality filters. All make no sense to me and make the code hard to to
understand, so remove it.

Please review the changes again.

PR #152, opened 2017-03-20
Stop timing DXL to Scalar Expr translation time [#141739201]

khannaekta:
Removing timing Scalar DXL To Scalar Expr Translation Time
as for certain queries it produces many lines of noise.


d:
Ship it!

PR #156, opened 2017-03-31
Improve Match Part Constraints logic

oarap:
We do not need to call FSubsetPartCnstr() twice to check for equality.
Two part constraints can are equal if their lengths match and
one is subset of the other.

Signed-off-by: Omer Arap <oarap@pivotal.io>


oarap:
I think this is already in, so closing

PR #157, opened 2017-03-31
Remove dead code about SharedScan

hsyuan:
When I was trying to understand how does Orca generate plan for CTE using
shared input scan, I found that the share input scan is generated during CTE
producer & consumer DXL node to PlannedStmt translation stage, instead of Expr
to DXL  stage inside Orca. It turns out CDXLPhysicalSharedScan is not used
anywhere, so remove all the related dead code.


vraghavan78:
Can do the same to GPDB4

hsyuan:
I think so. But will build and run on GPDB4.

PR #162, opened 2017-04-10
Remove relation mdid from indices

d:
While working on https://github.com/greenplum-db/gporca/pull/160 we found that the abstraction in ORCA contains "the relation of an index". We don't use this information anywhere, instead, we get indices from "table descriptors", not the other way around. And retaining this dead abstraction actually introduces undue complexity for https://github.com/greenplum-db/gporca/pull/160 and https://github.com/greenplum-db/gpdb/pull/2163 . We are removing all mentions of "relation of an index".


d:
@vraghavan78 said LGTM

PR #164, opened 2017-04-10
Remove CDXLScalarInitPlan related dead code [#143323299]

vraghavan78:
CDXLScalarInitPlan was introduced in 2010, because at that time we only had DXL<->PlStmt translation. Orca does not generate InitPlans, so this is dead code and can be removed.


hsyuan:
enum `EdxlopScalarInitPlan` and `EdxltokenScalarInitPlan` should also be global searched and removed.

vraghavan78:
@hsyuan I fixed the remainder.

PR #165, opened 2017-04-11
Include missing transform for partial index scans

d:
For some reason we didn't disable
InnerJoinWithInnerSelect2PartialDynamicIndexGetApply but we should.


dhanashreek89:
LGTM

Jemish and Dhanashree

PR #167, opened 2017-04-20
    Handle NULL while deriving propagation expression [#143443847]

bhuvnesh2703:
While constructing the propagation expression in case of queries involving default list partitions which have indexes, we ended up with NULLs for Start and End Range End Points. Default Partition are expected not to have a Start / End Range Point.
This caused a crash. Currently, we fallback to planner. But we need to handle it in a better way in future.


bhuvnesh2703:
@hsyuan @vraghavan78 @oarap @d @jpatel-pivotal @khannaekta pls review.

PR #168, opened 2017-04-21
Do not record time needed to access metadata when profiling is disabled

vraghavan78:
We were unnecessarily recording md access time when it is not required. This issue is now resolved.


vraghavan78:
@ashwinstar @oarap @d @hsyuan

vraghavan78:
Merged.

PR #170, opened 2017-04-22
Simplify subquery with outer ref as output and its child as limit over constant table get

vraghavan78:
Consider the SQL example:

SELECT * FROM foo, bar where foo.a = (select bar.b limit 1);

Orca generates a logical plan with scalar subquery that has limit over a CTG. By the addition to the preprocessing step, we will convert

foo.a = (select bar.b limit 1);

TO

foo.a = bar.b

### DIFF
--- a/libgpopt/src/operators/CExpressionPreprocessor.cpp
+++ b/libgpopt/src/operators/CExpressionPreprocessor.cpp
@@ -33,6 +33,9 @@
 #include "naucrates/md/IMDScalarOp.h"
 #include "naucrates/md/IMDType.h"
 #include "naucrates/statistics/CStatistics.h"
+#include "naucrates/base/IDatumInt2.h"
+#include "naucrates/base/IDatumInt4.h"
+#include "naucrates/base/IDatumInt8.h"
 
 #include "naucrates/traceflags/traceflags.h"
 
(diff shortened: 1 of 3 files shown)


vraghavan78:
@hsyuan @d @oarap @bhuvnesh2703 @khannaekta @dhanashreek89 this does not handle

```
explain SELECT *  FROM foo, bar  WHERE foo.a = (select bar.b from x limit 1) ;
explain SELECT *  FROM foo, bar  WHERE foo.a = (select bar.b from x limit 10 offset 1) ;
```
I created an issue https://github.com/greenplum-db/gporca/issues/171. For these case we need a different solution (may be a clean of the subquery handle).

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -414,17 +417,63 @@ CExpressionPreprocessor::PexprRemoveSuperfluousLimit
...
+
+			if (COperator::EopScalarConst == popCount->Eopid())
+			{
+				IDatum *pdatumCount = CScalarConst::PopConvert(popCount)->Pdatum();
+				if (!pdatumCount->FNull())
+				{
+					CMDAccessor *pmda = COptCtxt::PoctxtFromTLS()->Pmda();
+					IMDType::ETypeInfo eti = pmda->Pmdtype(pdatumCount->Pmdid())->Eti();
+
+					if ((IMDType::EtiInt2 == eti) || (IMDType::EtiInt4 == eti) || (IMDType::EtiInt8 == eti))
```
hsyuan: as of Postgres 8.2, the offset and count expressions are expected to yield int8, rather than int4 as before. Do we need this?

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -414,17 +417,63 @@ CExpressionPreprocessor::PexprRemoveSuperfluousLimit
...
+							lValue = dynamic_cast<IDatumInt4 *>(pdatumCount)->IValue();
+						}
+						else
+						{
+							lValue = dynamic_cast<IDatumInt8 *>(pdatumCount)->LValue();
+						}
+
+						CLogicalConstTableGet *popCTG = CLogicalConstTableGet::PopConvert(pexprCTG->Pop());
+
+						if (lValue >= popCTG->Pdrgpdrgpdatum()->UlLength())
```
hsyuan: I don't understand what does the right hand side do.

oarap:
I would be more confident if we can provide a generic and rebust solution for all the cases. They seem to be very close cases and follow the same code path. The level of `if`s also bothers me. 

-1

vraghavan78:
This may need to rethink. :(

PR #175, opened 2017-05-05
Retain additional conditions while translating Correlated Nested Loop Join

khannaekta:
For a query using a correlated subselect as predicate, such as:

```
CREATE TABLE partitioned_table (a int, pk int) DISTRIBUTED BY (a)
	PARTITION BY range(pk) (end(5), end(10));
CREATE TABLE other_table (c int, d int) DISTRIBUTED BY (c);
INSERT INTO partitioned_table VALUES (1, 1), (2, 4), (3, 9);
ANALYZE ROOTPARTITION partitioned_table;

EXPLAIN SELECT pk from partitioned_table WHERE a > (SELECT d FROM other_table WHERE c = a) AND pk < 12;
```

ORCA will generate a Correlated Nested Loop Left Outer Join, which
should translate to a DXL Scan under a DXL SubPlan filter. However, the
translation happened in the following order:

0. Translate outer child (which has a filter) of correlated NLJ
0. Build a DXL SubPlan using the inner child, which is intended to serve as an
   "additional filter condition" on top of the outer child
0. Now based on the DXL plan for the outer child, we decide whether or not to
   use the additional condition (generated in #2 or #3) as a filter in the
   final result.
0. If the outer child is a Physical Sequence, we discarded the condition
   assuming that filter condition is already present in the partition
   selector.
0. This code to discard the subplan was added in
   e99325c because, previously, we were always
   inserting additional filter as "the second child" based on the assumption
   that every DXL node has a filter child in the 2nd place. As it turned out,
   the DXL Sequence node is one counterexample: it has no filter, and it's
   second child is expected to be a partition selector.
0. We didn't catch this error in e99325c because the test cases had a trivial
   additional filter condition of a constant true, so dropping it didn't really
   raise any eyebrows. However, the generally correct approach should be to
   retain this additional condition, either as an additional DXL Result node on
   top of the DXL for the outer child, or for appropriate types of nodes,
   inline the condition into existing filters.

This patch set fixes that.


libgpopt/src/translate/CTranslatorExprToDXL.cpp
```diff
@@ -1273,38 +1273,69 @@ CTranslatorExprToDXL::PdxlnResult
...
 {
 	GPOS_ASSERT(NULL != pexprRelational);
-	GPOS_ASSERT(NULL != pdxlprop);
 
 	// translate relational child expression
 	CDXLNode *pdxlnChild = Pdxln(pexprRelational, pdrgpcr, pdrgpdsBaseTables, pulNonGatherMotions, pfDML, false /*fRemap*/, false /*fRoot */);
 
-	// wrap condition in a DXL filter node
-	CDXLNode *pdxlnFilter = PdxlnFilter(pdxlnScalar);
+	return PdxlnResultFromScalarConditionAndDxlChild
```
oarap: Either put everything in one line or indent the parameters more so that it alings with the funtion. This kinda looks as if they are `return`'s parameters.

libgpopt/src/translate/CTranslatorExprToDXL.cpp
```diff
@@ -1273,38 +1273,69 @@ CTranslatorExprToDXL::PdxlnResult
...
-										pdxlprop,
-										pdxlnPrL,
-										pdxlnFilter,
-										pdxlnOneTimeFilter,
-										pdxlnChild
-										);
+		// create an empty one-time filter
+		CDXLNode *pdxlnOneTimeFilter = GPOS_NEW(m_pmp) CDXLNode(m_pmp, GPOS_NEW(m_pmp) CDXLScalarOneTimeFilter(m_pmp));
+
+		return CTranslatorExprToDXLUtils::PdxlnResult
```
oarap: same as above

libgpopt/src/translate/CTranslatorExprToDXL.cpp
```diff
@@ -3489,16 +3503,38 @@ CTranslatorExprToDXL::PdxlnResultFromNLJoinOuter
...
 			// empty filter
-			CDXLNode *pdxlnFilter = PdxlnFilter(pdxlnCond);
+			CDXLNode *pdxlnFilter = PdxlnFilter(pdxlnBoolExpr);
 			pdxlnResult->ReplaceChild(ulIndexFilter /*ulPos*/, pdxlnFilter);
 		}
 			break;
 		case EdxlopPhysicalSequence:
-			pdxlnCond->Release();
+			pdxlprop->AddRef();
+			pdxlnResult = PdxlnResultFromScalarConditionAndDxlChild
```
oarap: same as above for indentation

libgpopt/src/translate/CTranslatorExprToDXL.cpp
```diff
@@ -3489,16 +3503,38 @@ CTranslatorExprToDXL::PdxlnResultFromNLJoinOuter
...
+			// get the original condition from the filter node
+			// create a new AND expression
 			ULONG ulIndexFilter = UlIndexFilter(edxlopid);
 			GPOS_ASSERT(ulIndexFilter != ULONG_MAX);
+			CDXLNode *pdxlnOrigFilter = (*pdxlnResult)[ulIndexFilter];
+			GPOS_ASSERT(EdxlopScalarFilter == pdxlnOrigFilter->Pdxlop()->Edxlop());
+			CDXLNode *pdxlnOrigCond = (*pdxlnOrigFilter)[0];
+			pdxlnOrigCond->AddRef();
+
+			CDXLNode *pdxlnBoolExpr = PdxlnScBoolExpr
```
oarap: same thing

libgpopt/src/translate/CTranslatorExprToDXL.cpp
```diff
@@ -1273,38 +1273,69 @@ CTranslatorExprToDXL::PdxlnResult
...
 	BOOL *pfDML,
 	CDXLNode *pdxlnScalar,
 	CDXLPhysicalProperties *pdxlprop
 	)
 {
 	GPOS_ASSERT(NULL != pexprRelational);
-	GPOS_ASSERT(NULL != pdxlprop);
 
 	// translate relational child expression
 	CDXLNode *pdxlnChild = Pdxln(pexprRelational, pdrgpcr, pdrgpdsBaseTables, pulNonGatherMotions, pfDML, false /*fRemap*/, false /*fRoot */);
```
vraghavan78: may be pdxlnRelation

libgpopt/src/translate/CTranslatorExprToDXL.cpp
```diff
@@ -1273,38 +1273,69 @@ CTranslatorExprToDXL::PdxlnResult
...
+	pdxlnScalar,
+	pdxlprop
+	);
+}
 
-	GPOS_ASSERT(NULL != pexprRelational->Prpp());
-	CColRefSet *pcrsOutput = pexprRelational->Prpp()->PcrsRequired();
-	CDXLNode *pdxlnPrL = PdxlnProjList(pcrsOutput, pdrgpcr);
+CDXLNode *
+CTranslatorExprToDXL::PdxlnResultFromScalarConditionAndDxlChild
```
vraghavan78: PdxlnAddScalarFilterOnRelationalChild

libgpopt/include/gpopt/translate/CTranslatorExprToDXL.h
```diff
@@ -218,6 +218,17 @@ namespace gpopt
 
 			CDXLNode *PdxlnResult(CExpression *pexprFilter, DrgPcr *pdrgpcr, DrgPds *pdrgpdsBaseTables, ULONG *pulNonGatherMotions, BOOL *pfDML);
 
+			// given a DXL plan tree pdxlnChild which represents the physical plan pexprRelational, construct a DXL
+			// Result node that filters on top of it using the scalar condition pdxlnScalar
+			CDXLNode *PdxlnResultFromScalarConditionAndDxlChild
```
vraghavan78: PdxlnAddScalarFilterOnRelationalChild

libgpopt/include/gpopt/translate/CTranslatorExprToDXL.h
```diff
@@ -218,6 +218,17 @@ namespace gpopt
...
+				(
+				CExpression *pexprRelational,
+				CDXLNode *pdxlnChild,
```
vraghavan78: pdxlnRelationalChild

libgpopt/include/gpopt/translate/CTranslatorExprToDXL.h
```diff
@@ -218,6 +218,17 @@ namespace gpopt
...
+				DrgPcr *pdrgpcr,
+				CDXLNode *pdxlnScalar,
```
vraghavan78: pdxlnScalarChild

libgpopt/include/gpopt/translate/CTranslatorExprToDXL.h
```diff
@@ -218,6 +218,17 @@ namespace gpopt
 
 			CDXLNode *PdxlnResult(CExpression *pexprFilter, DrgPcr *pdrgpcr, DrgPds *pdrgpdsBaseTables, ULONG *pulNonGatherMotions, BOOL *pfDML);
 
+			// given a DXL plan tree pdxlnChild which represents the physical plan pexprRelational, construct a DXL
+			// Result node that filters on top of it using the scalar condition pdxlnScalar
+			CDXLNode *PdxlnResultFromScalarConditionAndDxlChild
+				(
+				CExpression *pexprRelational,
```
vraghavan78: can you tell why you need pexprRelational now if you already have the DXL node representing it.

libgpopt/src/translate/CTranslatorExprToDXL.cpp
```diff
@@ -1273,38 +1273,69 @@ CTranslatorExprToDXL::PdxlnResult
...
+	(
+	CExpression *pexprRelational,
```
vraghavan78: If all you need is the output columns, then pass that.

libgpopt/src/translate/CTranslatorExprToDXL.cpp
```diff
@@ -1273,38 +1273,69 @@ CTranslatorExprToDXL::PdxlnResult
...
+	CDXLNode *pdxlnChild,
+	DrgPcr *pdrgpcr,
```
vraghavan78: can we rename this array?

dhanashreek89:
Addressed the comments. @vraghavan78 @oarap Please review

libgpopt/include/gpopt/translate/CTranslatorExprToDXL.h
```diff
@@ -218,6 +218,17 @@ namespace gpopt
...
 
+			// given a DXL plan tree pdxlnChild which represents the physical plan pexprRelational, construct a DXL
+			// Result node that filters on top of it using the scalar condition pdxlnScalar
+			CDXLNode *PdxlnAddScalarFilterOnRelationalChild
+				(
+				CColRefSet *pcrsOutput,
+				CDXLNode *pdxlnRelationalChild,
+				DrgPcr *pdrgpcrOrder,
+				CDXLNode *pdxlnScalarChild,
+				CDXLPhysicalProperties *pdxlprop
```
vraghavan78: can you re-order the parameters:

pdxlnRelationalChild
pdxlnScalarChild
pdxlprop
pcrsOutput
pdrgpcrOrder

PR #178, opened 2017-05-30
Remove the intentional fallback to planner

bhuvnesh2703:
We must not come to the path to create a test propagation expression
for queries over partitioned table with default partitions and indexes
all over the parts. Thus, removing the fallback to planner as we fixed
the issue in commit `19cfbcf6e2885c1901f59f863563842ed3912430` in gpdb4
repository


libgpopt/src/translate/CTranslatorExprToDXLUtils.cpp
```diff
@@ -589,17 +589,8 @@ CTranslatorExprToDXLUtils::PdxlnPartialScanTestRange
...
-			// Currently we fallback and should handle it better in future.
-			GPOS_RAISE
-				(
-				gpdxl::ExmaDXL,
-				gpdxl::ExmiExpr2DXLUnsupportedFeature,
-				GPOS_WSZ_LIT("Queries over default list partition that have indexes")
-				);
-		}
+		// Datum should not be NULL.
+		GPOS_ASSERT(pdatum != NULL);
```
hsyuan: No, don't remove it. I don't think this change is needed. How are you sure we will not have default partition coming here for other queries?

PR #181, opened 2017-06-02
Only use locks when GPORCA is in parallel mode

vraghavan78:


libnaucrates/src/statistics/CStatistics.cpp
```diff
@@ -2882,7 +2882,11 @@ CStatistics::UlIndexUpperBoundNDVs
 {
 	GPOS_ASSERT(NULL != pcr);
  	CAutoMutex am(m_mutexCardUpperBoundAccess);
```
hsyuan: this should go to if block

vraghavan78: i do not think so.

hsyuan: I take it back. I was wrong. If we put it inside if block, CAutoMutex will be destroyed after we go outside of if block, then the lock will not take effective.

hsyuan:
Oh, just FYI: If you want to bump the version, bump the patch version, don't bump the minor version. Because there is no API changes.

oarap:
-1 
I do not like this approach. We need to decide how we do this first. This is so `if`fy.

hsyuan:
How about this:

```c++
#define GPOS_LOCK(am)  \
	do { \
		if (GPOS_FTRACE(EopttraceParallel)) \
			am.Lock(); \
	} while(0)
```

So we can just replace `am.Lock();` with `GPOS_LOCK(am)`.
lol

PR #183, opened 2017-06-06
Remove inl files

oarap:
The template classes in orca mostly do the implementation in `.inl` files while some implementation also exists in `.h` files. Personally I find it hard to traverse the code in .inl files since some IDEs do not recognize formatting. Therefore I moved the implementation to the `.h` files wherever is applicable. 

This pull request does not port implementation fro `.inl` where there exists `.cpp` implementation file, `.h` only has function definitions such as `CUtils.h`.

I will squash the commits into one, whenever it is approved to merge, for now to ease the review, I left it as separate commits.


hsyuan:
Nice work, just some indent is weird.

libgpopt/include/gpopt/base/CStateMachine.inl
```diff
@@ -1,595 +0,0 @@
...
-		m_tenumstate = TesInitial();
-		m_fInit = true;
-	}
-
-
-	//---------------------------------------------------------------------------
-	//	@function:
-	//		CStateMachine::FTransition
-	//
-	//	@doc:
```
khannaekta: It will be good to have these descriptive comments in the .h file too.

oarap: We actually do not like those comments. It is also in our agenda to remove all of these useless comments from the codebase.

PR #185, opened 2017-06-09
Added new operator version for Constant Table

bhuvnesh2703:
In case of Full Outer Join, the child expression are implemented as
CTE Expression and if we have a constant table under CTE consumer in the
outer child of Full Outer Join, we produced a plan where the consumer
was executed on the master but the producer was not available on the master,
due to which the query execution will hang. Constant Table operator has
UNIVERSAL distribution spec which means that it can be evaluation at the
segments or the master due to which full outer join pulls the inner child to the
to the master.
This patch fixes the issue by introducing a new Constant Table operator with
Replicated Distribution Spec which replaces the Constant Table Operator
with singleton spec during expression preprocessing only if its under a
FULL OUTER JOIN.


vraghavan78:
-1. Can we first have a means of detecting a back plan and fall back.

vraghavan78:
Closing the PR since we decided to move in another direction

PR #186, opened 2017-06-21
Remove part oid

bhuvnesh2703:
In GPDB, PartOidExpr is not used however ORCA still generates it.
But HAWQ uses PartOid for sorting while inserting into Append Only
Row / Parquet Partitioned tables.

This patch uses Parquet Storage and Number of Partitions in a Append
Only row partitioned table to decide if PartOid should be generated.
In case of GPDB, Parquet storage is not supported and the GUC to control
the number of partitions above which sort should be used is set to int
max which is practically not feasible, so in case of GPDB PartOid expr
will never be generated, however HAWQ can control the generation of
PartOid based on the value of already existing GUCs in HAWQ.


hsyuan:
I suspect HAWQ doesn't use PartOid for sorting while inserting at all. We should remove PartOid completely, no matter what.

vraghavan78:
@hsyuan HAWQ will eventually update Orca.

libgpopt/src/operators/CLogicalDML.cpp
```diff
@@ -93,13 +93,17 @@ CLogicalDML::CLogicalDML
 	GPOS_ASSERT(NULL != pdrgpcrSource);
 	GPOS_ASSERT(NULL != pbsModified);
 	GPOS_ASSERT(NULL != pcrAction);
-	GPOS_ASSERT(NULL != pcrTableOid);
+//	GPOS_ASSERT(NULL != pcrTableOid);
 	GPOS_ASSERT_IMP(EdmlDelete == edmlop || EdmlUpdate == edmlop,
 					NULL != pcrCtid && NULL != pcrSegmentId);
```
vraghavan78: can we have assert that checks pcrTableOid is not NULL for HAWQ?

libgpopt/src/operators/CLogicalDML.cpp
```diff
@@ -400,9 +404,14 @@ CLogicalDML::OsPrint
 	CUtils::OsPrintDrgPcr(os, m_pdrgpcrSource);
 	os	<< "], Action: (";
 	m_pcrAction->OsPrint(os);
-	os << "), Oid: (";
-	m_pcrTableOid->OsPrint(os);
 	os << ")";
```
vraghavan78: move 407 below 414

libgpopt/src/operators/CPhysicalDML.cpp
```diff
@@ -66,7 +66,7 @@ CPhysicalDML::CPhysicalDML
 	GPOS_ASSERT(NULL != pdrgpcrSource);
 	GPOS_ASSERT(NULL != pbsModified);
 	GPOS_ASSERT(NULL != pcrAction);
-	GPOS_ASSERT(NULL != pcrTableOid);
+//	GPOS_ASSERT(NULL != pcrTableOid);
```
vraghavan78: same as above, have a GPOS_ASSERT_IMP

libgpopt/src/operators/CPhysicalDML.cpp
```diff
@@ -661,9 +665,14 @@ CPhysicalDML::OsPrint
 	CUtils::OsPrintDrgPcr(os, m_pdrgpcrSource);
 	os	<< "], Action: (";
 	m_pcrAction->OsPrint(os);
-	os << "), Oid: (";
-	m_pcrTableOid->OsPrint(os);
 	os << ")";
```
vraghavan78: move 668 below 676

libgpopt/src/operators/CLogicalDML.cpp
```diff
@@ -93,13 +93,17 @@ CLogicalDML::CLogicalDML
 	GPOS_ASSERT(NULL != pdrgpcrSource);
 	GPOS_ASSERT(NULL != pbsModified);
 	GPOS_ASSERT(NULL != pcrAction);
-	GPOS_ASSERT(NULL != pcrTableOid);
+//	GPOS_ASSERT(NULL != pcrTableOid);
```
vraghavan78: no dead code please

libgpopt/src/xforms/CXformUtils.cpp
```diff
@@ -1296,36 +1296,50 @@ CXformUtils::PexprLogicalDMLOverProject
 	CColRef *pcrAction = NULL;
 	CColRef *pcrOid = NULL;
 
+	BOOL fInsertSortOnParquet = (!GPOS_FTRACE(EopttraceDisableSortForDMLOnParquet) && ptabdesc->Erelstorage() == IMDRelation::ErelstorageAppendOnlyParquet);
+
+	COptimizerConfig *poconf = COptCtxt::PoctxtFromTLS()->Poconf();
+	BOOL fInsertSortOnRows = (ptabdesc->Erelstorage() == IMDRelation::ErelstorageAppendOnlyRows) && (poconf->Phint()->UlMinNumOfPartsToRequireSortOnInsert() <= ptabdesc->UlPartitions());
+
+	BOOL fGeneratePartOid = fInsertSortOnParquet || fInsertSortOnRows;
+
```
hsyuan: Can this part be a separate function?

libgpopt/src/operators/CLogicalDML.cpp
```diff
@@ -400,8 +403,13 @@ CLogicalDML::OsPrint
...
 	m_pcrAction->OsPrint(os);
-	os << "), Oid: (";
-	m_pcrTableOid->OsPrint(os);
+	
+	if (m_pcrTableOid != NULL)
+	{
+		os << ", Oid: (";
+		m_pcrTableOid->OsPrint(os);
+		os << ")";
+	}
```
hsyuan: I think this is wrong. What you want to do is just wrapping the removed 2 lines in a `if` block, right?

libgpopt/src/operators/CPhysicalDML.cpp
```diff
@@ -661,8 +664,13 @@ CPhysicalDML::OsPrint
...
 	m_pcrAction->OsPrint(os);
-	os << "), Oid: (";
-	m_pcrTableOid->OsPrint(os);
+	
+	if (m_pcrTableOid != NULL)
+	{
+		os << ", Oid: (";
+		m_pcrTableOid->OsPrint(os);
+		os << ")";
+	}
```
hsyuan: same here

PR #189, opened 2017-06-22
Fix cardinality cast func

bhuvnesh2703:
Fix Cardinality Estimation for predicate with cast

    If the predicate is of the form where col::cast == literal,
    we currently do not recurse into the expression to identify the
    ident and derive stats on it due to which we consider the
    predicate as unsupported and put default stats which causes
    cardinality mis-estimation.
    This patch recurses into the cast expression to extract the ident col
    and calculate the stats on it.

For the query in this issue, the predicate is of the form
```
+--CScalarCmp (=) 
     |--CScalarCast
     |   +--CScalarFunc (varchar)
     |        |--CScalarCast
     |        |  +--CScalarIdent "b" (1)
     |        |--CScalarConst (14)
     |        +--CScalarConst (1)
     +--CScalarConst

Filter: "varchar"(b::character varying, 14, true)::text = 'efg'::text
```


vraghavan78:
Closing this PR.

PR #172, opened 2017-04-26
Enable caching for Motion Operator [#144378477]

jpatel-pivotal:
This PR introduces a request caching mechanism in `CPhysicalMotion` using a Hashmap similar
to what we did with NL join for handling duplicate partition propogation spec requests.

PR #154, opened 2017-03-29
142624555 caching nl join ppe

oarap:
This PR introduces caching for NL join as it was utilized for hash join for handling partition propagation spec. 

We lifted up the implementation from hash join to `CPhysicalJoin` so that NL join can also benefit from the shared logic.

PR #192, opened 2017-06-28
Move Caching for Join and Motion Operators into CPhysical

jpatel-pivotal:
Similar to https://github.com/greenplum-db/gporca/pull/154 and https://github.com/greenplum-db/gporca/pull/172 we want to move the implementation of caching from `CPhysicalJoin` into  `CPhysical` so that different Operators can benefit from caching partition propagation requests (key) and their corresponding partition propagation spec (value). If the operator sees the same request in the cache, the operator should re-use the partition propagation spec from the cache instead of issuing a new partition propagation request.

This PR moves the caching implementation into `CPhysical` and enables `CPhysicalJoin` and `CPhysicalMotion` to use it. 

We are **not looking to merge this PR** at this point specifically because we measured timings pre and post changes and did not see the positive results that we saw when we implemented caching individually in `CPhysicalJoin` and `CPhysicalMotion`. Below is the relevant timing information I captured:

ICW tests before change: `21m19.199s`
ICW tests  after change: `22m29.977s`

Orca Unit tests before change: `44.05 sec`
Orca Unit tests after change: `46.40 sec`

Long running query before change: `201724.838 ms` OR `3.362080633333 mins`
Long running query after change: `218304.640 ms` OR `3.63841066667 mins`

TPCDS  1TB explain plan times before:` 2m52.453s`
TPCDS  1TB explain plan times after: `3m7.468s` 

**Avg. Change:** `-6.94%` **slower** 

Would like some help to figure out:

1. Did we miss something or should we do something differently?
2. Any other experimentation that we can do?
3. Any other suggestions?


hsyuan:
I didn't see the necessity to move partition propagation from physicaljoin to physical base operator. 
But putting this aside, I think you should 
1. override the function `PpprCreate` in CPhysicalJoin and CphysicalMotion. `PpprCreate` in CPhysical should just return NULL.
2. The hashmap insert and lookup performance may be a problem too. It uses '%' to calculate bucket index instead of `&`. But I don't know how much it is going to affect.

hsyuan:
every time the operator is destructed, the hashmap is also gone. It actually doesn't cache anything, right?

vraghavan78:
Closing this since it has not gotten traction and not doing what is needed

PR #193, opened 2017-06-29
Enable GPORCA to generate better plans for non-correlated exists subquery in the WHERE clause

vraghavan78:
Consider the following exists subquery, `(select 1 from bar)`. GPORCA generates an elaborate count based implementation of this subquery. If bar is a fact table, the count is going to be expensive.

```
vraghavan=# explain select count(*) from foo where foo.a = foo.b and exists (select 1 from bar);
                                                 QUERY PLAN
 Aggregate  (cost=0.00..1331807.20 rows=1 width=8)
   ->  Result  (cost=0.00..1331807.20 rows=1 width=1)
         ->  Nested Loop  (cost=0.00..1331807.20 rows=1 width=1)
               Join Filter: true
               ->  Result  (cost=0.00..438.59 rows=1 width=1)
                     Filter: (count((count()))) > 0::bigint
                     ->  Aggregate  (cost=0.00..438.59 rows=1 width=8)
                           ->  Gather Motion 3:1  (slice2; segments: 3)  (cost=0.00..438.59 rows=1 width=8)
                                 ->  Aggregate  (cost=0.00..438.59 rows=1 width=8)
                                       ->  Table Scan on bar  (cost=0.00..437.97 rows=333603 width=1)
                           ->  Table Scan on foo  (cost=0.00..431.00 rows=1 width=1)
                                 Filter: a = b
 Optimizer status: PQO version 2.34.0
(15 rows)
```
Planner on the other hand uses LIMIT as shown in the INIT plan.

```
vraghavan=# explain select count(*) from foo where foo.a = foo.b and exists (select 1 from bar);
                                        QUERY PLAN
 Aggregate  (cost=1.11..1.12 rows=1 width=8)
   InitPlan  (slice3)
           ->  Gather Motion 3:1  (slice2; segments: 3)  (cost=0.00..0.03 rows=1 width=0)
                       ->  Seq Scan on bar  (cost=0.00..11109.09 rows=333603 width=0)
   ->  Gather Motion 3:1  (slice1; segments: 3)  (cost=1.02..1.07 rows=1 width=8)
         ->  Aggregate  (cost=1.02..1.03 rows=1 width=8)
               ->  Result  (cost=0.00..1.01 rows=1 width=8)
                     One-Time Filter: $0
                     ->  Seq Scan on foo  (cost=0.00..1.01 rows=1 width=8)
                           Filter: a = b
 Settings:  optimizer=off
(14 rows)
```

While GPORCA doesnot support init-plan, we can nevertheless generate a better plan by using LIMIT instead of count. After this PR, GPORCA will generate the following plan with LIMIT clause.

```
vraghavan=#  explain select count(*) from foo where foo.a = foo.b and exists (select 1 from bar);
                                                    QUERY PLAN
 Aggregate  (cost=0.00..1357263.96 rows=1 width=8)
   ->  Gather Motion 3:1  (slice3; segments: 3)  (cost=0.00..1357263.96 rows=1 width=8)
         ->  Aggregate  (cost=0.00..1357263.96 rows=1 width=8)
               ->  Nested Loop  (cost=0.00..1357263.96 rows=133442 width=1)
                     Join Filter: true
                     ->  Broadcast Motion 1:3  (slice2)  (cost=0.00..431.76 rows=3 width=1)
                           ->  Limit  (cost=0.00..431.76 rows=1 width=1)
                                 ->  Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..431.76 rows=1 width=1)
                                       ->  Limit  (cost=0.00..431.76 rows=1 width=1)
                                             ->  Table Scan on bar  (cost=0.00..431.69 rows=33190 width=1)
                     ->  Table Scan on foo  (cost=0.00..461.91 rows=133442 width=1)
                           Filter: foo.a = foo.b
 Optimizer status: PQO version 2.34.0
(13 rows)
```

### DIFF
--- a/libgpopt/src/operators/CExpressionPreprocessor.cpp
+++ b/libgpopt/src/operators/CExpressionPreprocessor.cpp
@@ -41,14 +41,7 @@ using namespace gpopt;
 // maximum number of equality predicates to be derived from existing equalities
 #define GPOPT_MAX_DERIVED_PREDS 50
 
-//---------------------------------------------------------------------------
-//	@function:
-//		CExpressionPreprocessor::PexprEliminateSelfComparison
-//
-//	@doc:
-//		Eliminate self comparisons in the given expression
-//
-//---------------------------------------------------------------------------
+// eliminate self comparisons in the given expression
 CExpression *
 CExpressionPreprocessor::PexprEliminateSelfComparison
 	(
@@ -81,15 +74,7 @@ CExpressionPreprocessor::PexprEliminateSelfComparison
 	return GPOS_NEW(pmp) CExpression(pmp, pop, pdrgpexprChildren);
 }
 
-//---------------------------------------------------------------------------
-//	@function:
-//		CExpressionPreprocessor::PexprPruneSuperfluousEquality
-//
-//	@doc:
-//		Remove superfluous equality operations
-//
-//
-//---------------------------------------------------------------------------
+// remove superfluous equality operations
 CExpression *
 CExpressionPreprocessor::PexprPruneSuperfluousEquality
 	(
@@ -121,20 +106,13 @@ CExpressionPreprocessor::PexprPruneSuperfluousEquality
 	return GPOS_NEW(pmp) CExpression(pmp, pop, pdrgpexprChildren);
 }
 
-//---------------------------------------------------------------------------
-//	@function:
-//		CExpressionPreprocessor::PexprTrimExistentialSubqueries
-//
-//	@doc:
-//		An existential subquery whose inner expression is a GbAgg
-//		with no grouping columns is replaced with a Boolean constant
+// an existential subquery whose inner expression is a GbAgg
+// with no grouping columns is replaced with a Boolean constant
 //
 //		Example:
 //
 //			exists(select sum(i) from X) --> True
 //			not exists(select sum(i) from X) --> False
-//
-//---------------------------------------------------------------------------
 CExpression *
 CExpressionPreprocessor::PexprTrimExistentialSubqueries
 	(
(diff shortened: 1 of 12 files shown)


oarap:
I like the pull request and the idea however I think Preprocessing is not the place to put it in. 

It looks more like a transformation rather than a preprocessing. I would suggest to look at `CXformSimplifySubquery` which seems to be a good candidate for the logic.

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -188,6 +188,133 @@ CExpressionPreprocessor::PexprTrimExistentialSubqueries
...
+// simply existential subquery in the WHERE CLAUSE example;
+// Example 1:
+// select * from foo where exists (select * from bar) ->
+// will be transformed to
+// select * from foo, (select * from bar LIMIT 1) where true
+
+// Example 2:
+// select * from foo where exists (select * from bar) and  foo.a = 10 AND not exists (select * from x);
+// will be transformed to
+// select * from (select * from foo where foo.a = foo.b AND not exists (select * from x)), (select * from bar LIMIT 1), (select * from x) where true
```
vraghavan78: sorry, I was playing with two different input queries. One with predicate foo.a = 10 and one with foo.a = foo.b.

Will fix this,

vraghavan78:
@oarap do u agree that a LIMIT based approach is always cheaper or equal cost to an aggregate based implementation. If so, this should be a pre-processing step rather than an xform

oarap:
@vraghavan78 Limit based approach is better. 

I think you only need to preprocess `exists (select * from bar)` to `exists (select * from bar LIMIT 1)`

oarap:
`exists (select * from bar LIMIT 1)` is transformed into `exists (select * from bar LIMIT 1 LIMIT 1)`

there might be a check for limit to remove extra limit

vraghavan78:
I agree with @oarap that this is doing more that a regular pre-processing. Let me think over this.

PR #194, opened 2017-06-30
Refactor the code in CPredicateUtils

vraghavan78:
Minor clean of duplicate code.


oarap:
I approved this but the function naming is a bit problematic. Cmp is different than IDF and function name is kinda hiding IDF.

oarap:
+1, looks better

vraghavan78:
Merged thanks!

PR #195, opened 2017-06-30
Reorder scalar childs

bhuvnesh2703:
Preprocess query to ensure Scalar Ident is on LHS

    CScalarCmp operator must have the CScalarIdent operator
    on the LHS and CScalarConst operator on the RHS. If it's
    not the case the predicate is assigned a default selectivity
    due to which the Cardinality estimate is impacted.

    This patch fixes the issue by reordering the children of
    CScalarCmp operator if CScalarIdent operator is on the RHS
    on CScalarConst is on LHS. Also the Comparision operator is changed
    due to the reordering of the arguments, if supported. If the
    corresponding comparision operator does not exist or supported, the
    children are not reordered.
    Only cases of the type CONST = VAR are handled by this patch.


libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -2206,6 +2206,59 @@ CExpressionPreprocessor::PexprPruneProjListProjectOrGbAgg
...
+CExpressionPreprocessor::PexprReorderScalarCmpChildren
+	(
+	IMemoryPool *pmp,
+	CExpression *pexpr
+	)
+{
+	GPOS_ASSERT(NULL != pexpr);
+
+	COperator *pop = pexpr->Pop();
+	if (COperator::EopScalarCmp == pop->Eopid())
```
vraghavan78: what about INDF and IDF?

bhuvnesh2703: Thanks, included it now.

vraghavan78:
Hey can you refactor the code because we are using the same code in two places?

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -2206,6 +2206,82 @@ CExpressionPreprocessor::PexprPruneProjListProjectOrGbAgg
...
+	GPOS_ASSERT(NULL != pexpr);
+
+	COperator *pop = pexpr->Pop();
+	if (COperator::EopScalarCmp == pop->Eopid() || COperator::EopScalarIsDistinctFrom == pop->Eopid())
+	{
+		GPOS_ASSERT(2 == pexpr->UlArity());
+		CExpression *pexprLeft = (*pexpr)[0];
+		CExpression *pexprRight = (*pexpr)[1];
+
+		if (COperator::EopScalarConst == pexprLeft->Pop()->Eopid() && COperator::EopScalarIdent == pexprRight->Pop()->Eopid())
```
hsyuan: there are utility function `CUtils::FScalarIdent` and `CUtils::FScalarConst`

vraghavan78: On second thought what about cast?

bhuvnesh2703:
@vraghavan78  as of now it does not look for casts. Some of the cast cases are handled prior to preprocessing so queries as on https://github.com/greenplum-db/gpdb/pull/2734/files take benefit from it as the explicit case is already removed, but not all.

There are couple of cases of cast, for instance. Let me know what all you want to be handled.
```
Case 1:
create table foo (a int, b text);
explain select * from foo where 1 = b::int;

Algebrized query:
+--CLogicalSelect
   |--CLogicalGet "foo" ("foo"), Columns: ["a" (0), "b" (1), "ctid" (2), "xmin" (3), "cmin" (4), "xmax" (5), "cmax" (6), "tableoid" (7), "gp_segment_id" (8)] Key sets: {[2,8]}
   +--CScalarCmp (=)
      |--CScalarConst (1)
      +--CScalarCoerceViaIO
         +--CScalarIdent "b" (1)

Case 2:
explain select * from foo where 1 = b::int2;
Algebrized query:
+--CLogicalSelect
   |--CLogicalGet "foo" ("foo"), Columns: ["a" (0), "b" (1), "ctid" (2), "xmin" (3), "cmin" (4), "xmax" (5), "cmax" (6), "tableoid" (7), "gp_segment_id" (8)] Key sets: {[2,8]}
   +--CScalarCmp (=)
      |--CScalarConst (1)
      +--CScalarFunc (int2)
         +--CScalarIdent "b" (1)


Case 3:
explain select * from foo where 1 = b::int8;
Algebrized query:
+--CLogicalSelect
   |--CLogicalGet "foo" ("foo"), Columns: ["a" (0), "b" (1), "ctid" (2), "xmin" (3), "cmin" (4), "xmax" (5), "cmax" (6), "tableoid" (7), "gp_segment_id" (8)] Key sets: {[2,8]}
   +--CScalarCmp (=)
      |--CScalarConst (1)
      +--CScalarCast
         +--CScalarIdent "b" (1)
```

libgpopt/src/base/CUtils.cpp
```diff
@@ -6448,4 +6448,43 @@ CUtils::ExecLocalityType
 	return eelt;
 }
 
+// get imdid of the corresponding commutativity operator from metadata accessor,
```
d: Please don't write the exact same comment twice. Duplicate things will diverge without you even noticing.

libgpopt/src/base/CUtils.cpp
```diff
@@ -6448,4 +6448,43 @@ CUtils::ExecLocalityType
...
+	GPOS_ASSERT(COperator::EopScalarCmp == pop->Eopid() || COperator::EopScalarIsDistinctFrom == pop->Eopid());
+	CScalarCmp *popScalarCmp = NULL;
+
+	if (COperator::EopScalarCmp == pop->Eopid())
+	{
+		popScalarCmp = CScalarCmp::PopConvert(pop);
+	}
+	else if (COperator::EopScalarIsDistinctFrom == pop->Eopid())
+	{
+		popScalarCmp = dynamic_cast<CScalarCmp*>(CScalarIsDistinctFrom::PopConvert(pop));
```
d: Redundant cast

libgpopt/src/operators/CScalarCmp.cpp
```diff
@@ -179,6 +180,50 @@ CScalarCmp::Eber
...
+	(
+	CMDAccessor *pmda,
+	COperator *pop
+	)
+{
+	CScalarCmp *popScalarCmp = dynamic_cast<CScalarCmp *>(pop);
+	const IMDScalarOp *pmdScalarCmpOp = pmda->Pmdscop(popScalarCmp->PmdidOp());
+	IMDId *pmdidScalarCmpCommute = pmdScalarCmpOp->PmdidOpCommute();
+	return pmdidScalarCmpCommute;
+	}
```
vraghavan78: * move }
* new line above pmdidScalarCmpCommute

libgpopt/src/operators/CPredicateUtils.cpp
```diff
@@ -2215,20 +2215,15 @@ CPredicateUtils::PexprIndexLookupKeyOnRight
 	if (CUtils::FScalarCmp(pexprScalar))
 	{
 		CScalarCmp *popScCmp = CScalarCmp::PopConvert(pexprScalar->Pop());
-		
-		const IMDScalarOp *pmdscalarop = pmda->Pmdscop(popScCmp->PmdidOp());
-		IMDId *pmdidscalaropCommute = pmdscalarop->PmdidOpCommute();
+		CScalarCmp *popScCmpCommute = popScCmp->PopCommutedOp(pmp, pexprScalar->Pop());
 
-		if (pmdidscalaropCommute->FValid())
+		if (popScCmpCommute)
```
vraghavan78: please use FValid. That way we will check invalid MDId

libgpopt/include/gpopt/base/CUtils.h
```diff
@@ -1075,6 +1075,10 @@ namespace gpopt
 			// get execution locality
 			static
 			EExecLocalityType ExecLocalityType(CDistributionSpec *pds);
+
+			// check if expression is scalar IDF comparison
+			static
+			BOOL FScalarIsDistinctFrom(CExpression *pexpr);
```
vraghavan78: why not use CPredicateUtils::FIDF

libgpopt/src/base/CUtils.cpp
```diff
@@ -6448,4 +6448,14 @@ CUtils::ExecLocalityType
 	return eelt;
 }
 
+// check if given expression is a scalar is distinct form expression
+BOOL
+CUtils::FScalarIsDistinctFrom
```
vraghavan78: no need for this function

libgpopt/src/operators/CPredicateUtils.cpp
vraghavan78: ignore my comment

PR #199, opened 2017-07-07
Enable GPORCA to generate better plans for non-correlated exists subquery in the WHERE clause

vraghavan78:
Consider the following exists subquery, `(select * from bar)`. GPORCA generates an elaborate count based implementation of this subquery. If bar is a fact table, the count is going to be expensive.

```
vraghavan=# explain select * from foo where foo.a = foo.b and exists (select * from bar);
                                                    QUERY PLAN
 Gather Motion 3:1  (slice3; segments: 3)  (cost=0.00..1368262.79 rows=400324 width=8)
   ->  Nested Loop  (cost=0.00..1368250.86 rows=133442 width=8)
         Join Filter: true
         ->  Table Scan on foo  (cost=0.00..461.91 rows=133442 width=8)
               Filter: a = b
         ->  Materialize  (cost=0.00..438.57 rows=1 width=1)
               ->  Broadcast Motion 1:3  (slice2)  (cost=0.00..438.57 rows=3 width=1)
                     ->  Result  (cost=0.00..438.57 rows=1 width=1)
                           Filter: (count((count()))) > 0::bigint
                           ->  Aggregate  (cost=0.00..438.57 rows=1 width=8)
                                 ->  Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..438.57 rows=1 width=8)
                                       ->  Aggregate  (cost=0.00..438.57 rows=1 width=8)
                                             ->  Table Scan on bar  (cost=0.00..437.95 rows=332395 width=1)
 Optimizer status: PQO version 2.35.1
(14 rows)
```
Planner on the other hand uses LIMIT as shown in the INIT plan.

```
vraghavan=# explain select * from foo where foo.a = foo.b and exists (select * from bar);
                                           QUERY PLAN
 Gather Motion 3:1  (slice2; segments: 3)  (cost=0.03..13611.14 rows=1001 width=8)
   ->  Result  (cost=0.03..13611.14 rows=334 width=8)
         One-Time Filter: $0
         InitPlan  (slice3)
                 ->  Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..0.03 rows=1 width=0)
                             ->  Seq Scan on bar  (cost=0.00..11072.84 rows=332395 width=0)
         ->  Seq Scan on foo  (cost=0.00..13611.11 rows=334 width=8)
               Filter: a = b
 Settings:  optimizer=off
(12 rows)
```

While GPORCA doesnot support init-plan, we can nevertheless generate a better plan by using LIMIT instead of count. After this PR, GPORCA will generate the following plan with LIMIT clause.

```
vraghavan=# explain select * from foo where foo.a = foo.b and exists (select * from bar);
                                                 QUERY PLAN
 Gather Motion 3:1  (slice3; segments: 3)  (cost=0.00..1368262.73 rows=400324 width=8)
   ->  Nested Loop EXISTS Join  (cost=0.00..1368250.80 rows=133442 width=8)
         Join Filter: true
         ->  Table Scan on foo  (cost=0.00..461.91 rows=133442 width=8)
               Filter: a = b
         ->  Materialize  (cost=0.00..438.57 rows=1 width=1)
               ->  Broadcast Motion 1:3  (slice2)  (cost=0.00..438.57 rows=3 width=1)
                     ->  Limit  (cost=0.00..438.57 rows=1 width=1)
                           ->  Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..438.57 rows=1 width=1)
                                 ->  Limit  (cost=0.00..438.57 rows=1 width=1)
                                       ->  Table Scan on bar  (cost=0.00..437.95 rows=332395 width=1)
 Optimizer status: PQO version 2.35.1
(12 rows)
```

In https://github.com/greenplum-db/gporca/pull/193, I tried doing it in the pre-processing step. But the feedback was right that such an approach violated Orca's design principle that pre-processing does not introduce plan alternatives.

In this attempt, I have added it as part of the subquery handler. While it does not reduce the search space like https://github.com/greenplum-db/gporca/pull/193, it does however generate the better plan.


libgpopt/include/gpopt/base/CUtils.h
```diff
@@ -1075,6 +1075,9 @@ namespace gpopt
 			// get execution locality
 			static
 			EExecLocalityType ExecLocalityType(CDistributionSpec *pds);
+
+            // generate a limit expression on top of the given relational child with the given offset and limit count
```
bhuvnesh2703: identation

server/src/unittest/gpopt/minidump/CExistsSubqueryTest.cpp
```diff
@@ -0,0 +1,93 @@
...
+	return eres;
+}
+
+//---------------------------------------------------------------------------
+//	@function:
+//		CExistsSubqueryTest::EresUnittest_RunTests
+//
+//	@doc:
+//		Run all Minidump-based tests with plan matching
+//
```
bhuvnesh2703: planning to keep single line comment or these blocks?

libgpopt/src/xforms/CSubqueryHandler.cpp
```diff
@@ -1649,12 +1649,23 @@ CSubqueryHandler::FRemoveExistentialSubquery
...
 		if (COperator::EopScalarSubqueryExists == eopid)
 		{
+			CColRefSet *pcrsOuterRefs = pdpInner->PcrsOuter();
+
+			if (0 == pcrsOuterRefs->CElements())
+			{
+				// add a limit operator on top of the inner child if the subquery does not have
+				// any outer references. Adding Limit for the correlated case hinders pulling up
+				// predicates into an EXISTS join
+				pexprInner = CUtils::PexprLimit(pmp, pexprInner, 0, 1);
```
dhanashreek89: We need not add an extra Limit if the EXISTS sublink already contains one:

```
pivotal=# explain select * from foo where foo.a = foo.b and exists (select * from x limit 1);
                                                    QUERY PLAN
 Gather Motion 3:1  (slice3; segments: 3)  (cost=0.00..1324032.44 rows=4 width=8)
   ->  Nested Loop EXISTS Join  (cost=0.00..1324032.44 rows=2 width=8)
         Join Filter: true
               Filter: a = b
               ->  Broadcast Motion 1:3  (slice2)  (cost=0.00..431.00 rows=3 width=1)
                     ->  Limit  (cost=0.00..431.00 rows=1 width=1)
                           ->  Limit  (cost=0.00..431.00 rows=1 width=1)
                                       ->  Table Scan on x  (cost=0.00..431.00 rows=1 width=1)
 Settings:  optimizer=on
 Optimizer status: PQO version 2.35.1
(13 rows)
```

Can we skip the above step if `pexprInner` is already `EopLogicalLimit` ?

hsyuan: 👍

oarap:
For the future, I would suggest not to mix comment reformatting commits into actual PR. We should create a separate PR for comment reformatting. It makes it hard to review.

libgpopt/src/xforms/CSubqueryHandler.cpp
vraghavan78: That needs to be handled by a different xform.

server/src/unittest/gpopt/minidump/CExistsSubqueryTest.cpp
vraghavan78: will fix it.

PR #200, opened 2017-07-08
Add Check for existing tag before publishing

dhanashreek89:
Each ORCA commit must BUMP the version. If the version is not bumped new releases will not be pushed to the ORCA repository. This commit adds the check to validate the version of the current commit with the tag version existing on the repository.

### DIFF
--- a/concourse/check_existing_tag.py
+++ b/concourse/check_existing_tag.py
@@ -1,0 +1,24 @@
+#!/usr/bin/env python
+import subprocess
+import sys
+from packaging import version
+
+
+if __name__ == "__main__":
+  cmd = subprocess.Popen(['env', 'GIT_DIR=orca_src/.git', 'git', 'describe', '--abbrev=0', '--tags'], stdout=subprocess.PIPE)
+  existing_tag = cmd.stdout.read()
+  print ("Latest Tag of ORCA: {0}".format(existing_tag.strip()))
+
+  tag_to_be_published = ''
+  with open('orca_github_release_stage/tag.txt', 'r') as fh:
+    tag_to_be_published = fh.read().strip()
+
+  if tag_to_be_published == '':
+     sys.exit('Unable to read the tag from tag.txt')
+
+  print ("Tag from this commit: {0}".format(tag_to_be_published))
+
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ project(gpopt LANGUAGES CXX C)
 
 set(GPORCA_VERSION_MAJOR 2)
 set(GPORCA_VERSION_MINOR 35)
-set(GPORCA_VERSION_PATCH 1)
+set(GPORCA_VERSION_PATCH 2)
```
hsyuan: we can try to not bump version in this patch, to verify the correctness of this patch. :)

concourse/check_existing_tag.py
```diff
@@ -0,0 +1,24 @@
...
+from packaging import version
+
+
+if __name__ == "__main__":
+  cmd = subprocess.Popen(['env', 'GIT_DIR=orca_src/.git', 'git', 'describe', '--abbrev=0', '--tags'], stdout=subprocess.PIPE)
+  existing_tag = cmd.stdout.read()
+  print ("Latest Tag of ORCA: {0}".format(existing_tag.strip()))
+
+  tag_to_be_published = ''
+  with open('orca_github_release_stage/tag.txt', 'r') as fh:
```
hsyuan: where is the code that writes tag to file `orca_github_release_stage/tag.txt`?

hsyuan: OK, I found it. It is in `pipeline.yml`

CMakeLists.txt
oarap: Exactly 👍

khannaekta:
Removed the ORCA version increase from this PR and the pipeline failed as expected.

```
Latest Tag of ORCA: v2.35.1
Tag from this commit: v2.35.1
Tag v2.35.1 already present on ORCA repository
Please BUMP the ORCA version
```

d:
please use 4-space indentation for python scripts in the future ...

hawqtoolsmiths:
On that note - these python scripts fail pylint pretty bad.

Issue #196, opened 2017-07-03
On debian install error : unrecognized comand line option '-Wno-tautological-undefined-compare'


flochman:
I try compile gporca on debian stretch and after ninja install or make install /I tried both install commands from gpdb and gporca readme/ I got cc1plus error : : unrecognized comand line option '-Wno-tautological-undefined-compare' [-Werror]

feew lines befour this scope error is compile error:

/home/vric/gporca-master/libnaucrates/src/md/CMDIdGPDB.ccp:337:22: error nonull argument 'this' compared to NULL [-Werrorr-nonull-compare] 

return NULL != this && !FEquals (&CMDIdGPDB::m_mdidInvalidKey);
                               ^

So is no possible to compile gporca.

Please help and fix the error.

d:
What's your GCC version? The flag might have been renamed

flochman:
@d 

gcc 6.3.0 20170516 (Debian 6.3.0-18)

hsyuan:
Fixed by commit: https://github.com/greenplum-db/gporca/commit/ee5ef334c8a311d162df2c642ee159d6c9dfd7e8
Closing this issue.

jefffleming:
I'm seeing this issue in:
libgpopt/src/base/CCTEMap.cpp line 432.
    return (NULL == &cm) ? os : cm.OsPrint(os);
and 
libgpopt/src/base/CPartIndexMap.cpp line 1074
    return (NULL == &pim) ? os : pim.OsPrint(os);

This is on Debian with gcc 6.3.0 and gcc 7.2.0

PR #202, opened 2017-07-14
Remove all the "NULL == this" and "NULL != this"

hsyuan:
The only concern I have is the change of `FValid()`. There are so many callers of this function, not all of them have NULL check before calling `FValid()`.

Fixes issue #196.


hsyuan:
Applied feedback.
Most of the `FValid()` are called inside `GPOS_ASSERT()`, so no need to change that, it goes away in retail build anyway. Some has NULL check, for the remaining ones I changed to static function `IMDId::FValid()`. Please review again.

@oarap @vraghavan78

PR #206, opened 2017-07-26
Constant fold quantified subquery that compares a NULL constant to

vraghavan78:
scalar subquery

Consider two table foo (a, b) and bar (c, d) and the following SQL
query.

```
select * from bar where NULL in (select f2.a from foo f2);
+--CLogicalSelect
   |--CLogicalGet "bar" ("bar"), Columns: ["c" (0), "d" (1), "ctid" (2), "xmin" (3), "cmin" (4), "xmax" (5), "cmax" (6), "tableoid" (7), "gp_segment_id" (8)] Key sets: {[2,8]}
   +--CScalarSubqueryAny(=)["a" (9)]
      |--CLogicalGet "foo" ("foo"), Columns: ["a" (9), "b" (10), "ctid" (11), "xmin" (12), "cmin" (13), "xmax" (14), "cmax" (15), "tableoid" (16), "gp_segment_id" (17)] Key sets: {[2,8]}
      +--CScalarConst (null)
```

Here we are comparing a NULL constant to a result of subquery. Since
cannot to compared to any other value, the result of such a subquery
will always be false.

This CL adds the minimization of such quantified subqueries to false.

### DIFF
--- a/server/src/unittest/gpopt/minidump/CICGTest.cpp
+++ b/server/src/unittest/gpopt/minidump/CICGTest.cpp
@@ -40,6 +40,8 @@ ULONG CICGTest::m_ulNegativeIndexApplyTestCounter = 0;
 // minidump files
 const CHAR *rgszFileNames[] =
 	{
+		"../data/dxl/minidump/NULLComparisonInWhereClause.mdp",
+		"../data/dxl/minidump/NULLComparisonInProject.mdp",
 		"../data/dxl/minidump/ArrayCoerceExpr.mdp",
 		"../data/dxl/minidump/DisableLargeTableBroadcast.mdp",
 		"../data/dxl/minidump/InferPredicatesForLimit.mdp",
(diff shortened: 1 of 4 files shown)


server/src/unittest/gpopt/minidump/CICGTest.cpp
```diff
@@ -40,6 +40,8 @@ ULONG CICGTest::m_ulNegativeIndexApplyTestCounter = 0;
 // minidump files
 const CHAR *rgszFileNames[] =
 	{
+		"../data/dxl/minidump/NULLComparisonInWhereClause.mdp",
+		"../data/dxl/minidump/NULLComparisonInProject.mdp",
```
hsyuan: ICGTest is one of the most time-consuming test, can we move to other test suites?

vraghavan78: hold one. found a bug.

vraghavan78:
it should be constant folded to NULL and not false.

PR #207, opened 2017-07-27
Constant fold quantified subquery under SELECT clause

vraghavan78:
Consider two table foo (a, b) and bar (c, d) and the following SQL query.

```
select * from bar where NULL in (select f2.a from foo f2);
+--CLogicalSelect
   |--CLogicalGet "bar" ("bar"), Columns: ["c" (0), "d" (1), "ctid" (2), "xmin" (3), "cmin" (4), "xmax" (5), "cmax" (6), "tableoid" (7), "gp_segment_id" (8)] Key sets: {[2,8]}
   +--CScalarSubqueryAny(=)["a" (9)]
      |--CLogicalGet "foo" ("foo"), Columns: ["a" (9), "b" (10), "ctid" (11), "xmin" (12), "cmin" (13), "xmax" (14), "cmax" (15), "tableoid" (16), "gp_segment_id" (17)] Key sets: {[2,8]}
      +--CScalarConst (null)
```

Here we are comparing a NULL constant to a result of subquery. Since
cannot to compared to any other value, the result of such a subquery
will always be NULL.

### DIFF
--- a/data/dxl/minidump/NULLComparisonInProject.mdp
+++ b/data/dxl/minidump/NULLComparisonInProject.mdp
@@ -1,4 +1,10 @@
 <?xml version="1.0" encoding="UTF-8"?>
+<!--
+ create table foo(a int, b int);
+ create table bar(c int, d int);
+ insert into bar values (1,1);
+ select NULL in (select f2.a from foo f2) from bar;
+-->
 <dxl:DXLMessage xmlns:dxl="http://greenplum.com/dxl/2010/12/">
   <dxl:Thread Id="0">
     <dxl:OptimizerConfig>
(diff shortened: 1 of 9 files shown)


hsyuan:
```
hyuan=# explain select * from foo where NULL not in (select c from bar);
                                  QUERY PLAN
------------------------------------------------------------------------------
 Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..431.00 rows=5 width=8)
         Filter: NULL::boolean
 Settings:  optimizer=on
 Optimizer status: PQO version 2.39.1
(5 rows)

hyuan=# explain select * from foo where NULL in (select c from bar);
                                  QUERY PLAN
------------------------------------------------------------------------------
 Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..431.00 rows=5 width=8)
         Filter: NULL::boolean
 Settings:  optimizer=on
 Optimizer status: PQO version 2.39.1
(5 rows)
```

The 1st one should always be true, the 2nd one should always be false, right?

vraghavan78:
let me think on it.

vraghavan78:
This PR needs more thought. It is incomplete as it stands. closing PR.

PR #214, opened 2017-08-09
Adding a limit to the subquery when converting EXISTS/NOT EXISTS subquery to an aggregate

vraghavan78:
…uery to an aggregate

Prior to this CL, we converted an EXISTS and NOT EXISTS subquery
into an aggregate subquery, and checked the count values (count > 0 for
EXISTS, and count == 0 for NOT EXISTS).

In this CL, I merely added a limit below the count, since all we need to
check is the existence / non existence of a single tuple on the inner
side.

### DIFF
--- a/data/dxl/minidump/DynamicIndexGet-OuterRefs.mdp
+++ b/data/dxl/minidump/DynamicIndexGet-OuterRefs.mdp
@@ -1755,653 +1755,1036 @@ WHERE
         </dxl:Or>
       </dxl:LogicalJoin>
     </dxl:Query>
-	<dxl:Plan Id="0" SpaceSize="13701356">
-	  <dxl:GatherMotion InputSegments="0,1" OutputSegments="-1">
-	    <dxl:Properties>
-	      <dxl:Cost StartupCost="0" TotalCost="1111.256380" Rows="72.000000" Width="8"/>
-	    </dxl:Properties>
-	    <dxl:ProjList>
-	      <dxl:ProjElem ColId="26" Alias="cd_gender">
-	        <dxl:Ident ColId="26" ColName="cd_gender" TypeMdid="0.1042.1.0"/>
-	      </dxl:ProjElem>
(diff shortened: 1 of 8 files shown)


hsyuan:
We have to check whether the subquery has limit or not:
```sql
explain select * from foo where exists (select * from bar where c = 10);
                                                 QUERY PLAN
 Gather Motion 3:1  (slice3; segments: 3)  (cost=0.00..1324032.24 rows=5 width=8)
   ->  Nested Loop EXISTS Join  (cost=0.00..1324032.24 rows=2 width=8)
         Join Filter: true
               ->  Broadcast Motion 1:3  (slice2)  (cost=0.00..431.00 rows=3 width=1)
                     ->  Limit  (cost=0.00..431.00 rows=1 width=1)
                                 ->  Table Scan on bar  (cost=0.00..431.00 rows=1 width=1)
                                       Filter: c = 10
 Settings:  optimizer=on
 Optimizer status: PQO version 2.39.1
(12 rows)

explain select * from foo where exists (select * from bar where c = 10 limit 10);
                                                    QUERY PLAN
 Gather Motion 3:1  (slice3; segments: 3)  (cost=0.00..1324032.24 rows=5 width=8)
   ->  Nested Loop EXISTS Join  (cost=0.00..1324032.24 rows=2 width=8)
         Join Filter: true
               ->  Broadcast Motion 1:3  (slice2)  (cost=0.00..431.00 rows=3 width=1)
                     ->  Limit  (cost=0.00..431.00 rows=1 width=1)
                           ->  Limit  (cost=0.00..431.00 rows=1 width=1)
                                       ->  Table Scan on bar  (cost=0.00..431.00 rows=1 width=1)
                                             Filter: c = 10
 Settings:  optimizer=on
 Optimizer status: PQO version 2.39.1
(13 rows)
```

vraghavan78:
@oarap can you please take over this work? I will leave it to you to keep it or abandon it.

PR #218, opened 2017-08-25
Fix broken CAutoPTest & make pipeline green

dhanashreek89:
Previous to c09a0ac, `CStackObject()` constructor did a validity check if the pointer is on the stack using `FOnStack()`. Since the function is removed; the following test in `CAutoPTest` is invalid :

CAutoPTest::EresUnittest_Allocation()

This causes test failure in debug build. Hence this commit removes the test.

### DIFF
--- a/libgpos/server/src/unittest/gpos/common/CAutoPTest.cpp
+++ b/libgpos/server/src/unittest/gpos/common/CAutoPTest.cpp
@@ -34,12 +34,6 @@ CAutoPTest::EresUnittest()
 	CUnittest rgut[] =
 		{
 		GPOS_UNITTEST_FUNC(CAutoPTest::EresUnittest_Basics)
-#ifdef GPOS_DEBUG
-#if (GPOS_i386 || GPOS_i686 || GPOS_x86_64)
-		,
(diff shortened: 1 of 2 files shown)


dhanashreek89:
@hlinnaka @vraghavan78 @oarap  could you please take a look? Thanks!

libgpos/server/src/unittest/gpos/common/CAutoPTest.cpp
```diff
@@ -107,33 +101,5 @@ CAutoPTest::EresUnittest_Basics()
...
-//---------------------------------------------------------------------------
-//	@function:
-//		CAutoPTest::EresUnittest_Allocation
-//
-//	@doc:
-//		Attempt illegal allocation of auto pointer outside of stack
-//
-//---------------------------------------------------------------------------
-GPOS_RESULT
-CAutoPTest::EresUnittest_Allocation()
```
vraghavan78: can you remove the entry in the header too?

dhanashreek89: Removed from header. Thanks!

PR #223, opened 2017-09-08
Minidump sanitization: system column width update

oarap:
This PR does following sanitization for existing minidumps:
- Reformat the minidumps with `xmllint`
- Change the system column widths in `<dxl:ColumnStatistics>`
- Update the plan and cost changes to the affected minidumps


vraghavan78:
@oarap so do we need a translator change in the column width of the system columns? What does the planner estimate for these columns?

oarap:
No, we do not need to change anything in the code at the moment. It seems like we didn't even bother to ask for the actual width for those columns and assigned Orca's default width which is 8 byte. Now, we accurately get the system columns. There is a new function to retrieve stats for system columns in `CTranslatorRelcacheToDXL.cpp` which is `PdxlcolstatsSystemColumn`. It basically assigns the width based on the type size.

PR #226, opened 2017-09-09
Use col position from table descriptor for index key retrieval

jpatel-pivotal:
ORCA would crash when index keys were used as the position to retrieve `CColRef` in `CLogical::PosFromIndex()`. This is because index keys are relative to the relation which includes dropped columns if there are any dropped columns in that relation. The list of columns in `pdrgpcr` that the lookup happens from is relative to the table descriptor and does not include any dropped columns.

Consider the case where we had 20 columns in a table. We create an index that covers col number 20 as one of its keys. Then we drop columns 10 through 15. Now the index key still points to col number 20 but the column ref list in `pdrgpcr` will only have 15 elements in it and cause
ORCA to crash with an `Out of Bounds` exception when `CLogical::PosFromIndex()` gets called. 

This is because it was using the index keys as the position to do the retrieval. 

This commit fixes that issue by using the index key as the position to
retrieve the column from the relation. Uses the column's `attno` to get
the column's current position relative to the table descriptor `ulPos`.
Then uses this `ulPos` to retrieve the `CColRef` of the index key
column as shown below:

```
CColRef *pcr = (*pdrgpcr)[ulPos];
```

We also added the 2 test cases below to test for the above condition:

1. `DynamicIndexGetDroppedCols`
2. `LogicalIndexGetDroppedCols`

Both of the above test cases use a table with 30 columns; create a btree
index on 6 columns and then drop 7 columns so the table only has 23
columns remaining. 

This commit also bumps ORCA to version 2.43.0


libgpopt/src/operators/CLogical.cpp
```diff
@@ -149,30 +150,43 @@ CLogical::PdrgpdrgpcrCreatePartCols
...
 	const IMDIndex *pmdindex,
-	DrgPcr *pdrgpcr
+	DrgPcr *pdrgpcr,
+	const CTableDescriptor *ptabdesc,
+	const IMDRelation *pmdrel
 	)
 {
-	// compute the order spec
+	//
+	// compute the order spec after gettig the current position of the index key
```
vraghavan78: getting

libgpopt/src/operators/CLogical.cpp
```diff
@@ -149,30 +150,43 @@ CLogical::PdrgpdrgpcrCreatePartCols
...
+	// from the table descriptor. Index keys are relative to the
+	// relation. So consider a case where we had 20 columns in a table. We
+	// create an index that covers col # 20 as one of its keys. Then we drop
+	// columns 10 through 15. Now the index key still points to col #20 but the
+	// column ref list in pdrgpcr will only have 15 elements in it.
+	//
+
 	COrderSpec *pos = GPOS_NEW(pmp) COrderSpec(pmp);
 	const ULONG ulLenIncludeCols = pmdindex->UlKeys();
```
vraghavan78: rename this as ulLenKeys

libgpopt/src/operators/CLogical.cpp
```diff
@@ -149,30 +150,43 @@ CLogical::PdrgpdrgpcrCreatePartCols
...
 	for (ULONG  ul = 0; ul < ulLenIncludeCols; ul++)
 	{
-		ULONG ulPos = pmdindex->UlKey(ul);
+		// This is the postion of the index key column relative to the relation
+		ULONG ulPosWRTRel = pmdindex->UlKey(ul);
```
vraghavan78: ulPosRel

libgpopt/src/operators/CLogical.cpp
```diff
@@ -149,30 +150,43 @@ CLogical::PdrgpdrgpcrCreatePartCols
...
+
+		// Get the column and it's attno from the relation
```
vraghavan78: get

libgpopt/src/operators/CLogical.cpp
```diff
@@ -149,30 +150,43 @@ CLogical::PdrgpdrgpcrCreatePartCols
...
+		const IMDColumn *pmdcol = pmdrel->Pmdcol(ulPosWRTRel);
+		INT iAttno = pmdcol->IAttno();
```
vraghavan78: const

libgpopt/src/operators/CLogical.cpp
```diff
@@ -149,30 +150,43 @@ CLogical::PdrgpdrgpcrCreatePartCols
...
+
+		// This is the position of the index key column relative to the table descriptor
```
vraghavan78: get the position ....

libgpopt/src/operators/CLogical.cpp
```diff
@@ -149,30 +150,43 @@ CLogical::PdrgpdrgpcrCreatePartCols
...
+		ULONG ulPos = ptabdesc->UlPosition(iAttno);
```
vraghavan78: const ulPosTabDesc

libgpopt/src/operators/CLogicalIndexGet.cpp
```diff
@@ -88,8 +88,12 @@ CLogicalIndexGet::CLogicalIndexGet
 	// create the index descriptor
 	m_pindexdesc  = CIndexDescriptor::Pindexdesc(pmp, ptabdesc, pmdindex);
 
+	// Get relation from the metadata accessor using metadata id
+	CMDAccessor *pmda = COptCtxt::PoctxtFromTLS()->Pmda();
+	const IMDRelation *pmdrel = pmda->Pmdrel(ptabdesc->Pmdid());
```
vraghavan78: on second thought move the following code to CLogical.cpp

```
const IMDRelation *pmdrel = pmda->Pmdrel(ptabdesc->Pmdid());
```

libgpopt/include/gpopt/operators/CLogical.h
```diff
@@ -176,7 +176,9 @@ namespace gpopt
 				(
 				IMemoryPool *pmp,
 				const IMDIndex *pmdindex,
-				DrgPcr *pdrgpcr
+				DrgPcr *pdrgpcr,
+				const CTableDescriptor *ptabdesc,
+				const IMDRelation *pmdrel
```
vraghavan78: actually on second thought please just pass the ptabdesc and move the following to CLogical.cpp

```
const IMDRelation *pmdrel = pmda->Pmdrel(ptabdesc->Pmdid());
```

jpatel-pivotal: will revert it to the way I had it before.

libgpopt/src/operators/CLogical.cpp
jpatel-pivotal: Did you mean const ULONG ulPosRel?

vraghavan78: Yes

CMakeLists.txt
```diff
@@ -4,8 +4,8 @@ cmake_minimum_required(VERSION 3.0 FATAL_ERROR)
 project(gpopt LANGUAGES CXX C)
 
 set(GPORCA_VERSION_MAJOR 2)
-set(GPORCA_VERSION_MINOR 42)
-set(GPORCA_VERSION_PATCH 1)
+set(GPORCA_VERSION_MINOR 43)
```
oarap: If there is no corresponding changes on GPDB, only the patch version should be incremented.

jpatel-pivotal: Ok got it.

PR #227, opened 2017-09-11
Added handling for winstar & winagg fields

bhuvnesh2703:
With commit 387c485d62955fba79c9a376820aade32047eb45 winstar and winagg
fields were added in WindowRef Node, so this commit adds handling for
them in ORCA.


hsyuan:
What is WindowStarArg? Should it be WindowStarAgg?

bhuvnesh2703:
Yes Arg here refers argument as winstar == > TRUE if argument list was really '*'

PR #228, opened 2017-09-11
Stats width only

oarap:
GPORCA should not spend time extracting column statistics that are not needed for cardinality estimation. This PR eliminates this overhead of requesting and generating the statistics for columns that are not used in cardinality estimation unnecessarily.

E.g:
`CREATE TABLE foo (a int, b int, c int);`

For table foo, the query below only needs for stats for column `a` which is the distribution column and column `c` which is the column used in where clause.
    
`select * from foo where c=2;`

However, prior to that PR, the column statistics for column `b` is also calculated and passed for the cardinality estimation. The only information needed by the optimizer is the `width` of column `b`. For this tiny information, we transfer every stats information for that column.

This PR and its counterpart commit in GPDB ensures that the column width information is passed and extracted in the `dxl:Relation` metadata information.

Preliminary results for short running queries provides up to 65x performance improvement.

This change affect significant number of test files, in order to make the review process easier, commits should be reviewed separately. The commits are mutually exclusive.


libgpopt/src/mdcache/CMDAccessor.cpp
```diff
@@ -1164,6 +1158,25 @@ CMDAccessor::Pmdcolstats
 	return pmdcolstats;
 }
 
+// Return the column widths of a table with passed relation ID
+DrgPdouble *
+CMDAccessor::Pmdcolwidths
```
vraghavan78: move this function to CMDAccessorUtils since it is a utility function on a cached object

libgpopt/src/mdcache/CMDAccessor.cpp
```diff
@@ -1164,6 +1158,25 @@ CMDAccessor::Pmdcolstats
...
+	(
+	IMemoryPool *pmp,
+	IMDId *pmdidRel
+	)
+{
+	const IMDRelation *pmdrel = Pmdrel(pmdidRel);
+	ULONG ulColumns = pmdrel->UlColumns();
```
vraghavan78: const

vraghavan78:
I did not check the minidump but the code change looks correct. Good work

libgpopt/src/mdcache/CMDAccessor.cpp
oarap: CMDAccessorUtils provide different sets of utility functions and what `Pmdcolwidths` provides does not fit in that category. I will leave this as is.

libnaucrates/include/naucrates/md/CMDRelationCtasGPDB.h
```diff
@@ -96,6 +96,9 @@ namespace gpmd
 			// vartypemod list
 			DrgPi *m_pdrgpiVarTypeMod;
 
+			// array of column widths including dropped columns
```
vraghavan78: no dropped columns in CTAS I think

oarap:
merged

PR #230, opened 2017-09-13
GPORCA incorrectly collapsing FOJ with condition false into a contrad…

vraghavan78:
…iction that will not return any rows

Prior to this fix, the logic that calculated max cardinality for each
logical operator assumed that the Full Outer Join with condition false will always
return empty set. This was used by the following preprocessing step

CExpressionPreprocessor::PexprPruneEmptySubtrees

to eliminate the FOJ subtrees (since it thought had zero output cardinality),
replacing them with a const table get and zero tuples

```
vraghavan=# explain select * from foo a full outer join foo b on false;
                   QUERY PLAN
------------------------------------------------
   ->  Result  (cost=0.00..0.00 rows=0 width=8)
         One-Time Filter: false
 Optimizer status: PQO version 2.43.1
(4 rows)
```
This collapsing is incorrect. In this CL, the max cardinality logic has been
fixed to ensure that GPORCA plan generates correct plan.

After the fix:

```
vraghavan=# explain select * from foo a full outer join foo b on false;
                                                            QUERY PLAN
----------------------------------------------------------------------------------------------------------------------------------
 Gather Motion 3:1  (slice2; segments: 3)  (cost=0.00..2207585.75 rows=35 width=8)
   ->  Result  (cost=0.00..2207585.75 rows=12 width=8)
         ->  Sequence  (cost=0.00..2207585.75 rows=12 width=8)
               ->  Shared Scan (share slice:id 2:0)  (cost=0.00..431.00 rows=2 width=1)
                     ->  Materialize  (cost=0.00..431.00 rows=2 width=1)
                           ->  Table Scan on foo  (cost=0.00..431.00 rows=2 width=34)
               ->  Sequence  (cost=0.00..2207154.75 rows=12 width=8)
                     ->  Shared Scan (share slice:id 2:1)  (cost=0.00..431.00 rows=2 width=1)
                           ->  Materialize  (cost=0.00..431.00 rows=2 width=1)
                                 ->  Table Scan on foo  (cost=0.00..431.00 rows=2 width=34)
                     ->  Append  (cost=0.00..2206723.75 rows=12 width=8)
                           ->  Nested Loop Left Join  (cost=0.00..882691.26 rows=10 width=68)
                                 Join Filter: false
                                 ->  Shared Scan (share slice:id 2:0)  (cost=0.00..431.00 rows=2 width=34)
                                 ->  Result  (cost=0.00..0.00 rows=0 width=34)
                                       One-Time Filter: false
                           ->  Result  (cost=0.00..1324032.49 rows=2 width=68)
                                 ->  Nested Loop Left Anti Semi Join  (cost=0.00..1324032.49 rows=2 width=34)
                                       Join Filter: false
                                       ->  Shared Scan (share slice:id 2:1)  (cost=0.00..431.00 rows=2 width=34)
                                       ->  Materialize  (cost=0.00..431.00 rows=5 width=1)
                                             ->  Broadcast Motion 3:3  (slice1; segments: 3)  (cost=0.00..431.00 rows=5 width=1)
                                                   ->  Result  (cost=0.00..431.00 rows=2 width=1)
                                                         ->  Shared Scan (share slice:id 1:0)  (cost=0.00..431.00 rows=2 width=1)
 Optimizer status: PQO version 2.43.1
(25 rows)
```


hsyuan:
Sigh... Can you use full name of FOJ and CL?

hsyuan:
Just glance at the description, I don't understand why it is incorrect and what is the correct plan.

vraghavan78:
Updated the comment. Hope this gives u more insight of what I am talking

vraghavan78:
Merged thanks!

PR #231, opened 2017-09-13
Map GPOS severity level to GPDB Severity Levels

bhuvnesh2703:
GPOS raises exception with different severity level, but
they were being logged to GPDB logs at LOG severity level.
This disabled users to not turn off logging for GPOS exceptions, unless
GPDB log setting was changed higher than LOG severity level.

This is the initial commit which introduces the functionality. If an
exception is created without the GPDB severity level, it will default to
LOG severity level in GPDB.


vraghavan78:
@bhuvnesh2703 let us not target this for GPDB 4.3.17.0. Let us wait for the next 4.3 release along with 5.1

libgpos/src/error/CException.cpp
```diff
@@ -143,6 +163,39 @@ CException::Raise
 }
 
 
+void
+CException::Raise
```
oarap: Is it possible to refactor? Both version of Raise seems to be sharing significant code.

bhuvnesh2703: i think we cannot, as VA_START excepts the functions from which it is called to be a variable length and in some cases we want to send ulMinor and in some cases ulSeverityLevel, so would prefer keep it this way. Also i believe in future once we change all the calls for other GPOS_RAISE we can remove this other function completely.

libgpos/include/gpos/error/CException.h
```diff
@@ -22,6 +22,12 @@
 #define GPOS_RAISE(...) \
 	gpos::CException::Raise(__FILE__, __LINE__, __VA_ARGS__)
 
+// raises GPOS exception,
+// these exceptions can later be translated to GPDB log severity levels
+// so that they can be written in GPDB with appropriate severity level.
+#define GPOS_THROW_EXCEPTION(...) \
+gpos::CException::Raise(__FILE__, __LINE__, __VA_ARGS__)
```
dhanashreek89: Indentation is off here

PR #232, opened 2017-09-14
Incorrect Decorrelation results in wrong plan

bhuvnesh2703:
While attempting to decorrelate the subquery, we were
incorrectly pulling up the join before calculating the window function
of the results of the join. In cases where we have subqueries with
window function and the subquery has outer references we should not
attempt decorrelating it.

There are further optimization which can be done in case of existential
queries but this PR fixes the plan.


libgpopt/src/xforms/CDecorrelator.cpp
```diff
@@ -680,9 +680,12 @@ CDecorrelator::FProcessProject
 		CExpressionHandle exprhdl(pmp);
 		exprhdl.Attach(pexpr);
 		exprhdl.DeriveProps(NULL /*pdpctxt*/);
-		if (CLogicalSequenceProject::PopConvert(pexpr->Pop())->FHasLocalOuterRefs(exprhdl))
+		// fail if a relational child of LogicalSequenceProject has outer references or
+		// if a SequenceProject has local outer references
+		// Ex: select C.j from C where C.i in (select rank() over (order by B.i) from B where B.i=C.i) order by C.j;
+		// The subquery has outer references and result of window function is projected from subquery
```
dhanashreek89: Could you reword this comment a bit? `The subquery has outer references and result of window function is projected from subquery` is this part a continuation of the first part ?

libgpopt/src/xforms/CDecorrelator.cpp
```diff
@@ -680,9 +680,12 @@ CDecorrelator::FProcessProject
...
+		if (CLogicalSequenceProject::PopConvert(pexpr->Pop())->FHasLocalOuterRefs(exprhdl) || CUtils::FHasOuterRefs((*pexpr)[0]))
```
vraghavan78: what does the first function do?

bhuvnesh2703: it checks if the logical sequence project node has any outer references in the order by, partition over and window frame.

oarap:
Can you add the example in the commit message as well?

vraghavan78:
what if you have the following query:

```sql
create table foo(a int, b int, c int);
create table bar(d int, e int,  f int);
explain select * from foo where foo.c in (select avg(bar.e) over (partition by bar.f) from bar where bar.f = foo.b)
```

bhuvnesh2703:
@vraghavan78 in this query as well, we will not de-correlate it currently. However, it could be decorrelated. Will create a story for it.

PR #237, opened 2017-09-22
Push ORCA src to bintray repository for conan

bhuvnesh2703:
ORCA source will be pushed to bintray repository for each version bump.
Developers can use conan dependency manager to build orca before
building GPDB via minimal steps and need not worry about the version
dependency.
As we currently bump orca version used by GPDB, we will ensure that 
`conanfile.txt` gets updated for each bump of ORCA so that someone 
who downloads gpdb need not worry to get the specific orca version, 
conan will do that.

To build & install orca from GPDB:
Step 1: Ensure that xerces-c is already installed (open source works)
Step 2: cd <path_to_gpdb>/depends
Step 3: env CONAN_CMAKE_GENERATOR=Ninja conan install -s build_type=Debug --build
where: build_type can be `Debug` or `Release`

It will install the files to `/usr/local/include` `/usr/local/lib` directories as currently we expect for GPDB.


concourse/upload_orca_source_bintray.yml
```diff
@@ -0,0 +1,20 @@
+platform: linux
+image_resource:
+  type: docker-image
+  source:
+    repository: bhuvnesh2703/conan
```
bhuvnesh2703: i will push this to pivotaldata/conan

PR #239, opened 2017-09-25
Support `IS DISTINCT FROM const` expressions in static partition elimination

hardikar:
Given the DDL: 
```
CREATE TABLE pt (id int, gender text)  
DISTRIBUTED BY (id) 
PARTITION BY LIST (gender) 
( PARTITION girls VALUES ('F', NULL),    
PARTITION boys VALUES ('M'),    
DEFAULT PARTITION other );
```

ORCA would ignore IDFs. With this PR, it converts it to an equivalent range expression and NULL test and enable partition selection.
```
shardikar=# explain select * from pt where gender is distinct from 'M';
                                            QUERY PLAN
--------------------------------------------------------------------------------------------------
 Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..431.00 rows=1 width=12)
   ->  Sequence  (cost=0.00..431.00 rows=1 width=12)
         ->  Partition Selector for pt (dynamic scan id: 1)  (cost=10.00..100.00 rows=34 width=4)
               Filter: gender < 'M'::text OR gender > 'M'::text OR gender IS NULL
               Partitions selected: 2 (out of 3)
         ->  Dynamic Table Scan on pt (dynamic scan id: 1)  (cost=0.00..431.00 rows=1 width=12)
               Filter: gender IS DISTINCT FROM 'M'::text
 Optimizer status: PQO version 2.43.1
(8 rows)
```

*Why do we need to convert an IDF to a range?*
Each partition in a partitioned table corresponds to a PartitionRule, which is a check-constraint on the tuples in that partition. E.g given the DDL above, for the table boys, the constraint is `gender = 'M'`. Static partition elimination for list partitions uses an ARRAYOPEXPR. It compares the list of values to the scalar expression from the query to determine if the given partition needs to be selected. 
E.g. for the query `select * from pt where gender = 'M'`, to determine if `boys` should be selected we generate an ARRAYOPEXPR expression : `'M' = {'M'}`. Unfortunately, there is no equivalent test for doing a IS DISTINCT FROM comparison. For that reason, `var IS DISTINCT FROM const` must be converted to a `var != const OR bar IS NULL`. ORCA converts `var != const` to `var < const OR var > const`.

This PR also works for queries with a `IS DISTINCT FROM NULL` and corresponding `IS NOT DISTINCT FROM` varieties.

*Removing redundant IS NOT NULL filters*
`IS DISTINCT FROM NULL` generates an `IS NOT NULL` filter in `CPredicateUtils::PexprPartPruningPredicate()`. However, this is redundant in some cases (e.g `a IS NOT NULL OR a < 5`) and in turn prevents ORCA/GPDB from using `partEqFilters` for partition elimination. So the PR checks and removes the `IS NOT NULL` filter in case it is redundant.

There are a few changes in the mdp files. Most seem beneficial since partition elimination is now possible (because of an added filter), when it wasn't possible before. This is not exclusively in IDF queries - since the `ConstraintInterval` method is shared for other kinds of WHERE clauses.

In GPDB, only one ICG fails - partpruning - because of static partition elimination.

-Shreedhar & Ekta.


libgpopt/src/operators/CPredicateUtils.cpp
```diff
@@ -1312,6 +1326,38 @@ CPredicateUtils::FNullCheckOnColumn
...
+		 * |--CScalarNullTest
+		 * |  |--CScalarIdent
+		 * Hence the Ident is the pexpr child's child
+		 */
+		CExpression *pexprChild = (*pexpr)[0];
+		CExpression *pexprIdent = (*pexprChild)[0];
+		return (CUtils::FScalarIdent(pexprIdent, pcr) || CCastUtils::FBinaryCoercibleCastedScId(pexprIdent, pcr));
+	}
+
+	return false;
```
hsyuan: You can simplify this function body as
`FScalarBoolOp(pexpr, CScalarBoolOp::EboolopNot) && FNullCheckOnColumn(pexpr, pcr)`

libgpopt/src/base/CConstraintInterval.cpp
```diff
@@ -430,6 +434,60 @@ CConstraintInterval::PciIntervalFromScalarCmp
...
+		// constant
+		CScalarConst *popScConst = CScalarConst::PopConvert((*pexpr)[1]->Pop());
+		// operator
+		CScalarIsDistinctFrom *popScCmp = CScalarIsDistinctFrom::PopConvert(pexpr->Pop());
+
+		GPOS_ASSERT (CScalar::EopScalarConst == popScConst->Eopid());
+		GPOS_ASSERT (IMDType::EcmptIDF == popScCmp->Ecmpt());
+
+		IDatum *pdatum = popScConst->Pdatum();
+		CConstraintInterval *pcri = NULL, *pcriChild;
```
hsyuan: pcriChild is only used when pdatum->FNull() is true, why not move there?

libgpopt/src/base/CConstraintInterval.cpp
```diff
@@ -430,6 +434,60 @@ CConstraintInterval::PciIntervalFromScalarCmp
...
+
+		if (pdatum->FNull())
+		{
+			// col IS DISTINCT FROM NULL
+			pcriChild = GPOS_NEW(pmp)
+							CConstraintInterval(pmp, pcr, GPOS_NEW(pmp) DrgPrng(pmp), true /*fIncludesNull*/);
+			pcri = pcriChild->PciComplement(pmp);
+			pcriChild->Release();
```
hsyuan: I haven't looked at the implementation of `PciComplement`. But I am curious what kind of Interval will represent `col is not null`?

Will the constructor CConstraintInterval create interval contains NULL only, and PciComplement return interval with any range rather than NULL?

hardikar: `col is not null` is : (-inf, inf) with m_fIncludesNull = false

> Will the constructor CConstraintInterval create interval contains NULL only, and PciComplement return interval with any range rather than NULL?

Yup.

libgpopt/src/operators/CPredicateUtils.cpp
```diff
@@ -1225,6 +1225,20 @@ CPredicateUtils::PexprPartPruningPredicate
 		
 	}
 
+	// Remove an "IS NOT NULL" partition filter if it is in redundant e.g. "a = b AND a is NOT NULL"
```
hardikar: grammar if it is in redundant

PR #241, opened 2017-09-27
Enable partition selection when a cast is present

hardikar:
Enabled by including CScalarCmp expressions that contain a CScalarCast
in the types of expressions considered for partition filters. Observe
that if the expression contains a CScalarCast over a CScalarIdent, it
must be preserved all the way to the final plan. That is, we cannot
"peak" and extract the identifer under the cast. For this reason, in
case it is an equality comparison with a cast, levelEqExprs can no
longer be used.

Also, refactor SplitPartPredicates() to make it easier to read.

```
create table R (a int, b int) partition by range(b) (start(1) end(10) every (3));
shardikar=# explain select * from R where b::int8 = 5::int8;
                                           QUERY PLAN
 Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..431.00 rows=1 width=8)
         ->  Partition Selector for r (dynamic scan id: 1)  (cost=10.00..100.00 rows=34 width=4)
               Filter: b::bigint = 5::bigint
               Partitions selected: 1 (out of 3)
         ->  Dynamic Table Scan on r (dynamic scan id: 1)  (cost=0.00..431.00 rows=1 width=8)
               Filter: b::bigint = 5::bigint
 Optimizer status: PQO version 2.46.3
(8 rows)
```

- Added a new mdp file to test this case (and that partition selection is happening)
- There are not corresponding changes in GPDB.
- But, I found no relevant tests in GPDB. These need to be added, and I can create a PR for those once we're close to finalizing this PR.
- Casting to text won't work because that uses a different kind of cast : CoerceViaIO.

### DIFF
--- a/libgpopt/include/gpopt/operators/CPredicateUtils.h
+++ b/libgpopt/include/gpopt/operators/CPredicateUtils.h
@@ -427,6 +427,15 @@ namespace gpopt
 					IMDType::ECmpType *pecmpt
 					);
 
+			// Expression is a comparison with a simple identifer on at least one side
+			static
+			BOOL FIdentCompare
+				(
+				CExpression *pexpr,
+				IMDType::ECmpType pecmpt,
+				CColRef *pcr
+				);
+
 			// check if given column is a constant
 			static
 			BOOL FConstColumn(CConstraint *pcnstr, const CColRef *pcr);
--- a/libgpopt/src/base/CPartitionPropagationSpec.cpp
+++ b/libgpopt/src/base/CPartitionPropagationSpec.cpp
@@ -340,16 +340,36 @@ CPartitionPropagationSpec::SplitPartPredicates
 		// find conjuncts for this key and mark their positions
 		DrgPexpr *pdrgpexprKey = PdrgpexprPredicatesOnKey(pmp, pdrgpexprConjuncts, pcr, pcrsKeys, &pbsUsed);
 		const ULONG ulLen = pdrgpexprKey->UlLength();
-		if (0 == ulLen)
+		if (ulLen == 0)
 		{
 			// no predicates on this key
 			pdrgpexprKey->Release();
 			continue;
 		}
 
-		if (1 < ulLen || (!CPredicateUtils::FEquality((*pdrgpexprKey)[0])))
+		if (ulLen == 1 && CPredicateUtils::FIdentCompare((*pdrgpexprKey)[0], IMDType::EcmptEq, pcr))
 		{
-			// more than one predicate on this key or one non-equality predicate
+			// EqFilters
+			// one equality predicate (key = expr); take out the expression
+			// and add it to the equality filters map
+			CExpression *pexprPartKey = NULL;
+			CExpression *pexprOther = NULL;
+			IMDType::ECmpType ecmpt = IMDType::EcmptOther;
+
+			CPredicateUtils::ExtractComponents((*pdrgpexprKey)[0], pcr, &pexprPartKey, &pexprOther, &ecmpt);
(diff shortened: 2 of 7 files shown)


libgpopt/include/gpopt/operators/CPredicateUtils.h
```diff
@@ -427,6 +427,15 @@ namespace gpopt
 					IMDType::ECmpType *pecmpt
 					);
 
+			// Expression is a comparison with a simple identifer on at least one side
```
vraghavan78: E -> e

libgpopt/include/gpopt/operators/CPredicateUtils.h
```diff
@@ -427,6 +427,15 @@ namespace gpopt
...
+			static
+			BOOL FIdentCompare
```
vraghavan78: FIdentCompareIgnoreCast 

We have other such functions

libgpopt/include/gpopt/operators/CPredicateUtils.h
```diff
@@ -427,6 +427,15 @@ namespace gpopt
...
+				(
```
vraghavan78: will this work for 1::int8 is distinct from a::int8?

libgpopt/src/base/CPartitionPropagationSpec.cpp
```diff
@@ -340,16 +340,36 @@ CPartitionPropagationSpec::SplitPartPredicates
 		// find conjuncts for this key and mark their positions
 		DrgPexpr *pdrgpexprKey = PdrgpexprPredicatesOnKey(pmp, pdrgpexprConjuncts, pcr, pcrsKeys, &pbsUsed);
 		const ULONG ulLen = pdrgpexprKey->UlLength();
-		if (0 == ulLen)
+		if (ulLen == 0)
```
vraghavan78: inside Orca we have the convention CONST CompareOp Variable

libgpopt/include/gpopt/operators/CPredicateUtils.h
hardikar: Yes but this is checking something else.

libgpopt/include/gpopt/operators/CPredicateUtils.h
hardikar: I don't think so. This commit needs to get rebase on the other PR for that to work.

libgpopt/src/operators/CPredicateUtils.cpp
```diff
@@ -627,22 +631,59 @@ CPredicateUtils::ExtractComponents
 	IMDType::ECmpType ecmpt =
 			CScalarCmp::PopConvert(pexprScCmp->Pop())->Ecmpt();
 
-	if (CUtils::FScalarIdent(pexprLeft, pcrKey) || CCastUtils::FBinaryCoercibleCastedScId(pexprLeft, pcrKey))
+	if (CUtils::FScalarIdent(pexprLeft, pcrKey) ||
+		CScalarIdent::FCastedScId(pexprLeft, pcrKey) ||
+		CCastUtils::FBinaryCoercibleCastedScId(pexprLeft, pcrKey))
```
hardikar: Note to myself : FCastedScId will subsumes FBinaryCoercibleCastedScId.

hardikar:
Thanks for all the reviews. I had found an issue with these changes for the query:
```explain select * from R where b::int8 < 5::numeric;```.

This PR succeeded by using a int4 to int8 comparison. That is it didn't actually cast the partition key. For that reason, I'm closing this PR, and opening a new one with the fixes.

PR #242, opened 2017-10-02
Enable partition selection when a cast is present

hardikar:
Enabled by including CScalarCmp expressions that contain a CScalarCast in the types of expressions considered for partition filters. Observe that if the expression contains a CScalarCast over a CScalarIdent, it must be preserved all the way to the final plan. That is, we cannot
"peak" and extract the identifer under the cast. For this reason, in case it is an equality comparison with a cast, levelEqExprs can no longer be used. 
In case of list partitions, a ScalarArrayCoerceExpr is created on top of the PartListValues. This is already supported in GPDB 5. We will open a separate PR in GPDB 4 to convert `ArrayCoerceExpr` to a `FuncExpr` or `array_coerce`.

Also, refactor SplitPartPredicates() to make it easier to read.

```
create table R (a int, b int) partition by range(b) (start(1) end(10) every (3));
shardikar=# explain select * from R where b::int8 = 5::int8;
                                           QUERY PLAN
 Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..431.00 rows=1 width=8)
         ->  Partition Selector for r (dynamic scan id: 1)  (cost=10.00..100.00 rows=34 width=4)
               Filter: b::bigint = 5::bigint
               Partitions selected: 1 (out of 3)
         ->  Dynamic Table Scan on r (dynamic scan id: 1)  (cost=0.00..431.00 rows=1 width=8)
               Filter: b::bigint = 5::bigint
 Optimizer status: PQO version 2.46.3
(8 rows)
```

- Added a new mdp file to test this case (and that partition selection is happening)
- There are no corresponding changes in GPDB. Only the version in GPDB needs to be updated.
- Casting to text won't work because that uses a different kind of cast : CoerceViaIO.


libgpopt/src/base/CPartitionPropagationSpec.cpp
```diff
@@ -340,16 +340,36 @@ CPartitionPropagationSpec::SplitPartPredicates
...
 		{
-			// more than one predicate on this key or one non-equality predicate
+			// EqFilters
+			// one equality predicate (key = expr); take out the expression
+			// and add it to the equality filters map
+			CExpression *pexprPartKey = NULL;
+			CExpression *pexprOther = NULL;
+			IMDType::ECmpType ecmpt = IMDType::EcmptOther;
+
+			CPredicateUtils::ExtractComponents((*pdrgpexprKey)[0], pcr, &pexprPartKey, &pexprOther, &ecmpt);
```
bhuvnesh2703: question: do we now have a dependency on FIdentCompare before using ExtractComponents? 
In FIdentCompare we validate if the expr is a binary coercible expression, and in ExtractComponents we take the keys if there is a cast on it. Will there be use cases where we cannot use ExtractComponents directly and have to check before if they are binary coercible by calling FIdentCompare?

dhanashreek89: `ExtractComponents` now checks for the over all cast and not specifically for BinaryCoercibleCast. So if you want to exclude the BinaryCoercibleCasts then that needs to be handled in the caller function. As of now; ExtractComponents is being specifically used for partition elimination.

libgpopt/src/translate/CTranslatorExprToDXLUtils.cpp
```diff
@@ -780,48 +785,130 @@ CTranslatorExprToDXLUtils::PdxlnRangePointPredicate
...
+					pdxlnPartKeyIdent
+					);
+	}
+	else
+	{
+		// Not supported - should be unreachable.
+		CWStringDynamic *pstr = GPOS_NEW(pmp) CWStringDynamic(pmp);
+		pstr->AppendFormat(GPOS_WSZ_LIT("Unsupported part filter operator for list partitions : %ls"),
+						   pexprPartKey->Pop()->SzId());
+		GPOS_RAISE(gpopt::ExmaGPOPT, gpopt::ExmiUnsupportedOp, pstr->Wsz());
```
bhuvnesh2703: you might want to use GPOS_THROW_EXCEPTION instead of GPOS_RAISE. Not mandatory.
GPOS_THROW_EXCEPTION enables to log the orca message with either debug1 or log level in gpdb logs. Earlier we got a story in which the customer complained that we log the fallback message at LOG level and it fills up the log. The above message will also be logged currently at LOG level

libgpopt/src/base/CPartitionPropagationSpec.cpp
bhuvnesh2703: Will this impact CTranslatorExprToDXL::PdxlnScCmpPartKey call for extract components?
CPredicateUtils::ExtractComponents(pexprScCmp, pcrPartKey, &pexprPartKey, &pexprOther, &ecmpt);

hardikar: No, it shouldn't. Do you see any particular scenario?

libgpopt/src/translate/CTranslatorExprToDXLUtils.cpp
hardikar: Hmm I'm not sure how important this error message in - but sure I can change this to use `GPOS_THROW_EXCEPTION`

libgpopt/src/operators/CPredicateUtils.cpp
```diff
@@ -1087,6 +1087,39 @@ CPredicateUtils::FCompareIdentToConst
 }
 
 // is the given expression is of the form (col IS DISTINCT FROM const)
+BOOL
+CPredicateUtils::FIdentIDFConst
+(
```
bhuvnesh2703: indentation

PR #244, opened 2017-10-06
Support subqueries in the Scalar child of ANY/ALL Subqueries

dhanashreek89:
While un-nesting the ANY/ALL subqueries, ORCA generates a predicate to
be used for Join Expression. In this process; ORCA always assumed that
scalar child of the subquery is ScalarIdent or ScalarConst and did not
take into account ScalarSubqueries. This resulted in a crash for queries
like :
```
EXPLAIN SELECT * FROM foo WHERE (SELECT COUNT(*) FROM foo) IN (SELECT 1 FROM bar);
```
The preprocessed query looks like this:
```
+--CLogicalSelect
   |--CLogicalGet "foo" ("foo"), Columns: ["a" (0), "ctid" (1), "xmin" (2), "cmin" (3), "xmax" (4), "cmax" (5), "tableoid" (6), "gp_segment_id" (7)] Key sets: {[1,7]}
   +--CScalarSubqueryAny(=)["?column?" (16)]
      |--CLogicalProject
      |  |--CLogicalGet "bar" ("bar"), Columns: ["b" (8), "ctid" (9), "xmin" (10), "cmin" (11), "xmax" (12), "cmax" (13), "tableoid" (14), "gp_segment_id" (15)] Key sets: {[1,7]}
      |  +--CScalarProjectList
      |     +--CScalarProjectElement "?column?" (16)
      |        +--CScalarConst (1)
      +--CScalarSubquery["count" (25)]
         +--CLogicalGbAgg( Global ) Grp Cols: [][Global], Minimal Grp Cols: [], Generates Duplicates :[ 0 ]
            |--CLogicalGet "foo" ("foo"), Columns: ["a" (17), "ctid" (18), "xmin" (19), "cmin" (20), "xmax" (21), "cmax" (22), "tableoid" (23), "gp_segment_id" (24)] Key sets: {[1,7]}
            +--CScalarProjectList
               +--CScalarProjectElement "count" (25)
                  +--CScalarAggFunc (count , Distinct: false , Aggregate Stage: Global)
```

For this tree, ORCA generates a join expression with incorrect predicate-
```
--CLogicalLeftSemiApplyIn (Reqd Inner Cols: “?column?” (17))
   |--CLogicalGet “foo” (“foo”), Columns: [“a” (0), “b” (1), “ctid” (2), “xmin” (3), “cmin” (4), “xmax” (5), “cmax” (6), “tableoid” (7), “gp_segment_id” (8)] Key sets: {[2,8]}   origin: [Grp:0, GrpExpr:0]
   |--CLogicalSelect
   |  |--CLogicalProject   origin: [Grp:5, GrpExpr:0]
   |  |  |--CLogicalGet “bar” (“bar”), Columns: [“c” (9), “ctid” (10), “xmin” (11), “cmin” (12), “xmax” (13), “cmax” (14), “tableoid” (15), “gp_segment_id” (16)] Key sets: {[1,7]}   origin: [Grp:1, GrpExpr:0]
   |  |  +--CScalarProjectList   origin: [Grp:4, GrpExpr:0]
   |  |     +--CScalarProjectElement “?column?” (17)   origin: [Grp:3, GrpExpr:0]
   |  |        +--CScalarConst (1)   origin: [Grp:2, GrpExpr:0]
   |  +--CScalarCmp (=)
   |     |--CScalarSubquery[“count” (27)]
   |     +--CScalarIdent “?column?” (17)
   +--CScalarConst (1)
```

This join expression does not contain the subtree that produces `CScalarSubquery["count" (27)]` and later it hits an assert `1 == exprhdl.UlArity` in CScalarSubquery.cpp.

This commit fixes this issue by recursively trying to unnest the subqueries appearing in scalar part of ANY/ALL subqueries.


libgpopt/include/gpopt/operators/CScalarSubqueryQuantified.h
```diff
@@ -117,8 +118,10 @@ namespace gpopt
 			virtual
 			CExpression *PexprSubqueryPred
 				(
-				IMemoryPool *pmp,
-				CExpressionHandle &exprhdl
+				CSubqueryHandler &sh,
```
hsyuan: we can give it better name: subqueryHandler

libgpopt/src/operators/CScalarSubqueryQuantified.cpp
```diff
@@ -217,29 +218,105 @@ CScalarSubqueryQuantified::PpartinfoDerive
...
+//		scalar predicate comparison between the new Scalar and Logical children of
+//		the Subquery. It returns the new Logical child via ppexprResult.
+//
+//		For example, for the query :
+//		select * from foo where
+//		    (select a from foo limit 1) in (select 1 from bar);
+//
+//		+--CLogicalSelect
+//		   |--CLogicalGet "foo"
+//		   +--CScalarSubqueryAny(=)["b" (10)]
```
hsyuan: where does b come from? should the subquery be `select b from bar` instead of `select 1 from bar`?

libgpopt/src/xforms/CXformSubqueryUnnest.cpp
```diff
@@ -196,6 +196,7 @@ CXformSubqueryUnnest::Transform
 	}
 
 	// second, generate the alternative where a correlated Apply always enforced
+
```
hsyuan: is it a typo?

libgpopt/include/gpopt/operators/CScalarSubqueryQuantified.h
dhanashreek89: Upon taking a look at other places in the code; I observed that the convention is to use `sh` for CSubqueryHandler; so keeping it as it is to remain consistent.

PR #249, opened 2017-10-19
Use PROJECT_SOURCE_DIR instead of CMAKE_SOURCE_DIR

d:
This is semantically more precise, and also enables ORCA to be included
in other CMake projects

[ci skip]


hardikar:
Ah thanks for doing this!

PR #251, opened 2017-10-21
Conan now no longer installs into /usr/local

pivotal-mike:
This was a bit troublesome as:

  - Conan never explicitly stated it was installing into the filesystem

  - Many users either did not have permissions normally, or did not want
    to install this directly into /usr/local

  - Conan is more of a 'ruby gems' style package manager and not RPM or
    homebrew.  In that regards the philosophy has been to stage the
    builds somewhere for another project to use and finally install

  - This posed problems for how we are working with ORCA in GPDB, in
    particular it prevented use from packaging ORCA with GPDB both in an
    open source build and for Pivotal's closed source building

  - This would only install on the compilation machine, making
    installing GPDB more complicated if we are not using an install
    provided by Pivotal

Thus we are now deferring the installation, packaging and  linking
etc... of ORCA to whatever application is currently using ORCA instead
of relying on ORCA being installed directly by Conan.

Future PR will land on GPDB to install GPDB either into the file system or into GPDB proper


bhuvnesh2703:
> This would only install on the compilation machine, making
installing GPDB more complicated if we are not using an install
provided by Pivotal

Currently, the way it is being used in ORCA pipeline is to just upload the source of orca to bintray.. the compilation was done on the host machine where gpdb pulls it up from bintray. 

With the change of not installing in /usr/local by default, we will have to add steps to copy that to /usr/local, so i think we should include those changes as well before merging it (in gpdb). This will also impact the Ubuntu team as well.  @goutamtadi1 @larham fyi

Should we keep the `self.copy ` files part as that would copy only in the .conan home directory folder for the package?

larham:
Yes, as @bhuvnesh2703 said, it seems like this change to remove copying would be a breaking change, and adding a commit to copy the files (by whatever means is approved ) would keep us green.

pivotal-mike:
Correct - I've a branch set up right now for building and installing it via
GPDB make files but it's blocked on another open PR in ORCA as it won't
compile with GCC 6 at the moment.

w.r.t. ubuntu team - again compiling ORCA is kinda broken on GCC 6 so it's
all in limbo until that's done.  Down side is I need this in place before I
can push any changes to GPDB.

dhanashreek89:
> Correct - I've a branch set up right now for building and installing it via
GPDB make files but it's blocked on another open PR in ORCA as it won't
compile with GCC 6 at the moment.

We just merged https://github.com/greenplum-db/gporca/pull/246. Thanks.

dhanashreek89:
We will need to bump patch version in CMakeLists.txt.

pivotal-mike:
@bhuvnesh2703 @larham see https://github.com/greenplum-db/gpdb/pull/3615

Part of what I wanted to do was update the Ubuntu build to support this but was blocked on gcc support

d:
LGTM

PR #258, opened 2017-11-08
Support requests to interrupt and abort optimization

hardikar:
GPOS_CHECK_ABORT calls sprinkled around in ORCA code checks if ORCA is
running and if there is a request to abort. If so, it raises an
exception (see GPOS_ABORT) of type CException::ExmiAbort which stops all
running tasks.

See corresponding changes in GPDB : https://github.com/greenplum-db/gpdb/pull/3823


d:
I'm a bit yuck about the global-variable nature of `CWorker::abort_requested`, but well.

LGTM

PR #259, opened 2017-11-10
Supporting Join Optimization Levels in GPORCA 

bhuvnesh2703:
The concept of optimization levels is known in many enterprise optimizers. It enables user to handle the degree of optimization that is being employed. The optimization levels allow the grouping of transformations into bags of rules (where each is assigned a particular level). By default all rules are applied, but if a user wants to apply fewer rules they are able to. This decision is made by them based on domain knowledge, and they know even with fewer rules being applied the plan generated satisfies their needs.

The Cascade optimizer, on which GPORCA is based on, allows grouping of transformation rules into optimization levels. This concept of optimization levels has also been extended to join ordering allowing user to pick the join order via the query or, use greedy approach or use exhaustive approach.

Postgres based planners use join_limit and from_limit to reduce the search space. While the objective of Optimization/Join is also to reduce search space, but the way it does it is different. It is requesting the optimizer to apply or not apply a subset of rules and providing more flexibility to the customer. This is one of the most frequently requested feature from our enterprise clients who have high degree of domain knowledge.

This PR introduces this concept. In the immediate future we are planning to add different polynomial join ordering techniques with guaranteed bound as part of the "Greedy" search.


dhanashreek89:
Great work!

hsyuan:
The GUC name in DDL of minidump is obselete. 

I don't see any corresponding orca traceflag or hint variable. How is the optimizer GUC reflected in minidumps? 

The 2 minidumps have the same plan even with different optimization model, it won't help test the correctness unless generating different plans given different heuristic model.

bhuvnesh2703:
> The GUC name in DDL of minidump is obselete.
Will update the GUC name.

> I don't see any corresponding orca traceflag or hint variable. How is the optimizer GUC reflected in minidumps?
the traceflags are different `102002,102003,` vs `102002`

> The 2 minidumps have the same plan even with different optimization model, it won't help test the correctness unless generating different plans given different heuristic model.
The greedy search picks up the plan with lower cost which in this case is same as that of the query model. But i see your point, i could add one more mdp which picks up the plan based on the cardinality.

hsyuan:
I see, thanks!

PR #262, opened 2017-11-17
Remove deadcode

d:
Surfaced by removal of trivial destructor for CDouble.

In Clang/LLVM and GCC, if we explicitly declare a trivial destructor,
the compilers would stop generating the warning for unused variables and
unused-but-set variables. That's how we were able to carry so much dead
code for 6 years without getting a friendly reminder at the compiler
department.

Bump ORCA version to 2.49.0


d:
I merged first, so #258 #259 and #261 will have to change their version numbers, if it's included

PR #263, opened 2017-11-18
ORCA should not hard code `row_number` function

vraghavan78:
The principle in ORCA has been to assume as little as possible about the host, 
which means that we abstract away a lot of things, and perform lookups for 
concrete types and metadata for the host (think: table oid, operator existence, 
commutativity of operators, function strictness). 

This commit fixes one that slipped through the cracks where we hardcoded a magic number 
function OID for `row_number` in ORCA.


vraghavan78:
After this the only change that needs to happen in GPDB is 

```
[gpdemo: vraghavan{magic_number}]$ git diff
diff --git a/src/backend/gpopt/utils/COptTasks.cpp b/src/backend/gpopt/utils/COptTasks.cpp
index 094135c878..3671afd991 100644
--- a/src/backend/gpopt/utils/COptTasks.cpp
+++ b/src/backend/gpopt/utils/COptTasks.cpp
@@ -763,7 +763,8 @@ COptTasks::PoconfCreate
                                                                ulBroadcastThreshold,
                                                                false /* don't create Assert nodes for constraints, we'll
                                                                      * enforce them ourselves in the executor */
-                                                               )
+                                                               ),
+                                               GPOS_NEW(pmp) CWindowOids(7000, 70001)
                                                );
 }
```

d:
@vraghavan78 
Did you check in the changes to `libnaucrates/src/parser/CParseHandlerFactory.cpp`? Intuitively I feel like we always need to touch there if we add a new tag...

d:
I had some cosmetic changes I'll push in in a second

libnaucrates/src/CDXLUtils.cpp
```diff
@@ -1218,6 +1219,10 @@ CDXLUtils::SerializeOptimizerConfig
 	xmlser.AddAttribute(CDXLTokens::PstrToken(EdxltokenBroadcastThreshold), phint->UlBroadcastThreshold());
 	xmlser.AddAttribute(CDXLTokens::PstrToken(EdxltokenEnforceConstraintsOnDML), phint->FEnforceConstraintsOnDML());
 	xmlser.CloseElement(CDXLTokens::PstrToken(EdxltokenNamespacePrefix), CDXLTokens::PstrToken(EdxltokenHint));
+
+	xmlser.OpenElement(CDXLTokens::PstrToken(EdxltokenNamespacePrefix), CDXLTokens::PstrToken(EdxltokenWindowOids));
+	xmlser.AddAttribute(CDXLTokens::PstrToken(EdxltokenOidRowNumber), pwindowoids->OidRowNumber());
+	xmlser.AddAttribute(CDXLTokens::PstrToken(EdxltokenOidRank), pwindowoids->OidRank());
 }
```
dhanashreek89: we are missing a `xmlser.CloseElement(CDXLTokens::PstrToken(EdxltokenNamespacePrefix), CDXLTokens::PstrToken(EdxltokenWindowOids));` here.
This is causing incorrect <Windowoid> tag in the minidump:
```
<dxl:WindowOids RowNumber="7000" Rank="70001"
```

vraghavan78:
Fixed.
ᐧ

vraghavan78:
Thanks for catching this. Fixed.
ᐧ

dhanashreek89:
@vraghavan78 
```
[gpdemo: vraghavan{magic_number}]$ git diff
diff --git a/src/backend/gpopt/utils/COptTasks.cpp b/src/backend/gpopt/utils/COptTasks.cpp
index 094135c878..3671afd991 100644
--- a/src/backend/gpopt/utils/COptTasks.cpp
+++ b/src/backend/gpopt/utils/COptTasks.cpp
@@ -763,7 +763,8 @@ COptTasks::PoconfCreate
                                                                ulBroadcastThreshold,
                                                                false /* don't create Assert nodes for constraints, we'll
                                                                      * enforce them ourselves in the executor */
-                                                               )
+                                                               ),
+                                               GPOS_NEW(pmp) CWindowOids(7000, 70001)
                                                );
 }
```

After the Window changes get merged in gpdb master (https://github.com/greenplum-db/gpdb/pull/3426), the new oids for `row_number` and `rank` are `3100` and `3101` respectively. 

Hence the new patch for gpdb should be:
```
--- a/src/backend/gpopt/utils/COptTasks.cpp
+++ b/src/backend/gpopt/utils/COptTasks.cpp
@@ -763,7 +763,8 @@ COptTasks::PoconfCreate
                                                                ulBroadcastThreshold,
                                                                false /* don't create Assert nodes for constraints, we'll
                                                                      * enforce them ourselves in the executor */
-                                                               )
+                                                               ),
+                                                               GPOS_NEW(pmp) CWindowOids(3100, 3101)
                                                );
 }
```

We tested this PR along with the above patch; it works successfully and also fixes one ORCA test failure. Thanks

libgpopt/include/gpopt/base/CWindowOids.h
```diff
@@ -0,0 +1,61 @@
...
+//---------------------------------------------------------------------------
+#ifndef GPOPT_CWindowOids_H
+#define GPOPT_CWindowOids_H
+
+#include "gpos/common/CRefCount.h"
+#include "naucrates/dxl/gpdb_types.h"
+
+#define DUMMY_ROW_NUMBER_OID OID(7000)
+#define DUMMY_WIN_RANK OID(7001)
+
```
dhanashreek89: These oids will be obsolete after the window merge. What is the intention of these Dummy oids ?

vraghavan78: This is the default because I did not want to change all existing files. Will change it.

vraghavan78: After some thought, we need this since not all minidumps have OptimizerConfig. They all need to change.

sambitesh:
Instead of this `<dxl:CTEConfig CTEInliningCutoff="0"/> <dxl:WindowOids RowNumber="7000" Rank="7001"/>`

We would like `<dxl:WindowOids RowNumber="7000" Rank="7001"/>` to be in a separate line.

vraghavan78:
oh.... come on. I am not gonna change 900+ files.

vraghavan78:
@sambitesh @oarap all tests file have been modified

vraghavan78:
In GPDB 4 & 5 the oid is 7000 and 7001 while in GPDB 6 it is 3100 and 3101

bhuvnesh2703:
@d @oarap @hsyuan looks like you reviewed this pr earlier, could you please take a look and approve it unless there is more feedback. i have tested it with the corresponding gpdb pr https://github.com/greenplum-db/gpdb/pull/4174.
I will squash the commits before pushing on this pr.

PR #264, opened 2017-11-21
Hash Join alternative not being generated for queries involving 3 or more tables (partitioned)

bhuvnesh2703:
Push Partition Request to join child assuming no DPE
Consider a query of the form as below joining 1 partitioned table (p1) with other non partitioned table (foo):
```sql
     set optimizer_join_order=query;
    create table p1(a int, b int) partition by range(b) (start (1) end (2) every (1));
    create table bar (a int, b int);
    select disable_xform('CXformInnerJoin2NLJoin');
    explain select * from p1, foo where p1.b = foo.b;
```

The Memo after optimization for the above query is as below:
```
    Group 1 (#GExprs: 6):
      0: CLogicalGet "foo" ("foo")
      1: CPhysicalTableScan "foo"
      2: CPhysicalMotionGather(master) [ 1 ]
      3: CPhysicalPartitionSelector, Scan Id: 1, Part Table: (24576,1.0) [ 1 ]
      4: CPhysicalMotionBroadcast  [ 1 ]
      5: CPhysicalMotionHashDistribute HASHED: [ +--CScalarIdent "b" (10)
     , nulls colocated ] [ 1 ]
    ",
    2017-11-20 16:03:14:376180 PST,THD000,TRACE,"
    Group 0 (#GExprs: 4):
      0: CLogicalDynamicGet "p1" ("p1"),
      1: CPhysicalDynamicTableScan "p1"
      2: CPhysicalMotionGather(master) [ 0 ]
      3: CPhysicalMotionHashDistribute HASHED: [ +--CScalarIdent "b" (1)
     , nulls colocated ] [ 0 ]
```

Current DPE framework always pushes the PS to the inner child (i.e
table foo Group 1), and while trying to identify the required partition
propagation spec for the outer child (i.e table p1, Group 0) if the PS
is already pushed to the inner child it does not insert the partition
selector in outer child (i.e Group 0). This works fine if there is a join
between 2 tables.
However, as we increase the number of tables joined we may result into
scenario where we may not be always able to generate a plan with DPE,
and we should pick a plan without DPE. In such case, we should have an
alternative which pushed PS to the join child without considering DPE.
For example:
```sql
    explain select * from bar, p1, foo where bar.b = p1.b and p1.b = foo.b;
```

Memo after optimization:
```
    Group 2 (#GExprs: 7):
      0: CLogicalGet "foo" ("foo")
      1: CPhysicalTableScan "foo" ("foo") [ ]
      2: CPhysicalMotionGather(master) [ 2 ]
      3: CPhysicalPartitionSelector, Scan Id: 1, Part Table: (24576,1.0) [ 2 ]
      4: CPhysicalMotionBroadcast  [ 2 ]
      5: CPhysicalMotionHashDistribute HASHED: [ +--CScalarIdent "b" (19)
     +--CScalarIdent "b" (19)   origin: [Grp:6, GrpExpr:0]
     , nulls colocated ] [ 2 ]
      6: CPhysicalMotionHashDistribute HASHED: [ +--CScalarIdent "b" (19)
     , nulls colocated ] [ 2 ]
    ",
    2017-11-20 16:10:11:342332 PST,THD000,TRACE,"
    Group 1 (#GExprs: 5):
      0: CLogicalDynamicGet "p1" ("p1")
      1: CPhysicalDynamicTableScan "p1" ("p1")
      2: CPhysicalMotionGather(master) [ 1 ]
      3: CPhysicalMotionBroadcast  [ 1 ]
      4: CPhysicalMotionHashDistribute HASHED: [ +--CScalarIdent "b" (10)
     , nulls colocated ] [ 1 ]
    ",
    2017-11-20 16:10:11:342381 PST,THD000,TRACE,"
    Group 0 (#GExprs: 2):
      0: CLogicalGet "bar" ("bar"), Columns: ["a" (0), "b" (1), "ctid" (2), "xmin" (3), "cmin" (4), "xmax" (5), "cmax" (6), "tableoid" (7), "gp_segment_id" (8)] Key sets: {[2,8]} [ ]
      1: CPhysicalTableScan "bar" ("bar") [ ]
```
This plan deduced from the above memo will not work.
```
    HJ (1)
    +----HJ (2)
    |     +------Scan on bar (Probe Side, outer side)
    |     +------Motion
                 +-------DTS on p1 (Hash Side, inner side)
    +----Motion
          +------Scan on foo
```
In the above memo & plan (manually made) after optimization, there is no
PS inserted into Group 1 corresponding to the DTS for p1 as it has
already inserted a partition selector in Group 2 expecting that it will
be able to do DPE. However, for HJ the table are joined on column due to
which a motion is required due to which we cannot push the PS on the DTS
using the filter populated by scanning foo. Thus it is not able to find
a plan matching the required properties.
The resulting plan should be of the form
```

    HJ (1)
    +----HJ (2)
    |     +------Motion
    |            +-------Scan on bar (Probe Side, outer side)
    |     +------Motion
    |             +-------Sequence
    |                     +-------PS
    |                     +-------DTS on p1 (Hash Side, inner side)
    +----Motion
         +------Scan on foo
```
In order to insert the grouping expression PS considering No DPE, this
commit adds the request to generate such alternative.

With DynamicProgramming Xform on, we generate several alternative so
this case is not observed as one of the alternative might have inserted
the PS or the plan chosen would have a different join order, however
when we restrict only to query optimizer_join_order we need to consider
this case as there is no other alternative.

With now a new alternative plan, the current cost framework marks the plan with no DPE as slightly cheaper than the plan with DPE and picks it up as the best plan in case of `optimizer_join_order=query`. To avoid this for `optimizer_join_order=exhaustive`, the optimization request to push PS to join child without DPE is only considered with Dynamic Programming is off.  We should tune the cost to penalize non dpe plans so that DPE is picked up. Will create a followup story for that.

### DIFF
--- a/libgpopt/src/operators/CPhysicalInnerHashJoin.cpp
+++ b/libgpopt/src/operators/CPhysicalInnerHashJoin.cpp
@@ -275,9 +275,22 @@ CPhysicalInnerHashJoin::PppsRequired
 	CPartitionPropagationSpec *pppsRequired,
 	ULONG ulChildIndex,
 	DrgPdp *pdrgpdpCtxt,
-	ULONG // ulOptReq
+	ULONG ulOptReq
 	)
 {
+
+	if (1 == ulOptReq)
+	{
+		// request (1): push partition propagation requests to join's children,
+		// do not consider possible dynamic partition elimination using join predicate here,
+		// this is handled by optimization request (0) below
+		return CPhysical::PppsRequiredPushThruNAry(pmp, exprhdl, pppsRequired, ulChildIndex);
+	}
+
+	// request (0): push partition progagation requests to join child considering
+	// DPE possibility. For HJ, PS request is pushed to the inner child if there
+	// is a consumer (DTS) on the outer side of the join.
+	GPOS_ASSERT(0 == ulOptReq);
 	return PppsRequiredJoinChild(pmp, exprhdl, pppsRequired, ulChildIndex, pdrgpdpCtxt, false);
 }
 
--- a/data/dxl/minidump/JoinOptimizationLevelQuery3WayHashJoinPartTbl.mdp
+++ b/data/dxl/minidump/JoinOptimizationLevelQuery3WayHashJoinPartTbl.mdp
@@ -1,0 +1,1380 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+    create table foo (a int, b int) distributed by (a);
+    create table bar (a int, b int) distributed by (a);
+    create table p1(a int, b int) partition by range(b) (start (1) end (5) every (1));
+    insert into foo select i, 1 from generate_series(1,10000)i;
+    insert into p1 select i, 1 from generate_series(1,10000)i;
+    insert into p1 select i, 2 from generate_series(1,10000)i;
+    analyze foo;
+    analyze p1;
+    explain select * from foo, p1, bar where foo.b = p1.b and p1.b = bar.b;
+                                                        QUERY PLAN
+     Gather Motion 3:1  (slice3; segments: 3)  (cost=0.00..12183.97 rows=1 width=24)
+       ->  Hash Join  (cost=0.00..12183.97 rows=1 width=24)
+             Hash Cond: foo.b = bar.b AND p1.b = bar.b
+             ->  Hash Join  (cost=0.00..2738.17 rows=33333334 width=16)
+                   Hash Cond: foo.b = p1.b
+                   ->  Table Scan on foo  (cost=0.00..431.07 rows=3334 width=8)
+                   ->  Hash  (cost=434.16..434.16 rows=20000 width=8)
+                         ->  Broadcast Motion 3:3  (slice1; segments: 3)  (cost=0.00..434.16 rows=20000 width=8)
+                               ->  Sequence  (cost=0.00..431.14 rows=6667 width=8)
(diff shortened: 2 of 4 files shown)


libgpopt/src/operators/CPhysicalInnerHashJoin.cpp
```diff
@@ -275,9 +275,19 @@ CPhysicalInnerHashJoin::PppsRequired
...
-	ULONG // ulOptReq
+	ULONG ulOptReq
 	)
 {
+
+	if (1 == ulOptReq)
+	{
+		// request (1): push partition propagation requests to join's children,
+		// do not consider possible dynamic partition elimination using join predicate here,
+		// this is handled by optimization request (0) below
```
vraghavan78: can you expand on request (0) also?

vraghavan78:
no tests? What if you had the 5 way join and a join optimization level?

bhuvnesh2703:
> no tests? 
added now

> What if you had the 5 way join and a join optimization level?
a plan with HJ will be generated now. (optimizer_join_order=query, join between p1..p6 part tbls).

libgpopt/src/operators/CPhysicalInnerHashJoin.cpp
bhuvnesh2703: added

PR #272, opened 2017-11-29
DXL changes to window spec due to Postgres 8.4 merge

vraghavan78:
In GPDB 5 and before, window specification cannot have a window frame specification
without sort columns. This changed in GPDB 6/Postgres 8.4+ where a query
select b,c, count(c) over (partition by b) from (select * from foo) s;
adds a window frame that is unbounded.


vraghavan78:
Thanks, merged.

PR #273, opened 2017-11-30
Remove hardcoded GPDB specific OID that changed between GPDB 5 v 6

vraghavan78:
This logic should be inside GPDB

### DIFF
--- a/libgpopt/src/translate/CTranslatorDXLToExpr.cpp
+++ b/libgpopt/src/translate/CTranslatorDXLToExpr.cpp
@@ -43,12 +43,6 @@
 
 #include "naucrates/traceflags/traceflags.h"
 
-#define GPDB_DENSE_RANK_OID 7002
-#define GPDB_PERCENT_RANK_OID 7003
-#define GPDB_CUME_DIST_OID 7004
-#define GPDB_NTILE_INT4_OID 7005
-#define GPDB_NTILE_INT8_OID 7006
-#define GPDB_NTILE_NUMERIC_OID 7007
 #define GPOPT_ACTION_INSERT 0
 #define GPOPT_ACTION_DELETE 1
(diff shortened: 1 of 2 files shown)


libgpopt/src/translate/CTranslatorDXLToExpr.cpp
```diff
@@ -2986,10 +2951,6 @@ CTranslatorDXLToExpr::PexprWindowFunc
 	pmdidFunc->AddRef();
 
 	CWStringConst *pstrName = GPOS_NEW(m_pmp) CWStringConst(m_pmp, CMDAccessorUtils::PstrWindowFuncName(m_pmda, pmdidFunc)->Wsz());
-	if (FUnsupportedWindowFunc(pmdidFunc))
```
khannaekta: Currently `FUnsupportedWindowFunc()` uses the hard-coded OIds to check if the Window function is supported by ORCA or not. If not supported, it raises an exception.
Shouldn't we still have this check and exception be raised on ORCA side rather than checking for unsupported window functions on GPDB/host side? 
Would it make sense to update `FUnsupportedWindowFunc()` to check the `PstrWindowFuncName` instead of checking with the hardcoded OIDs?

vraghavan78: There is a GPDB side that will throw an exception. I am having issue with a flakey test. I do not like name it's buggy.

vraghavan78: Raising exception inside Orca is too late. Why would you waste time algebrizing when you do not support it.

bhuvnesh2703:
Closing this PR, Refer https://github.com/greenplum-db/gporca/pull/287

Issue #268, opened 2017-11-25
Orca GUC optimizer_join_order_threshold doesn't take effect


hsyuan:
The GPDB GUC `optimizer_join_order_threshold` (Orca CHint param m_ulJoinOrderDPLimit, btw, which is really bad, inconsistent naming) is only valid when its value is less than `GPOPT_DP_JOIN_ORDERING_SIZE_THRESHOLD` (10). The orca config param is overrode by https://github.com/greenplum-db/gporca/blame/master/libgpopt/src/xforms/CJoinOrderDP.cpp#L1059

bhuvnesh2703:
we should set the max value of optimizer_join_order_threshold to 10/12 after removing this macro. It will be then completely controlled by the GUC which is tunable then a hardcoded macro..

bhuvnesh2703:
PR https://github.com/greenplum-db/gpdb/pull/4046, https://github.com/greenplum-db/gporca/pull/276

hsyuan:
fixed by commit: https://github.com/greenplum-db/gporca/commit/0171e2be91648380b03d598c7033a513df8ad06b

PR #274, opened 2017-12-01
Remove GPOPT_DP_JOIN_ORDERING_SIZE_THRESHOLD

hsyuan:
We already have GPDB GUC optimizer_join_order_threshold to control the maximum
number of relations that we can do join reordering using DP. The macro being removed
here is redundant.

Fixes issue #268

### DIFF
--- a/libgpopt/src/xforms/CJoinOrderDP.cpp
+++ b/libgpopt/src/xforms/CJoinOrderDP.cpp
@@ -29,7 +29,6 @@
 
 using namespace gpopt;
 
-#define GPOPT_DP_JOIN_ORDERING_SIZE_THRESHOLD	10
 #define GPOPT_DP_JOIN_ORDERING_CONNECTEDNESS_THRESHOLD	0.5
 #define GPOPT_DP_JOIN_ORDERING_TOPK	10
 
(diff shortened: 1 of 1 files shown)


libgpopt/src/xforms/CJoinOrderDP.cpp
```diff
@@ -1056,8 +1055,7 @@ CJoinOrderDP::PexprExpand()
 		(void) pbs->FExchangeSet(ul);
 	}
 
-	if (GPOPT_DP_JOIN_ORDERING_SIZE_THRESHOLD < m_ulComps &&
-		GPOPT_DP_JOIN_ORDERING_CONNECTEDNESS_THRESHOLD < DMaxConnectedness(pbs))
+	if (GPOPT_DP_JOIN_ORDERING_CONNECTEDNESS_THRESHOLD < DMaxConnectedness(pbs))
```
vraghavan78: I think we should instead the min of (GPOPT_DP_JOIN_ORDERING_SIZE_THRESHOLD, 
GPOPT_DP_JOIN_ORDERING_CONNECTEDNESS_THRESHOLD)

hsyuan: I didn't get it.

vraghavan78: We know for a fact that when m_ulComps when > 10 this algorithm should not be applied. Right?

dhanashreek89:
You are right. There is redundancy.  The applicability of the N-ary join expansion transform is decided by `optimizer_join_order_threshold` however the actual transform is controlled by `GPOPT_DP_JOIN_ORDERING_SIZE_THRESHOLD`

With some git digging, `GPOPT_DP_JOIN_ORDERING_SIZE_THRESHOLD` was added by https://github.com/greenplum-db/QP-ORCA/commit/8d22782429775be36bfe8258faba9c46c6c6f0dd and the hint `optimizer_join_order_threshold`, was added later via https://github.com/greenplum-db/gporca/commit/f6296f70e77c2bcfdb2a21df7efb50e838ba6880

I am completely +1 in removing this; the only question is, since `optimizer_join_order_threshold` is configurable (with default value of 10); will this have considerable/severe performance impact when set to something >10 ? If yes, should we document this explicitly ?

bhuvnesh2703:
@hsyuan i will close this PR as i didn't notice this and created the other with correctness also removed?

hsyuan:
new PR: https://github.com/greenplum-db/gporca/pull/276

PR #276, opened 2017-12-05
Remove redundant macro

bhuvnesh2703:
Macro GPOPT_DP_JOIN_ORDERING_SIZE_THRESHOLD is set to 10.
However, DP Xform is not scheduled if
relations > optimizer_join_order_threshold.
This also makes the condition to check connectivity irrelevant,
as in such cases its not scheduled.


bhuvnesh2703:
In this commit we have removed the check for connectedness as it was irrelevant because if the number of relation was more than 10 DP will not be applied as per `https://github.com/greenplum-db/gporca/blob/master/libgpopt/src/xforms/CXformUtils.cpp#L150-L153`

hsyuan:
I don't see the necessity to remove connectedness check. for query `(R0 inner join R1) cross join (R2 inner join R3) cross join (R4 inner join R5)`, there may be no need to do this transform.

CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ project(gpopt LANGUAGES CXX C)
 
 set(CMAKE_CXX_STANDARD 98)
 set(GPORCA_VERSION_MAJOR 2)
-set(GPORCA_VERSION_MINOR 51)
+set(GPORCA_VERSION_MINOR 52)
```
hsyuan: I don't think there is a need to bump minor version. there is no external API change.

hsyuan:
rebase on latest master. the version got changed

PR #277, opened 2017-12-05
WIP: Introduce a new graph-based join enumeration algorithm MinCutBranch

hsyuan:
This PR is more of a RFC, not intended to merge into master yet.

This recursive implementation is based on branch partitioning algorithm
described in paper "A New, Highly Efficient, and Easy To Implement Top-Down
Join Enumeration Algorithm". Maybe we can improve it by implementing it in
iterative way to prevent stack overflow for super large join graph.

The new algorithm gives us 8x optimizer performance improvement for chain
query with 10 relations (I haven't tested on other types of query):
```sql
explain 
select * from r0, r1, r2, r3, r4, r5, r6, r7, r8, r9 
where r3.c = r1.c and r1.a = r0.a 
and r0.b = r2.b and r2.c = r4.c 
and r4.a = r5.a and r5.b = r6.b 
and r6.c = r7.c and r7.a = r8.a 
and r8.b = r9.b;
```
Without this patch, Orca takes 1600ms to generate plan; with this patch, it
only takes 230ms to generate plan, and the generated plan are same.

### DIFF
--- a/libgpos/include/gpos/common/CBitSet.h
+++ b/libgpos/include/gpos/common/CBitSet.h
@@ -104,9 +104,6 @@ namespace gpos
 			// find link with offset less or equal to given value
 			CBitSetLink *PbslLocate(ULONG, CBitSetLink * = NULL) const;
 			
-			// reset set
-			void Clear();
-			
 			// compute target offset
 			ULONG UlOffset(ULONG) const;
 			
@@ -131,6 +128,9 @@ namespace gpos
 			// clear given bit; return previous value
 			BOOL FExchangeClear(ULONG ulBit);
 			
+			// reset set
+			void Clear();
+
 			// union sets
 			void Union(const CBitSet *);
 			
--- a/libgpopt/include/gpopt/xforms/CJoinOrder.h
+++ b/libgpopt/include/gpopt/xforms/CJoinOrder.h
@@ -109,6 +109,9 @@ namespace gpopt
 			// number of components
 			ULONG m_ulComps;
 			
+			// neighbors array
+			CBitSet **m_rgpbsAdj;
+
 			// compute cover of each edge
 			void ComputeEdgeCover();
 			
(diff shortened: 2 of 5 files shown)


libgpos/include/gpos/common/CBitSet.h
```diff
@@ -131,6 +128,9 @@ namespace gpos
 			// clear given bit; return previous value
 			BOOL FExchangeClear(ULONG ulBit);
 			
+			// reset set
+			void Clear();
+
```
vraghavan78: hmm... may be not?

meister87:
Hi,
I am just curious regarding the result/ speed up.
What approach is used currently? 
The traditional dynamic programming approach DP(SIZE) ?

libgpos/include/gpos/common/CBitSet.h
hsyuan: why?

hsyuan:
@meister87 Currently Orca is using top-down memorization with all possible subsets, even cross product. This new algorithm suppresses the generation of cross product as well as symmetric connected component pairs.

meister87:
Just two hints:

1. In the paper  "Analysis of two existing and one new dynamic programming algorithm for the generation of optimal bushy join trees without cross products", the effects of such change is discussed.
Not for the top-down enumeration, but for the dynamic programming approach.
In short, as long as you do not consider cross-joins/clique queries, you will be faster without altering the result.
If you consider cross-joins/clique queries, you either will be slower or might end up with different results.

2. In the paper "Optimal top-down join enumeration" further options to increase the efficiency are discussed. At least, the Accumulated-Cost Bounding is also quite easy to consider.

vraghavan78:
Thanks @meister87 for the pointers.

@hsyuan

* Analysis of two existing and one new dynamic programming algorithm for the generation of optimal bushy join trees without cross products
http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.134.3827&rep=rep1&type=pdf

* Optimal top-down join enumeration:

http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.91.1588&rep=rep1&type=pdf

hsyuan:
Yes, I got them, thanks!

vraghavan78:
@hsyuan these two papers have been cited in the paper from pit. Just look at the related work.

vraghavan78:
We should run the benchmark and see what is the effect

hsyuan:
Close it for now, reopen when you have chance to work on it.

PR #278, opened 2017-12-06
Add new transform to remove redundant DISTINCT from subquery

hsyuan:
New transformation rule `CXformRemoveSubqDistinct` is added to remove redundant DISTINCT from subquery. In an IN, EXISTS, NOT IN or NOT EXISTS subquery, any duplicates in the subselect will not affect the overall result, so we can add a new plan alternative to throw away any DISTINCT clause. Unless there's a LIMIT.

For example, for the following query:
```sql
create table foo(a int, b int) distributed by (a);
create table bar(c int, d int) distributed by (c);
select * from foo where a in (select distinct d from bar);
```
If `bar.d` has few number of distinct values, it may give us benefit by first aggregate distinct values so that we have less tuples to move between different segments and may require less memory to build the hash table. But if `bar.d` has large number of distinct values, or UNIQUE in extreme case, it won't help a lot to aggregate distinct values, or even make it worse. So we add this new transformation rule to generate different plan alternatives and let the optimizer choose the better plan based on the cost.

Signed-off-by: Abhijit Subramanya <asubramanya@pivotal.io>


libgpopt/src/xforms/CXformRemoveSubqDistinct.cpp
```diff
@@ -0,0 +1,114 @@
...
+	)
+	const
+{
+	GPOS_ASSERT(NULL != pxfctxt);
+	GPOS_ASSERT(NULL != pxfres);
+	GPOS_ASSERT(FPromising(pxfctxt->Pmp(), this, pexpr));
+	GPOS_ASSERT(FCheckPattern(pexpr));
+
+	IMemoryPool *pmp = pxfctxt->Pmp();
+	CExpression *pexprScalar = (*pexpr)[1]; // scalar subquery expression
```
vraghavan78: comment useless

libgpopt/src/xforms/CXformRemoveSubqDistinct.cpp
```diff
@@ -0,0 +1,114 @@
...
+	CExpression *pexprGbAgg = (*pexprScalar)[0];
+
+	if (COperator::EopLogicalGbAgg == pexprGbAgg->Pop()->Eopid())
+	{
+		CExpression *pexprGbAggProjectList = (*pexprGbAgg)[1];
+		if (0 == pexprGbAggProjectList->UlArity())
```
vraghavan78: may be add comment here explaining the logic

server/src/unittest/gpopt/minidump/CExistsSubqueryTest.cpp
```diff
@@ -34,6 +34,7 @@ const CHAR *rgszExistsFileNames[] =
 		"../data/dxl/minidump/NotExists-SuperfluousEquality.mdp",
 		"../data/dxl/minidump/SimplifyExistsSubquery2Limit.mdp",
 		"../data/dxl/minidump/NotExists-SuperflousOuterRefWithGbAgg.mdp",
+		"../data/dxl/minidump/Remove-Distinct-From-Subquery.mdp",
```
vraghavan78: do we have a test that does not pick this option? If not can we add that test?

hsyuan: yes, we have. in the newly added test

libgpopt/src/xforms/CXformRemoveSubqDistinct.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+			COperator *pop = pexprScalar->Pop();
+			pop->AddRef();
+			if (CUtils::FExistentialSubquery(pop))
+			{
+				// EXIST/NOT EXIST scalar subquery
+				pexprNewScalar = GPOS_NEW(pmp) CExpression(pmp, pop, pexprRelChild);
+			}
+			else
+			{
```
vraghavan78: especially here. took me a minute to get it.

libgpopt/src/xforms/CXformRemoveSubqDistinct.cpp
```diff
@@ -0,0 +1,147 @@
...
+// For quantified subqueries (IN / NOT IN), the following transformation will be applied:
+// CLogicalSelect
+// |--CLogicalGet "foo"
+// +--CScalarSubqueryAny(=)["c" (9)]
+//    |--CLogicalGbAgg( Global ) Grp Cols: ["c" (9)]
+//    |  |--CLogicalGet "bar"
+//    |  +--CScalarProjectList
+//    +--CScalarIdent "a" (0)
+//
+// will become
```
oarap: I would say produce or generate instead of "become". The old expression is still extractable in the final plan. Otherwise this would be in preprocessing.

hsyuan: make sense, will update

PR #282, opened 2017-12-14
 Reimplement ORCA interrupts using a callback function

hardikar:
As pointed out by Heikki, maintaining another variable to match one in
the database system will be error-prone and cubersome. This commit
initializes ORCA with a callback function into the host system that can
return if a cancel is requested by the system.


CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ project(gpopt LANGUAGES CXX C)
 set(CMAKE_CXX_STANDARD 98)
 set(GPORCA_VERSION_MAJOR 2)
 set(GPORCA_VERSION_MINOR 51)
-set(GPORCA_VERSION_PATCH 3)
+set(GPORCA_VERSION_PATCH 0)
```
hsyuan: version number is not correct.

PR #283, opened 2017-12-15
Fix cardinality for Filter condition with ScalarCast.

sambitesh:
When we have a ScalarCast on top of a ScalarIdent as shown below, the
cardinality of the Filter condition was incorrectly computed.

```
DrvdScalarProps
   |--CScalarCast
   |  DrvdScalarProps
   |  +--CScalarIdent
   |     DrvdScalarProps
   +--CScalarArray
      DrvdScalarProps
```

The ProcessArrayCmp() always assumed that a filter condition always had a
ScalarIdent as a child. But as in the case shown above, there can be a
ScalarCast on top of it. In such cases, ORCA assumed that the column didn't
exist in the histogram and hence ended up computing the cardinality
incorrectly. This patch fixes the issue by checking for the ScalarCast node and
extracting the ScalarIdent correctly.


libnaucrates/src/statistics/CStatsPredUtils.cpp
```diff
@@ -906,12 +906,19 @@ CStatsPredUtils::ProcessArrayCmp
...
+	CExpression *pexprIdent = NULL;
 	CScalarArrayCmp *popScArrayCmp = CScalarArrayCmp::PopConvert(pexprPred->Pop());
-	CExpression *pexprIdent = (*pexprPred)[0];
+	if (CUtils::FScalarIdent((*pexprPred)[0]))
+	{
+		pexprIdent = (*pexprPred)[0];
+	}
+	else if (CUtils::FScalarCast((*pexprPred)[0]))
+	{
+		pexprIdent = (*(*pexprPred)[0])[0];
```
oarap: How do you make sure that it is an Ident? Can it be a scalar subquery?

To be more specific, do you handle this case?
```
create table bar (a varchar(10));
select a from bar where (select a from bar limit 1) in ('some', 'string');
```

```
+--CLogicalSelect
   |--CLogicalGet "bar" ("bar"), Columns: ["a" (0), "ctid" (1), "xmin" (2), "cmin" (3), "xmax" (4), "cmax" (5), "tableoid" (6), "gp_segment_id" (7)] Key sets: {[1,7]}
   +--CScalarArrayCmp Any (=)
      |--CScalarCast
      |  +--CScalarSubquery["a" (8)]
      |     +--CLogicalLimit <empty> global
      |        |--CLogicalGet "bar" ("bar"), Columns: ["a" (8), "ctid" (9), "xmin" (10), "cmin" (11), "xmax" (12), "cmax" (13), "tableoid" (14), "gp_segment_id" (15)] Key sets: {[1,7]}
      |        |--CScalarConst (0)
      |        +--CScalarConst (1)
      +--CScalarArray: {eleMDId: (25,1.0), arrayMDId: (1009,1.0) CScalarConst (51684204.000) CScalarConst (674030452.000)}
```

libnaucrates/src/statistics/CStatsPredUtils.cpp
```diff
@@ -906,12 +906,19 @@ CStatsPredUtils::ProcessArrayCmp
...
 	GPOS_ASSERT(2 == pexprPred->UlArity());
-
+	CExpression *pexprIdent = NULL;
 	CScalarArrayCmp *popScArrayCmp = CScalarArrayCmp::PopConvert(pexprPred->Pop());
-	CExpression *pexprIdent = (*pexprPred)[0];
+	if (CUtils::FScalarIdent((*pexprPred)[0]))
+	{
+		pexprIdent = (*pexprPred)[0];
+	}
+	else if (CUtils::FScalarCast((*pexprPred)[0]))
```
hsyuan: should use `FBinaryCoercibleCast`

libgpopt/include/gpopt/operators/CPredicateUtils.h
```diff
@@ -204,7 +204,7 @@ namespace gpopt
 			// either the constant or the column can be casted
 			static
 			BOOL FIdentINDFConstIgnoreCast(CExpression *pexpr);
-
+
```
oarap: extra space

libgpopt/src/base/CUtils.cpp
```diff
@@ -3228,6 +3228,17 @@ CUtils::FScalarArray
 	return (COperator::EopScalarArray == pexpr->Pop()->Eopid());
 }
 
+// check if given expression is a scalar array
```
d: A wrong comment is worse than no comment

libgpopt/src/operators/CPredicateUtils.cpp
```diff
@@ -1227,6 +1227,29 @@ CPredicateUtils::FIdentINDFConstIgnoreCast
...
+CPredicateUtils::FCompareCastIdentToConstArray
+	(
+	CExpression *pexpr
+	)
+{
+	GPOS_ASSERT(NULL != pexpr);
+	
+	if (CUtils::FScalarArrayCmp(pexpr) &&
+		(CUtils::FScalarCast((*pexpr)[0]) &&
+		 CUtils::FScalarIdent((*(*pexpr)[0])[0])))
```
d: 1. I'd suggest you consider using `CScalarIdent::FCastedScId` here
1. Why are we sure that `(*expr)[1]` will be a scalar array?

d:
-1. See inline comments

vraghavan78:
Agree with @d. We need to check if the array contains all constants.

oarap:
@vraghavan78, @d Made the changes. Please take another look.

vraghavan78:
LGTM

d:
ship it

sambitesh:
Addressed all review comments and pushed. Thanks for review.

PR #285, opened 2017-12-19
Avoid cross join if the query has joins on relations

bhuvnesh2703:
cross join contains CScalarConst(1) as the join condition.  if the
input expression is as below with cross join at top level between
CLogicalInnerJoin and CLogicalGet "t3"
```
+--CLogicalInnerJoin
   |--CLogicalInnerJoin
   |  |--CLogicalGet "t1"
   |  |--CLogicalGet "t2"
   |  +--CScalarCmp (=)
   |     |--CScalarIdent "a" (0)
   |     +--CScalarIdent "b" (9)
   |--CLogicalGet "t3"
   +--CScalarConst (1)
```
for the above expression (lower) predicate generated for the cross join
between t1 and t3 will be: CScalarConst (1) In *only* such cases, donot
generate such alternative with the lower join as cross join example:
```
+--CLogicalInnerJoin
   |--CLogicalInnerJoin
   |  |--CLogicalGet "t1"
   |  |--CLogicalGet "t3"
   |  +--CScalarConst (1)
   |--CLogicalGet "t2"
   +--CScalarCmp (=)
      |--CScalarIdent "a" (0)
      +--CScalarIdent "b" (9)
```
Signed-off-by: Shreedhar Hardikar <shardikar@pivotal.io>


libgpopt/src/xforms/CXformJoinAssociativity.cpp
```diff
@@ -200,8 +235,48 @@ CXformJoinAssociativity::Transform
...
+	//     |  |--CLogicalGet "t1"
+	//     |  |--CLogicalGet "t3"
+	//     |  +--CScalarConst (1)
+	//     |--CLogicalGet "t2"
+	//     +--CScalarCmp (=)
+	//        |--CScalarIdent "a" (0)
+	//        +--CScalarIdent "b" (9)
+
+	// NOTE that we want to be careful to check that input lower join wasn't a
+	// crass join to begin with, because we want to build a join in this case even
```
sambitesh: typo : "cross join" instead of "crass join"

vraghavan78:
Ok cool

hsyuan:
Can we further improve `CJoinOrderDP::PexprPred` called in `CJoinOrderDP::PexprBestJoinOrderDP`?  it will still generate cross join there even 2 relation don't have predicate.

bhuvnesh2703:
> Can we further improve CJoinOrderDP::PexprPred called in CJoinOrderDP::PexprBestJoinOrderDP? it will still generate cross join there even 2 relation don't have predicate.

May be Yes, we used other queries where such alternatives are generated by the DP xform, but we limited this story for Associativity and not convolute it. We planned to create a separate story for that.

PR #287, opened 2017-12-27
Remove check for unsupported window func

bhuvnesh2703:
GPDB6 supports the queries which were banned by the FUnsupportedWindowFunc method. So will remove it. 
GPDB4 does not work without it. So currently will move this check to gpdb4 translator.


CMakeLists.txt
```diff
@@ -5,8 +5,8 @@ project(gpopt LANGUAGES CXX C)
 
 set(CMAKE_CXX_STANDARD 98)
 set(GPORCA_VERSION_MAJOR 2)
-set(GPORCA_VERSION_MINOR 53)
-set(GPORCA_VERSION_PATCH 1)
+set(GPORCA_VERSION_MINOR 54)
```
hsyuan: you don't need to bump minor version

PR #289, opened 2017-12-28
Remove redundant motion for hash distributed join relation

hsyuan:
For the following query, Orca generated redundant redistribution motion on join
output that is already distributed by join key:
```sql
create table foo(a int, b int, primary key(a));
create table bar(a int, b int, primary key(a));
create table zoo(a int, b int, primary key(a));
explain select * from foo left join bar on foo.a=bar.a left join zoo on zoo.a=bar.a;
                                           QUERY PLAN
 Gather Motion 3:1  (slice2; segments: 3)  (cost=0.00..439.00 rows=3 width=24)
   ->  Nested Loop Left Join  (cost=0.00..439.00 rows=1 width=24)
         Join Filter: true
         ->  Redistribute Motion 3:3  (slice1; segments: 3)  (cost=0.00..437.00 rows=1 width=16)
               Hash Key: bar.a
               ->  Nested Loop Left Join  (cost=0.00..437.00 rows=1 width=16)
                     Join Filter: true
                     ->  Index Scan using bar_pkey on bar  (cost=0.00..6.00 rows=1 width=8)
                           Index Cond: bar.a = foo.a
         ->  Index Scan using zoo_pkey on zoo  (cost=0.00..2.00 rows=1 width=8)
               Index Cond: zoo.a = bar.a
 Settings:  optimizer=on
 Optimizer status: PQO version 2.53.1
(14 rows)
```
The root cause is that when physical join derive its distribution spec, it just
return outer's distribution spec and it ignores the fact that the equiv class
has the same distribution spec. Fixed the issue by add the equiv hashed
distribution spec for inner and outer joins.

[#152327635]

### DIFF
--- a/libgpopt/src/operators/CPhysicalJoin.cpp
+++ b/libgpopt/src/operators/CPhysicalJoin.cpp
@@ -361,6 +361,30 @@ CPhysicalJoin::PdsDerive
 		pdsInner->AddRef();
 		return pdsInner;
 	}
+	else if (CDistributionSpec::EdtHashed == pdsOuter->Edt() &&
+			CDistributionSpec::EdtHashed == pdsInner->Edt() &&
+			(CUtils::FPhysicalInnerJoin(exprhdl.Pop()) ||
+					CUtils::FPhysicalOuterJoin(exprhdl.Pop())))
+	{
+		// If both inner and outer children are hash distributed or
+		// redistributed by the join key, and the join is a inner join
+		// or outer join, set the outer's distribution spec as inner's
+		// equiv hashed distribution spec, and return inner's
+		// distribution spec. This is pretty useful for such case
+		// 'R left join S on R.a = S.a left join T on T.a = S.a'
+		// (R, S, T are distributed by a). Without equiv hashed spec,
+		// Orca will assume (R left join S) is hash distributed by R.a,
+		// ignores the facts that it is also distributed by S.a, and
+		// request a redistribute motion on S.a, which is redundant.
+		// With the equiv hashed spec set for inner distribution, Orca
+		// will know that the equivclass has the same distribution
+		// spec.
+		CDistributionSpecHashed *pdsOuterHashed = CDistributionSpecHashed::PdsConvert(pdsOuter);
+		CDistributionSpecHashed *pdsInnerHashed = CDistributionSpecHashed::PdsConvert(pdsInner);
+		pdsInnerHashed->SetHashedEquiv(pdsOuterHashed);
+		pdsInnerHashed->AddRef();
+		return pdsInnerHashed;
+	}
 
 	// otherwise, return outer distribution
 	pdsOuter->AddRef();
--- a/data/dxl/minidump/FullOuterJoin.mdp
+++ b/data/dxl/minidump/FullOuterJoin.mdp
@@ -214,7 +214,7 @@
         </dxl:Comparison>
       </dxl:LogicalJoin>
     </dxl:Query>
-    <dxl:Plan Id="0" SpaceSize="200">
+    <dxl:Plan Id="0" SpaceSize="195">
       <dxl:GatherMotion InputSegments="0,1" OutputSegments="-1">
         <dxl:Properties>
           <dxl:Cost StartupCost="0" TotalCost="10.380859" Rows="3.000000" Width="16"/>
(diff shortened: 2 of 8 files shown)


libgpopt/src/operators/CPhysicalJoin.cpp
```diff
@@ -361,6 +361,17 @@ CPhysicalJoin::PdsDerive
 		pdsInner->AddRef();
 		return pdsInner;
 	}
+	else if (CDistributionSpec::EdtHashed == pdsOuter->Edt() &&
```
oarap: This `else if` requires a very explanatory comment. I do not understand how this helps. 
1. How do you use the `m_pdshashedEquiv` to decided it is already hashed on a column in the equiv class?
2. How do you know if they are hashed on the columns that are in the same equiv class? 
3. Is checking `CDistributionSpec::EdtHashed == pdsInner->Edt()` enough for that?

oarap: I see what you are doing here, you still need to add the comment.

hsyuan: will add comment

libgpopt/src/operators/CPhysicalJoin.cpp
```diff
@@ -361,6 +361,17 @@ CPhysicalJoin::PdsDerive
...
+			CDistributionSpec::EdtHashed == pdsInner->Edt() &&
```
hsyuan: Need to figure out the strategy for universal distribution spec, e.g. if the inner is a constant table get.

hsyuan:
added detailed comments.

libgpopt/src/operators/CPhysicalJoin.cpp
```diff
@@ -361,6 +361,30 @@ CPhysicalJoin::PdsDerive
...
+			(CUtils::FPhysicalInnerJoin(exprhdl.Pop()) ||
+					CUtils::FPhysicalOuterJoin(exprhdl.Pop())))
+	{
```
dhanashreek89: What about the remaining Physical joins:
```
				EopPhysicalLeftSemiNLJoin,
				EopPhysicalCorrelatedLeftSemiNLJoin,
				EopPhysicalCorrelatedInLeftSemiNLJoin,
				EopPhysicalLeftAntiSemiNLJoin,
				EopPhysicalCorrelatedLeftAntiSemiNLJoin,
				EopPhysicalLeftAntiSemiNLJoinNotIn,
				EopPhysicalCorrelatedNotInLeftAntiSemiNLJoin,
```

Although I currently do not see the redundant redistribute added in following case, so what makes the case you pointed out, special ?
```
explain select * from foo where exists (select 1 from bar left join zoo on zoo.a=bar.a and foo.a=bar.a ) ;
                                         QUERY PLAN
--------------------------------------------------------------------------------------------
 Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..1293.02 rows=1 width=8)
   ->  Hash Semi Join  (cost=0.00..1293.02 rows=1 width=8)
         Hash Cond: foo.a = bar.a
         ->  Hash  (cost=862.01..862.01 rows=34 width=4)
               ->  Result  (cost=0.00..862.01 rows=34 width=4)
                     ->  Hash Left Join  (cost=0.00..862.01 rows=34 width=4)
                           Hash Cond: bar.a = zoo.a
                           ->  Table Scan on bar  (cost=0.00..431.00 rows=34 width=4)
                           ->  Hash  (cost=431.00..431.00 rows=34 width=4)
                                 ->  Table Scan on zoo  (cost=0.00..431.00 rows=34 width=4)
 Optimizer status: PQO version 2.53.4
(12 rows)
```

hsyuan: For semi joins, the output relation are just outer child, there are no tuples from inner child. So the distribution spec of semi is always same with outer child's distribution, like what the current code is doing now.

hsyuan: For this query `explain select * from foo left join bar on foo.a=bar.a left join zoo on zoo.a=bar.a;`, if you change the last `bar.a` to `foo.a` (the query has the same meaning), you won't see redistribute motion.

dhanashreek89: Makes sense. LGTM

hsyuan:
It turned out I mistakenly ran ICW without turning Orca ON, hence didn't pass the pipeline after merging, and also affected the correctness of later PR to support left outer index NL join. So I reverted the 2 PR that are merged in, and opened new PR: https://github.com/greenplum-db/gporca/pull/302

PR #293, opened 2018-01-04
Remove header files included in CMakeLists.txt

hsyuan:
It is not necessary to add these header files, and it is a headache to maintain
the header files, so remove them.


libgpdbcost/CMakeLists.txt
```diff
@@ -6,13 +6,9 @@ include_directories(${CMAKE_CURRENT_SOURCE_DIR}/../libnaucrates/include)
 # for the generated config.h file.
 include_directories(${PROJECT_BINARY_DIR}/libgpos/include/)
 add_library(gpdbcost
-            include/gpdbcost/CCostModelGPDB.h
```
bhuvnesh2703: do you think we can have include_directories to include these header files if it helps in IDE's and takes away the headache of maintaining them or have something like below to search and include the required files (using an example only)
```
file(GLOB_RECURSE SRC_FILES
        "*.c"
        "*.cpp")

file(GLOB_RECURSE HEADER_FILES
	"*.h")
```
Just trying to think how do we maintain a separate list for IDE's.

hsyuan: It is already there, see line 7. How can it help IDE?

hsyuan: What would be the problem for IDE by removing header files here in CMakeLists.txt?

bhuvnesh2703: like for XCode and others to build the project? doesn't it uses the CMakeLists ?

hsyuan: Can you point it to me what would change for IDE if we do the changes in this patch?

bhuvnesh2703: if you take your branch (may need to run cmake -GXcode on cmdline and try to open CCostModel.h in xcode, it will not find it vs on master Xcode will find it.

hsyuan: Don't we just use this to generate Xcode project?
```
file(GLOB_RECURSE SRC_FILES
        "*.cpp")
```

How often would you regenerate the Xcode project?

bhuvnesh2703: naah not in gporca cmakelist..anytime it will do an index we will run into the problem i believe

hsyuan:
Yes, we need headers to make them visible in IDE. Will rethink about it.

PR #298, opened 2018-01-08
Split CICG test into multiple smaller tests

hsyuan:
First stab.


server/include/unittest/gpopt/minidump/CConstTblGetTest.h
```diff
@@ -0,0 +1,35 @@
...
+//	Copyright (C) 2018 Pivotal, Inc.
+//---------------------------------------------------------------------------
+#ifndef GPOPT_CConstTblGetTest_H
+#define GPOPT_CConstTblGetTest_H
+
+#include "gpos/base.h"
+
+namespace gpopt
+{
+	class CConstTblGetTest
```
bhuvnesh2703: may be a common base class for all these?

hsyuan: It has only static attributes and methods, so..

bhuvnesh2703:
LGTM, 
there are some mdp which although didn't had the name (atleast indicating the basis of why they were put into a specific suite) in the newly created test suite, but i believe its fine.

PR #299, opened 2018-01-09
Use GLOB_RECURSE in CMakeLists.txt

hsyuan:
Since everytime we make a code change, we need to touch CMakeLists.txt file and
bump at least the patch version, so it is pretty safe for Orca to use GLOB in
CMakeLists.txt, and it saves a lot of our efforts.


hardikar:
Here's the warning for future reference:
```
Note We do not recommend using GLOB to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate.
```

I'm not sure what we gain from this. It is rare that we add/remove files, and so I have not found it that cumbersome. I'm a soft -1 for this PR; but if others prefer this, I'll just have to touch CMakeLists.txt and that should do it (I think).

libgpdbcost/CMakeLists.txt
```diff
@@ -5,17 +5,15 @@ include_directories(${CMAKE_CURRENT_SOURCE_DIR}/../libgpopt/include)
...
-            include/gpdbcost/CCostModelParamsGPDBLegacy.h
-            src/CCostModelParamsGPDBLegacy.cpp
-            src/ICostModel.cpp
-        )
+
+file(GLOB_RECURSE hdrs ${CMAKE_CURRENT_SOURCE_DIR}/include/*.h
+                       ${CMAKE_CURRENT_SOURCE_DIR}/include/*.inl)
+file(GLOB_RECURSE srcs ${CMAKE_CURRENT_SOURCE_DIR}/src/*.cpp)
+
+# Add headers to make them visible in some IDEs (Clion, VS, Xcode)
```
hardikar: Oh I though it was to ensure a rebuild if a header file changes (because cmake/make can be stupid sometimes).

hsyuan:
The warning doesn't apply for us. As I mentioned in the PR comment, every time we make changes, we are going to bump version. You don't need to touch CMakeLists.txt.

hsyuan:
I recently worked a lot on adding, removing files.

PR #208, opened 2017-07-28
Add new Array Coerce Cast Metadata object

khannaekta:
Currently executor crashes for following query with ORCA ON :
```
CREATE TABLE FOO(a
integer NOT NULL, b double precision[]); SELECT b FROM foo UNION ALL SELECT
ARRAY[90, 90] as Cont_features;
```

In this query, we are appending an integer array (ARRAY[90, 90]) to a double
precision array (foo.b) and hence we need to apply a cast on ARRAY[90, 90] to
generate ARRAY[90, 90]::double precision[].
In gpdb5 there is not direct function available that can cast array of any type
to array of any other type. So in relcache to dxl translator we look into the
array elements and get their type and try to find a cast function for them.
For this query, source type is 23 (integer) and dest type is 701 (double precision)
 and we try to find if we have a conversion function for 23 -> 701.
Since that is available we send that function to ORCA as follows:
```
<dxl:MDCast
Mdid="3.1007.1.0;1022.1.0" Name="float8" BinaryCoercible="false"
SourceTypeId="0.1007.1.0" DestinationTypeId="0.1022.1.0"
CastFuncId="0.316.1.0"/>
```
Here we are misinforming ORCA by specifying that function with id 316 is available
to convert type 1007 (integer array) to 1022 (double precision) array.
However Function id 316 is simple int4 to float8 conversion function and it CAN NOT
convert an array of int4 to array of double precision.
ORCA generates a plan using this function but executor crashes while executing this
function because this function can not handle arrays.

This commit adds a new ArrayCoerceCast MetaData object which will be constructed
when we need to convert an array of one type to another; instead of constructing
Cast Metadata.

`CMDArrayCoerceCastGPDB` extends `CMDCastGPDB` in that it includes necessary information
to generate `CScalarArrayCoerceExpr`.

In Relcache Translator on GPDB, we will construct an object of `CMDArrayCoerceCastGPDB`
instead of `CMDCastGPDB` depending on the coercion path determined by `FCastFunc`.

Added relevant test cases.

PR #301, opened 2018-01-10
ProcessArrayCmp() should handle ScalarArray of constants and non constants

sambitesh:
When there is a comparison between a scalar ident and an array consisting of
constants and non constants, ORCA would not estimate the cardinality correctly.
This is because the ProcessArrayCmp() function would simply return the value of
CStatsPredUnsupported() when the array was not a ConstArray. This patch fixes
the issue by considering Arrays in which constants and non constants are
present together and makes sure that the cardinality is estimated correctly.

e.g.
```
CREATE TABLE bar (a int);

INSERT INTO bar SELECT i FROM generate_series(1,10000) i;
INSERT INTO bar SELECT i%10 FROM generate_series(1,10000) i;
ANALYZE bar;
```

The MCVs are in the range 0-9.

But below queries with and without MCV both estimated the same number of rows before this fix.

```sql
explain select * from bar where a in (1, (select count(*) from bar));
                                              QUERY PLAN
 Gather Motion 3:1  (slice3; segments: 3)  (cost=0.00..1407777.04 rows=8000 width=4)
   ->  Nested Loop  (cost=0.00..1407765.17 rows=2667 width=4)
         Join Filter: public.bar.a = ANY (ARRAY[1::bigint, (count((count())))])
         ->  Broadcast Motion 1:3  (slice2)  (cost=0.00..444.64 rows=3 width=8)
               ->  Aggregate  (cost=0.00..444.64 rows=1 width=8)
                     ->  Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..444.64 rows=1 width=8)
                           ->  Aggregate  (cost=0.00..444.64 rows=1 width=8)
                                 ->  Table Scan on bar  (cost=0.00..443.40 rows=20000 width=1)
         ->  Table Scan on bar  (cost=0.00..443.40 rows=20000 width=4)
 Optimizer status: PQO version 2.51.3
(10 rows)
```

Plan produced for IN clause with non MCV and non constant value in IN clause
```sql
explain select * from bar where a in (1000, (select count(*) from bar));
                                              QUERY PLAN
 Gather Motion 3:1  (slice3; segments: 3)  (cost=0.00..1407777.04 rows=8000 width=4)
   ->  Nested Loop  (cost=0.00..1407765.17 rows=2667 width=4)
         Join Filter: public.bar.a = ANY (ARRAY[1000::bigint, (count((count())))])
         ->  Broadcast Motion 1:3  (slice2)  (cost=0.00..444.64 rows=3 width=8)
               ->  Aggregate  (cost=0.00..444.64 rows=1 width=8)
                     ->  Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..444.64 rows=1 width=8)
                           ->  Aggregate  (cost=0.00..444.64 rows=1 width=8)
                                 ->  Table Scan on bar  (cost=0.00..443.40 rows=20000 width=1)
         ->  Table Scan on bar  (cost=0.00..443.40 rows=20000 width=4)
 Optimizer status: PQO version 2.51.3
(10 rows)
```


sambitesh:
We will bump the ORCA version before merging.

libnaucrates/src/statistics/CStatsPredUtils.cpp
```diff
@@ -918,7 +918,9 @@ CStatsPredUtils::ProcessArrayCmp
 	}
 	CExpression *pexprArray = CUtils::PexprScalarArrayChild(pexprPred);
 
-	BOOL fCompareToConst = CPredicateUtils::FCompareIdentToConstArray(pexprPred) || CPredicateUtils::FCompareCastIdentToConstArray(pexprPred);
+	BOOL fCompareToConst = CPredicateUtils::FCompareIdentToConstArray(pexprPred) ||
+						   CPredicateUtils::FCompareCastIdentToConstArray(pexprPred) ||
```
dhanashreek89: Do we need `FCompareIdentToConstArray` now? Wouldn't `FCompareScalarIdentToConstAndScalarIdentArray()` cover it as well?

libgpopt/src/base/CCastUtils.cpp
```diff
@@ -161,6 +161,18 @@ CCastUtils::FBinaryCoercibleCast
 	CScalarCast::PopConvert(pop)->FBinaryCoercible();
 }
 
+BOOL
+CCastUtils::FCast
```
hsyuan: FCast -> FScalarCast

libgpopt/src/base/CCastUtils.cpp
```diff
@@ -161,6 +161,18 @@ CCastUtils::FBinaryCoercibleCast
...
+	(
+	CExpression *pexpr
+	)
+{
+	GPOS_ASSERT(NULL != pexpr);
+	COperator *pop = pexpr->Pop();
+	
+	return COperator::EopScalarCast == pop->Eopid();
+}
+
```
dhanashreek89: Why not use `CCastUtils::FCast()` ?

libgpopt/src/base/CCastUtils.cpp
hsyuan: In general, this function should be in CUtils.cpp

libnaucrates/src/statistics/CStatsPredUtils.cpp
dhanashreek89: variable name `fCompareToConst`  should be changed now that we support comparing array of constants and non-constants

libgpopt/src/base/CCastUtils.cpp
```diff
@@ -161,6 +161,18 @@ CCastUtils::FBinaryCoercibleCast
 	CScalarCast::PopConvert(pop)->FBinaryCoercible();
```
dhanashreek89: Could we rewrite `FBinaryCoercibleCast()` as follows now:
```
return FCast(pexpr) && CScalarCast::PopConvert(pop)->FBinaryCoercible();
```

libgpopt/src/base/CCastUtils.cpp
dhanashreek89: We moved all cast related utilities to from CUtils.cpp to CCastUtils.cpp in https://github.com/greenplum-db/gporca/pull/208

libnaucrates/src/statistics/CStatsPredUtils.cpp
oarap: True, `FCompareScalarIdentToConstAndScalarIdentArray` should do the job

sambitesh:
@dhanashreek89 @hsyuan @oarap Addressed PR comments.

server/src/unittest/gpopt/minidump/CICGTest.cpp
```diff
@@ -40,6 +40,8 @@ ULONG CICGTest::m_ulNegativeIndexApplyTestCounter = 0;
 // minidump files
 const CHAR *rgszFileNames[] =
 	{
+		"../data/dxl/minidump/CastedInClauseWithMCV.mdp",
+		"../data/dxl/minidump/InClauseWithMCV.mdp",
```
hsyuan: please don't add tests into CICGTest

sambitesh: We were thinking the same thing too. Any suggestions on where to move it?

hsyuan:
You can create new test for testing arrays

~ ~ ~
Haisheng Yuan

PR #305, opened 2018-01-16
Auto generate minidump test classes

hsyuan:
Update server/CMakeLists.txt to generate minidump tests automatically. The original process to add minidump tests, especially adding new test classes that just test minidumps was rather labor intensive (at least to me). It is prohibiting Orca developers creating new tests suites, or move minidumps to others test suites, or rename the test suites to reveal its real intention. Just take a look how large the original CICGTest was. Sometimes, when we just add 1 or 2 minidump files, don't know where to put, and it doesn't seem worthwhile to add a new test class, then a lot of minidumps are accumulating in some irrelevant test classes, such as CICGTest. This PR eases the work of adding new test suites. Adding new test classes and regrouping minidumps couldn't be so easy.


server/CMakeLists.txt
```diff
@@ -49,6 +34,150 @@ function(add_orca_test TEST_NAME)
...
+IndexOnLeaf-SingleColumnHeterogenousIndexOnRoot-1-HEAP IndexOnLeaf-SingleColumnHeterogenousIndexOnRoot-2-AO
+IndexOnLeaf-SingleColumnHeterogenousIndexOnRoot-2-HEAP NonOverlappingHomogenousIndexesOnRoot-ANDPredicate-HEAP
+OverlappingHomogenousIndexesOnRoot-ANDPredicate-HEAP OverlappingHomogenousIndexesOnRoot-HEAP
+SingleColumnHomogenousIndexOnRoot-AO SingleColumnHomogenousIndexOnRoot-HEAP;
+
+CEquivClassTest:
+Equiv-HashedDistr-1 Equiv-HashedDistr-2 EquivClassesAndOr EquivClassesLimit IndexNLJoin_Cast_NoMotion
+")
+
+set(mdp_test_hdr_dir "${CMAKE_CURRENT_SOURCE_DIR}/include/unittest/gpopt/minidump")
```
d: -1

Generated files should go into either [`CMAKE_CURRENT_BINARY_DIR`](https://cmake.org/cmake/help/v3.1/variable/CMAKE_CURRENT_BINARY_DIR.html) or [`PROJECT_BINARY_DIR`](https://cmake.org/cmake/help/v3.1/variable/PROJECT_BINARY_DIR.html). This is both a best practice and a precedence, where our `config.h` was generated into `PROJECT_BINARY_DIR` (I personally prefer `CMAKE_CURRENT_BINARY_DIR` because it's more composable, but that aside).

d:
Empirically my biggest pain with adding a new test suite has been [mucking around with main.cpp](https://github.com/greenplum-db/gporca/blob/f30a0b972865a46f5af4956ef75f88667506377f/server/src/startup/main.cpp#L138) and manually adding the new test there. Does this PR address that?

Jesse

server/CMakeLists.txt
hsyuan: best practice doesn't apply to all the case and doesn't mean we have to do this.

hsyuan: Ah, looks worth a try. I will try it.

hsyuan:
yes

server/CMakeLists.txt
hsyuan: addressed comments, thanks!

hsyuan:
addressed comment. please take a look.

d:
I'm reviewing this

d:
Is it just me or is the new test suite not equivalent to the old test suite?

server/src/unittest/gpopt/minidump/CICGTest.cpp
```diff
@@ -89,6 +89,7 @@ const CHAR *rgszFileNames[] =
 		"../data/dxl/minidump/TPCDS-39-InnerJoin-JoinEstimate.mdp",
 		"../data/dxl/minidump/TPCH-Partitioned-256GB.mdp",
 		"../data/dxl/minidump/Tpcds-NonPart-Q70a.mdp",
+		"../data/dxl/minidump/Tpcds-10TB-Q37-NoIndexJoin.mdp",
```
dhanashreek89: So now that we are running this test for retail build as well;  will this have any impact on testing time as the old comment says ? Just wondering if you have noticed any.

hsyuan: there is no change with master on this.

hsyuan: master only run this when non debug mode

hsyuan:
There is no change for this test. We only run this test for retail build.

~ ~ ~
Haisheng Yuan

On Thu, Jan 18, 2018 at 1:33 PM, Dhanashree Kashid <notifications@github.com

server/CMakeLists.txt
```diff
@@ -49,6 +36,152 @@ function(add_orca_test TEST_NAME)
...
+
+  # generate header/src file name for current test
+  set(header_file ${mdp_test_hdr_dir}/${test_name})
+  set(source_file ${mdp_test_src_dir}/${test_name})
+
+  # auto generate test class from template
+  configure_file(${mdp_test_hdr_tpl} ${header_file})
+  configure_file(${mdp_test_src_tpl} ${source_file})
+
+  update_file(${header_file}.h ${header_file})
```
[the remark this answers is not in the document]
hsyuan: nice, i will change it.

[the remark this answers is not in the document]
hsyuan: It turns out it has some issues. I will leave it for now. Will revisit it later.

hsyuan:
pushed, thanks all for the review!

PR #306, opened 2018-01-16
Remove fEnableExtendedAsserts, which is useless

hsyuan:
It is adding a lot of debug test runtime overhead without adding any value.

### DIFF
--- a/server/src/unittest/gpopt/minidump/CIndexTest.cpp
+++ b/server/src/unittest/gpopt/minidump/CIndexTest.cpp
@@ -120,23 +120,12 @@ const CHAR *rgszIndexFileNames[] =
 GPOS_RESULT
 CIndexTest::EresUnittest()
 {
-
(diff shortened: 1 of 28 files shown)


server/src/unittest/gpopt/minidump/CIndexTest.cpp
```diff
@@ -120,24 +120,13 @@ const CHAR *rgszIndexFileNames[] =
 GPOS_RESULT
 CIndexTest::EresUnittest()
 {
-
-#ifdef GPOS_DEBUG
-	// enable extended asserts after running test
-	fEnableExtendedAsserts = true;
```
hsyuan: How dumb is this one! It is true! That is why the debug test is so slow!

hsyuan:
Thanks for the review, pushed to master. version is NOT bumped. I don't want to release a version for this commit.

PR #310, opened 2018-01-18
Suggest additional hint in common issues

doty-pivotal:
Add a hint to rebuild the shared library cache if the user follows the suggestion to add the 
usr/local/lib path to /etc/ld.so.conf file.


dhanashreek89:
Added [ci skip] in the commit message to skip the ORCA CI run.

PR #316, opened 2018-01-24
Do not maintain mdidType separately in CDXLScalarIdent

hardikar:
This information can be easily derived from CDXLColRef member of
CDXLScalarIdent. This now mirrors what is done in the ORCA types
CScalarIdent and CColRef.

Also bump version to 2.54.0.


vraghavan78:
does DXL to Expr need to change?

dhanashreek89:
>> does GPDB Translators or DXL to Expr need to change?

Yes. The corresponding GPDB PR : https://github.com/greenplum-db/gpdb/pull/4410

d:
It was in our way of supporting `typmod`

PR #322, opened 2018-01-30
Don't convert LOJ to Inner Join with IDF/INDF

melanieplageman:
This PR resolves issue: https://github.com/greenplum-db/gpdb/issues/4433
Previously, Orca would produce wrong results in the case of a `LEFT OUTER JOIN` with a predicate containing `IS DISTINCT FROM`

This was because, during pre-processing, Orca converted LOJ to `INNER JOIN` for most null-filtering scalar comparisons, which does not hold true for IDF/INDF.

When evaluating the expression with a NULL on the outer side and a non-NULL value on the inner side, if we were using simple equality/inequality it evaluates to NULL(and can be excluded from the result set making it safe to convert to an INNER JOIN), but if using IDF, it evaluates to true so we cannot exclude it from the result set.


libgpopt/src/operators/CNormalizer.cpp
```diff
@@ -372,7 +372,11 @@ CNormalizer::FSimplifySelectOnOuterJoin
 	CExpression *pexprOuterJoinPred = (*pexprOuterJoin)[2];
 
 	CColRefSet *pcrsOutput = CDrvdPropRelational::Pdprel(pexprOuterJoinInnerChild->PdpDerive())->PcrsOutput();
-	if (!GPOS_FTRACE(EopttraceDisableOuterJoin2InnerJoinRewrite) &&
```
vraghavan78: can the join predicate be a composite predicate? For instance:

```sql
explain select * from src left join trg on (src_k = trg_k or src_v is distinct from trg_v) where src_v is distinct from trg_v;
```

server/CMakeLists.txt
```diff
@@ -109,7 +109,17 @@ OverlappingHomogenousIndexesOnRoot-ANDPredicate-HEAP OverlappingHomogenousIndexe
...
+
+COuterJoinTest:
+LOJ-Condition-False FOJ-Condition-False FullOuterJoin FullOuterJoin2
+LeftJoin-With-Pred-On-Inner LeftJoin-With-Pred-On-Inner2
+LeftJoin-With-Col-Const-Pred LeftJoin-With-Coalesce LOJWithFalsePred
+LOJ-IsNullPred Select-Proj-OuterJoin OuterJoin-With-OuterRefs Join-Disj-Subqs
+Join-Dist-Key-No-Motion RightJoin-NonDistKey-EquivClass EffectOfLocalPredOnJoin
+EffectOfLocalPredOnJoin2 EffectOfLocalPredOnJoin3
+LeftJoin-UnsupportedFilter-Cardinality LeftOuter2InnerUnionAllAntiSemiJoin
+LOJ-IDF-NotInner
```
vraghavan78: any new tests added?

libgpopt/src/operators/CNormalizer.cpp
```diff
@@ -372,7 +372,11 @@ CNormalizer::FSimplifySelectOnOuterJoin
...
+
+	// if the predicate is an IDF/INDF, it is not NULL filtering, so we should not convert it to an inner join
+	if (CPredicateUtils::FIDF(pexprPred) || CPredicateUtils::FINDF(pexprPred))
+		return false;
+	else if (!GPOS_FTRACE(EopttraceDisableOuterJoin2InnerJoinRewrite) &&
 		CPredicateUtils::FNullRejecting(pmp, pexprPred, pcrsOutput))
```
hsyuan: you should change `FNullRejecting`

hsyuan: In addition, pexprPred can be a conjunction or disjunction.

vraghavan78: If you are going to add to this function, please rename it a FConvertLOJ2Inner or something more meaningful

server/CMakeLists.txt
melanieplageman: Yes, added the  new minidump.

libgpopt/src/operators/CNormalizer.cpp
melanieplageman: We have added an Eber implementation to CScalarIsDistinctFrom instead

libgpopt/src/operators/CScalarIsDistinctFrom.cpp
```diff
@@ -68,5 +68,15 @@ CScalarIsDistinctFrom::PopCommutedOp
...
 }
 
+CScalar::EBoolEvalResult
+CScalarIsDistinctFrom::Eber
+	(
+	DrgPul *
+	)
+const
+{
+	return EberUnknown;
```
hsyuan: Why Unknown? If both of the children is NULL, we should return TRUE, otherwise return FALSE.

melanieplageman:
Closing this as it needs some more attention.

PR #324, opened 2018-02-01
Do not push DTS into inner child with more than 1 join depth of NLJ

hsyuan:
For nestloop join with inner child being another join, and if there is a partition consumer
in the inner child, don't request dynamic partition elimination on the inner child,
because we may end up with materializing nestloop join's inner child, which is a join
contains dynamic partition consumer. This patch is to make sure that we don't generate
plan alternatives that materialize a join that contains DTS.


libgpopt/src/search/CGroupExpression.cpp
```diff
@@ -482,8 +482,6 @@ CGroupExpression::PccComputeCost
 			CCost cost = CostCompute(pmp, pcc);
 			pcc->SetCost(cost);
 		}
-		GPOS_ASSERT_IMP(COptCtxt::FAllEnforcersEnabled(), fValid &&
```
bhuvnesh2703: Why do we need to remove this? Removing the assertion suggests for this one case sounds like hacky. Could we have a better way to identify and ignore the specific invalid specs.

hsyuan:
Need more thought, so close this PR.

PR #329, opened 2018-02-08
support type modifiers in ORCA

d:
Sister PR at https://github.com/greenplum-db/gpdb/pull/4510


server/src/unittest/gpopt/translate/CTranslatorDXLToExprTest.cpp
```diff
@@ -275,7 +275,7 @@ namespace
 
 			void AddIntColumn(CWStringConst strColumnName, int iAttno, BOOL fNullable)
 			{
-				CColumnDescriptor *pcoldesc = GPOS_NEW(m_pmp) CColumnDescriptor(m_pmp, m_pmdtypeint4, CName(&strColumnName), iAttno, fNullable);
+				CColumnDescriptor *pcoldesc = GPOS_NEW(m_pmp) CColumnDescriptor(m_pmp, m_pmdtypeint4, -1 /* iTypeModifier */, CName(&strColumnName), iAttno, fNullable);
```
vraghavan78: can we have this as a macro?

hardikar: Check again. PR was opened a little too early..

libgpopt/include/gpopt/operators/CScalarFunc.h
```diff
@@ -43,6 +43,8 @@ namespace gpopt
 			// return type
 			IMDId *m_pmdidRetType;
 
+			const INT m_iRetTypeModifier;
```
vraghavan78: for consistency add a comment, else remove all useless comments

libgpopt/src/base/CUtils.cpp
```diff
@@ -1972,24 +1972,6 @@ CUtils::PexprAggFunc
 	return GPOS_NEW(pmp) CExpression(pmp, popScAggFunc, pdrgpexpr);
 }
 
-// construct a scalar function
-CExpression *
-CUtils::PexprScalarFunc
```
vraghavan78: you don't need this any more?

libnaucrates/include/naucrates/dxl/operators/CDXLDatum.h
```diff
@@ -93,10 +96,16 @@ namespace gpdxl
 				return m_pmdidType;
 			}
 
+			INT
+			ITypeModifier() const
+			{
+				return m_iTypeModifier;
+			}
+
```
bhuvnesh2703: all the other methods have the implementations in cpp, so why not move it there.

libgpopt/src/base/CColumnFactory.cpp
```diff
@@ -193,18 +196,14 @@ CColumnFactory::PcrCreate
 	ULONG ulOpSource
 	)
 {
-	CName *pnameCopy = GPOS_NEW(m_pmp) CName(m_pmp, name); 
-	CAutoP<CName> a_pnameCopy(pnameCopy);
-
```
d: Duh, I just realized why the `CAutoP` was needed here: if the `GPOS_NEW` below throws, we'd be leaking the `CName` object here.

The pattern of using `CAutoP<T>::PtReset()` is really indicative of needing something of a move semantic in our codebase, but that's out of scope of this PR...

PR #330, opened 2018-02-13
Fix handling of NULL datum in CDefaultComparator

sambitesh:
Consider the query below

```
create table test_country (a char(40));
select * from test_country where a not in (NULL,'');
```

`CDefaultComparator` is not able to handle NULL datum in above case.

Comparison in cases of text/char dataypes are converted into comparator expression
with result being a result expression like below.

```
+--CScalarCmp (>)
   |--CScalarConst (161096236.000)
   +--CScalarConst (161104428.000)

Result Expression:

+--CScalarConst (0)
```

The result expression is either a 0 or 1 CScalarConstant. But in case of NULL datums,
the comparator always evaluates to NULL.

```
+--CScalarCmp (>)
        |--CScalarConst (161096236.000)
        +--CScalarConst (null)

Result Expression:

 +--CScalarConst (null)
```

This commits handles the case of NULL datum by assuming :
1. NULL is less than everything not NULL.
2. NULL is equal to NULL.

Will bump the ORCA version before merging.


libgpopt/src/base/CDefaultComparator.cpp
```diff
@@ -122,6 +121,13 @@ CDefaultComparator::FEqual
 	}
 	CAutoMemoryPool amp;
 
+	// NULL datum is a special case and is being handled here. Assumptions made are
+	// NULL is less than everything else. NULL = NULL
+	if (pdatum1->FNull() && pdatum2->FNull())
+	{
+		return true;
```
vraghavan78: shouldn't this be unknown?

libgpopt/src/base/CDefaultComparator.cpp
```diff
@@ -155,6 +161,13 @@ CDefaultComparator::FLessThan
 	}
 	CAutoMemoryPool amp;
 
+	// NULL datum is a special case and is being handled here. Assumptions made are
+	// NULL is less than everything else. NULL = NULL
```
vraghavan78: talked with @sambitesh this whole logic only makes sense in interval computations

libgpopt/src/base/CDefaultComparator.cpp
sambitesh: This comparator is for interval calculation. So the comparator should either return 'less than', 'greater than' or 'equal to'. I will update the same in comments.

libgpopt/src/base/CDefaultComparator.cpp
```diff
@@ -254,6 +282,11 @@ CDefaultComparator::FGreaterThanOrEqual
...
+	if (pdatum2->FNull())
```
vraghavan78: Please put the following comment:

<code>
if (pdatum2->FNull())
{
// return true since either:
// 1. pdatum2 is NULL and pdatum1 is not NULL. Since NULL is considered less that not null values for interval computations we return true
// 2. when pdatum1 and pdatum2 are both NULL so they are equal for interval computation

return true;

}
<code>

PR #331, opened 2018-02-13
Avoid deadlock with Correlated Left Outer NestLoop Join

bhuvnesh2703:
In general, any node projecting elements from both the children, i.e outer and inner may impose a deadlock if there are motions in both children.

Plans containing CorrelatedLeftOuterNLJoin as shown below, where the project list contains elements from the outer child (tab3.i) and inner child (subplan), may result in deadlock if there is motion in both the children.

For the query:
```sql
explain select tab3.i, (select tab2.k from tab2 where tab2.i = tab3.i
limit 1) as subq from tab3 left join tab1 on tab1.j = tab3.j;
```
ORCA produces the below physical plan:

```sql
+--CPhysicalComputeScalar
   |--CPhysicalMotionGather(master)
   |  +--CPhysicalCorrelatedLeftOuterNLJoin
   |     |--CPhysicalLeftOuterHashJoin =====> outer child
   |     |  |--CPhysicalMotionHashDistribute =====> motion
   |     |  |  +--CPhysicalTableScan "tab3" ("tab3")
   |     |  |--CPhysicalMotionHashDistribute
   |     |  |  +--CPhysicalTableScan "tab1" ("tab1")
   |     |  +--CScalarCmp (=)
   |     |     |--CScalarIdent "j" (11)
   |     |     +--CScalarIdent "j" (1)
   |     |--CPhysicalLimit <empty>     =====> inner child
   |     |  |--CPhysicalFilter
   |     |  |  |--CPhysicalSpool
   |     |  |  |  +--CPhysicalMotionBroadcast ====> motion
   |     |  |  |     +--CPhysicalTableScan "tab2" ("tab2")
   |     |  |  +--CScalarCmp (=)
   |     |  |     |--CScalarIdent "i" (20)
   |     |  |     +--CScalarIdent "i" (0)
   |     |  |--CScalarConst (0)
   |     |  +--CScalarConst (1)
   |     +--CScalarConst (1)
   +--CScalarProjectList
      +--CScalarProjectElement "subq" (30)
         +--CScalarIdent "k" (22)
```

The above physical plan is converted to the below GPDB plan in
Expression to DXL Translator where CorrelatedLeftOuterNLJoin is
translated to a Result node which contains 2 project elements
1. SubPlan 1 => Inner child of CorrelatedLeftOuterNLJoin
2. tab3.i => Column projected from outer child of CorrelatedLeftOuterNLJoin

```sql
                                                        QUERY PLAN
--------------------------------------------------------------------------------------------------------------------------
 Result  (cost=0.00..170041830.77 rows=1664667 width=12)
   ->  Gather Motion 3:1  (slice4; segments: 3)  (cost=0.00..170041770.84 rows=1000 width=12)
         ->  Result  (cost=0.00..170041770.80 rows=334 width=12)
	     Output: tab3.i, (SubPlan 1)
	     ->  Hash Left Join  (cost=0.00..886.71 rows=1664667 width=4)
		   Hash Cond: tab3.j = tab1.j
		   ->  Redistribute Motion 3:3  (slice1; segments: 3)  (cost=0.00..431.30 rows=1665 width=8)
			 Hash Key: tab3.j
			 ->  Table Scan on tab3  (cost=0.00..431.10 rows=1665 width=8)
		   ->  Hash  (cost=431.02..431.02 rows=334 width=4)
			 ->  Redistribute Motion 3:3  (slice2; segments: 3)  (cost=0.00..431.02 rows=334 width=4)
			       Hash Key: tab1.j
			       ->  Table Scan on tab1  (cost=0.00..431.01 rows=334 width=4)
               SubPlan 1
                 ->  Limit  (cost=0.00..164733.72 rows=1 width=1)
                       ->  Result  (cost=0.00..164733.72 rows=500 width=1)
                             Filter: tab2.i = $0
                             ->  Materialize  (cost=0.00..431.12 rows=500 width=4)
                                   ->  Broadcast Motion 3:3  (slice3; segments: 3)  (cost=0.00..431.12 rows=500 width=4)
                                         ->  Table Scan on tab2  (cost=0.00..431.01 rows=167 width=4)
```
Notice the fact that, SubPlan contains a BroadCast motion, and the outer child of result node i.e Hash Left Join contains a Redistribute Motion on top of tab3 which is producing the column tab3.i. Such Result node may run into a deadlock situation.

This commit avoids the deadlock by creating a materialize node on top of the outer child of the Correlated-Left-Outer-NL-Join while it is being translated to a Result Node with a project list that contains the subplan and an output column from the outer child.  Adding a materialize ensures that the child (hash left join) of Result node is completely exhausted before we move to subplan execution. In case the outer child of Correlated-Left-Outer-NL-Join contains non-streaming operators like sort, hash aggregate, materialize or even a gather motion, then the outer child does not impose a deadlock hazard.  Similarly if the inner child translated as subplan does not contain broadcast motion, there is no deadlock hazard.  In such cases a materialize is not added.

GPDB Plan after the fix will be as below:

```sql
                                                        QUERY PLAN
--------------------------------------------------------------------------------------------------------------------------
 Result  (cost=0.00..170041830.77 rows=1664667 width=12)
   ->  Gather Motion 3:1  (slice4; segments: 3)  (cost=0.00..170041770.84 rows=1000 width=12)
         ->  Result  (cost=0.00..170041770.80 rows=334 width=12)
               ->  Materialize  (cost=0.00..886.71 rows=1664667 width=4)
                     ->  Hash Left Join  (cost=0.00..886.71 rows=1664667 width=4)
                           Hash Cond: tab3.j = tab1.j
                           ->  Redistribute Motion 3:3  (slice1; segments: 3)  (cost=0.00..431.30 rows=1665 width=8)
                                 Hash Key: tab3.j
                                 ->  Table Scan on tab3  (cost=0.00..431.10 rows=1665 width=8)
                           ->  Hash  (cost=431.02..431.02 rows=334 width=4)
                                 ->  Redistribute Motion 3:3  (slice2; segments: 3)  (cost=0.00..431.02 rows=334 width=4)
                                       Hash Key: tab1.j
                                       ->  Table Scan on tab1  (cost=0.00..431.01 rows=334 width=4)
               SubPlan 1  (slice4; segments: 3)
                 ->  Limit  (cost=0.00..164733.72 rows=1 width=1)
                       ->  Result  (cost=0.00..164733.72 rows=500 width=1)
                             Filter: tab2.i = $0
                             ->  Materialize  (cost=0.00..431.12 rows=500 width=4)
                                   ->  Broadcast Motion 3:3  (slice3; segments: 3)  (cost=0.00..431.12 rows=500 width=4)
                                         ->  Table Scan on tab2  (cost=0.00..431.01 rows=167 width=4)
```

While constructing a Result node with a new/remapped project list, this commit checks if the Result node imposes a motion hazard, If yes, we add a Materialize for precaution. This happens in `PdxlnComputeScalar()` and `PdxlnRemapOutputColumns()`.


libgpopt/src/translate/CTranslatorExprToDXLUtils.cpp
```diff
@@ -2429,4 +2429,197 @@ CTranslatorExprToDXLUtils::ExtractCastMdids
...
+CTranslatorExprToDXLUtils::FDXLOpExists
+	(
+	const CDXLOperator *pop,
+	const gpdxl::Edxlopid *peopid,
+	ULONG ulOps
+	)
+{
+	GPOS_ASSERT(NULL != pop);
+	GPOS_ASSERT(NULL != peopid);
+
```
oarap: `GPOS_ASSERT(GPOS_ARRAY_SIZE(peopid) == ulOps);`

libgpopt/src/translate/CTranslatorExprToDXLUtils.cpp
```diff
@@ -2429,4 +2429,197 @@ CTranslatorExprToDXLUtils::ExtractCastMdids
 	*ppmdidCastFunc = popCast->PmdidFunc();
 }
 
+BOOL
+CTranslatorExprToDXLUtils::FDXLOpExists
```
oarap: I would recommend to rethink this as a bitset to avoid a for loop. We do not have much motions in the end.

dhanashreek89: Yes that can be implemented; however currently this utility function is only being used for checking motions and we do not have a huge list of the motions to compare against. So implementing a bitset would not give us any performance gain here. In future if we have a different use-case of this function, this can be a bitset.

dhanashreek89:
>> Minor issues, when addressed it is good to go, nice work!!!

Thanks! Addresses the review comments.

libgpopt/src/translate/CTranslatorExprToDXL.cpp
```diff
@@ -7909,4 +7891,104 @@ CTranslatorExprToDXL::UlPosInArray
 	return ulSize;
 }
 
+CDXLNode *
```
vraghavan78: the code needs some detailed comments to be readable. Please do so.

libgpopt/src/translate/CTranslatorExprToDXLUtils.cpp
```diff
@@ -2429,4 +2429,195 @@ CTranslatorExprToDXLUtils::ExtractCastMdids
...
+		// input project list.
+		// If yes, then we have detected a motion hazard for the parent Result node.
+		CBitSet *pbsPrjList = GPOS_NEW(pmp) CBitSet(pmp);
+		ExtractIdentColIds((*pdxln)[0], pbsPrjList);
+		BOOL fDisJoint = pbsPrjCols->FDisjoint(pbsPrjList);
+		pbsPrjList->Release();
+
+		if (fDisJoint)
+			return false;
+		return true;
```
hsyuan: why not `return !fDisJoint`?

libgpopt/src/translate/CTranslatorExprToDXLUtils.cpp
```diff
@@ -2429,4 +2429,195 @@ CTranslatorExprToDXLUtils::ExtractCastMdids
...
+	}
+
+	// recursively check children
+	const ULONG ulArity = pdxln->UlArity();
+
+	// In ORCA, inner child of Hash Join is always exhausted first (prefect_inner is True always),
```
hsyuan: Orca knows nothing about `prefect_inner`, it is postgres specific, we can remove the mention of `prefect_inner`. btw, the spelling is wrong.

libgpopt/src/translate/CTranslatorExprToDXLUtils.cpp
```diff
@@ -2429,4 +2429,195 @@ CTranslatorExprToDXLUtils::ExtractCastMdids
...
+				return true;
+			}
+		}
+	}
+
+	return false;
+}
+
+BOOL
+CTranslatorExprToDXLUtils::FMotionHazardSafeOp
```
hsyuan: Am I the only who thinks the method name is counter intuitive? I would rather reverse the meaning of this function to `FMotionHazardOp`

libgpopt/src/translate/CTranslatorExprToDXLUtils.cpp
```diff
@@ -2429,4 +2429,195 @@ CTranslatorExprToDXLUtils::ExtractCastMdids
...
+		{
+			return true;
+		}
+	}
+
+	return false;
+}
+
+BOOL
+CTranslatorExprToDXLUtils::FProjListContainsSubplanWithBroadCast
```
hsyuan: Will redistribute motion have the same issue?

libgpopt/src/translate/CTranslatorExprToDXLUtils.cpp
vraghavan78: +1 for FMotionHazardOp

libgpopt/src/translate/CTranslatorExprToDXLUtils.cpp
dhanashreek89: `GPOS_ARRAY_SIZE` can not be used with a pointer to the array; it returns the size of pointer which is 8. Hence this assert can not be added here.

libgpopt/src/translate/CTranslatorExprToDXLUtils.cpp
dhanashreek89: The intent of this function is to check for all the blocking operators (Materialize, Sort, HashAgg) and Gather Motion since all these operators neutralize the motion hazard their parent may have. 

Are you suggesting to rename this function to `FMotionHazardOp` and check for all other operators here except the blocking ones listed above ?

Or are you suggesting to **just** rename this function with a more intuitive name?

Issue #309, opened 2018-01-18
Orca nestloop join should not broadcast co-located tables


hsyuan:
Repro:
```sql
create table R(a int, b int) distributed by(a);
create table S(a int, b int) distributed by(a);
create index s_idx_b on S(b);
set optimizer_enable_hashjoin=off;

explain select * from R, S where R.a=S.a;
                                         QUERY PLAN
---------------------------------------------------------------------------------------------
 Gather Motion 3:1  (slice2; segments: 3)  (cost=0.00..1324032.79 rows=1 width=16)
   ->  Nested Loop  (cost=0.00..1324032.79 rows=1 width=16)
         Join Filter: r.a = s.a
         ->  Broadcast Motion 3:3  (slice1; segments: 3)  (cost=0.00..431.00 rows=1 width=8)
               ->  Table Scan on r  (cost=0.00..431.00 rows=1 width=8)
         ->  Table Scan on s  (cost=0.00..431.00 rows=1 width=8)
 Settings:  optimizer=on
 Optimizer status: PQO version 2.53.9
(8 rows)

explain select * from R join S on R.a=S.a and R.b=S.b;
                                         QUERY PLAN
---------------------------------------------------------------------------------------------
 Gather Motion 3:1  (slice2; segments: 3)  (cost=0.00..437.00 rows=1 width=16)
   ->  Nested Loop  (cost=0.00..437.00 rows=1 width=16)
         Join Filter: true
         ->  Broadcast Motion 3:3  (slice1; segments: 3)  (cost=0.00..431.00 rows=1 width=8)
               ->  Table Scan on r  (cost=0.00..431.00 rows=1 width=8)
         ->  Index Scan using s_idx_b on s  (cost=0.00..6.00 rows=1 width=8)
               Index Cond: s.b = r.b
               Filter: r.a = s.a
 Settings:  optimizer=on
 Optimizer status: PQO version 2.53.9
(10 rows)
```

We should not broadcast table `R` in both queries, because R and S are co-located.

hardikar:
Solved for simple NLJ with  pr #409

vraghavan78:
vraghavan=# explain select * from R, S where R.a=S.a;
                                    QUERY PLAN
-----------------------------------------------------------------------------------
 Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..1324032.35 rows=1 width=16)
   ->  Nested Loop  (cost=0.00..1324032.35 rows=1 width=16)
[...]

PR #332, opened 2018-02-15
Generate a Nested loop Join alternative with no broadcast motion

khannaekta:
This PR addresses the gporca issue: https://github.com/greenplum-db/gporca/issues/309

Prior to this PR, for Inner Join queries with join predicate on the distribution column ORCA produced a plan as follows:

```
-> Gather Motion
   ->  Nested Loop
         Join Filter: distribcolforouter = distribcolforinner
         ->  Broadcast Motion
               ->  Table Scan on outerchild
         ->  Table Scan on innerchild
```
The above plan can be optimized further by removing the broadcast motion added on top of the outer child. Since the tuples are already co-located, the broadcast motion is unnecessary and can be dropped.
For `PhysicalInnerNLJoin`, ORCA send 2 requests :-
Request 0: requests outer child to be ANY distribution and inner as Replicated or have matching 
                   distribution if outer delivered Universal/Singleton
Request 1: requests outer child to be Replicated and inner as Non-Singleton or Singleton if outer
                   delivered Universal distribution
This PR, enhances Request 0 to additionally check if the outer child delivers a hash distribution then require the inner child to be hashed on the join predicate matching the outer child's hashed column. Due to this additional request, we see plan space size changes in the mdp files. 

Now, ORCA produces the following plan:
```
-> Gather Motion
   ->  Nested Loop
         Join Filter: distribcolforouter = distribcolforinner
         ->  Table Scan on outerchild
         ->  Table Scan on innerchild
```
There is a corresponding GPDB PR() for test updates for plan changes due to this fix


libgpopt/src/operators/CPhysicalInnerNLJoin.cpp
```diff
@@ -108,6 +111,50 @@ CPhysicalInnerNLJoin::PdsRequired
 
 	if (GPOS_FTRACE(EopttraceDisableReplicateInnerNLJOuterChild) || 0 == ulOptReq)
```
vraghavan78: logic seems wrong, why do we care if we disable replicate inner child or not?

libgpopt/src/operators/CPhysicalInnerNLJoin.cpp
```diff
@@ -128,5 +175,57 @@ CPhysicalInnerNLJoin::PdsRequired
 	return GPOS_NEW(pmp) CDistributionSpecNonSingleton();
 }
 
+// search the given array of predicates for an equality predicate
+// that has one side equal to the given expression,
+// if found, return the other side of equality, otherwise return NULL
+CExpression *
+CPhysicalInnerNLJoin::PexprMatchEqualitySide
```
vraghavan78: is there a similar logic that hash join does? If so, can we reuse it?

khannaekta: Hash join has a different implementation. `CPhysicalScan` uses the same logic.
Will move the logic to `CUtils`.

libgpopt/src/operators/CPhysicalInnerNLJoin.cpp
khannaekta: But we care about the OR part of the condition `0 == ulOptReq`. Not sure of the actual intent but looks like we do not want to request `ulOptReq = 1` at all when this guc is ON.
(ulOptReq = 1: requests outer child to be Replicated)

libgpopt/src/base/CUtils.cpp
```diff
@@ -1238,8 +1238,7 @@ CUtils::FPhysicalInnerJoin
 {
 	GPOS_ASSERT(NULL != pop);
 
-	return 	COperator::EopPhysicalInnerNLJoin == pop->Eopid() ||
-		COperator::EopPhysicalInnerIndexNLJoin == pop->Eopid() ||
```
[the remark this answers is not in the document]
khannaekta: The only caller of this function is `CPhysicalJoin::PdsDerive`. In `CPhysicalJoin::PdsDerive` it checks this condition and adds an equiv hashed spec set for inner distribution is used for distribution spec matching. But if equiv hashed spec set is set for `EopPhysicalInnerNLJoin`, distribution spec matching goes into an infinite loop resulting into a segfault. 

We can add it the check for `EopPhysicalInnerNLJoin` at the caller end, and add it back here to avoid confusion.

libgpopt/src/operators/CPhysicalInnerNLJoin.cpp
```diff
@@ -108,6 +111,50 @@ CPhysicalInnerNLJoin::PdsRequired
...
+
+					// create a matching hashed distribution request
+					BOOL fNullsColocated = pdshashed->FNullsColocated();
+					CDistributionSpecHashed *pdshashedEquiv = GPOS_NEW(pmp) CDistributionSpecHashed(pdrgpexprMatching, fNullsColocated);
+
+					return pdshashedEquiv;
+				}
+				pdrgpexprMatching->Release();
+			}
+		}
```
hsyuan: Why do we request redistribute for inner child to match outer instead of vice versa?
For join `t inner join pt on (t.a=pt.b)`, with both table distributed by column `a`, `pt` partitioned by `b`, will we miss the opportunity to do dynamic partition pruning?

libgpopt/src/operators/CPhysicalInnerNLJoin.cpp
dhanashreek89: Request 0: Outer child requested to be `Any`
Request 1: Outer child requested to be `Replicated`
So when `optimizer_enable_broadcast_nestloop_outer_child` (enable outer child of InnerNLJ to be replicated) is set to off; we still need to request `Any` for request 1. Hence the check.

libgpopt/src/operators/CPhysicalInnerNLJoin.cpp
khannaekta: The dynamic partition elimination plan is still an alternative. It does not get dropped.
```
set optimizer_enable_hashjoin to off;
explain select * from t inner join pt on t.a = pt.b;
                                             QUERY PLAN
 Gather Motion 3:1  (slice2; segments: 3)  (cost=0.00..1324061.80 rows=101 width=12)
   ->  Nested Loop  (cost=0.00..1324061.79 rows=34 width=12)
         Join Filter: t.a = pt.b
         ->  Partition Selector for pt (dynamic scan id: 1)  (cost=10.00..100.00 rows=34 width=4)
               Filter: pt.b = t.a
               ->  Broadcast Motion 3:3  (slice1; segments: 3)  (cost=0.00..431.00 rows=10 width=4)
                     ->  Table Scan on t  (cost=0.00..431.00 rows=4 width=4)
         ->  Dynamic Table Scan on pt (dynamic scan id: 1)  (cost=0.00..431.00 rows=34 width=8)
 Optimizer: PQO version 2.55.1
(9 rows)
```

libgpopt/src/base/CUtils.cpp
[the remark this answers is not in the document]
dhanashreek89: why does it go in the infinite loop ?
It would be helpful If we could get a reasoning for that and explain why `CPhysicalInnerNLJoin ` is an exception at L375 in `CPhysicalJoin::PdsDerive()`.

dhanashreek89:
We still have the same issue for `CPhysicalInnerIndexNLJoin` 
```
explain select * from R join S on R.a=S.a and R.b=S.b;
                                         QUERY PLAN
---------------------------------------------------------------------------------------------
 Gather Motion 3:1  (slice2; segments: 3)  (cost=0.00..437.00 rows=1 width=16)
   ->  Nested Loop  (cost=0.00..437.00 rows=1 width=16)
         Join Filter: true
         ->  Broadcast Motion 3:3  (slice1; segments: 3)  (cost=0.00..431.00 rows=1 width=8)
               ->  Table Scan on r  (cost=0.00..431.00 rows=1 width=8)
         ->  Index Scan using s_idx_b on s  (cost=0.00..6.00 rows=1 width=8)
               Index Cond: s.b = r.b
               Filter: r.a = s.a
 Optimizer: PQO version 2.55.1
(9 rows)
```

Can this be extended for `CPhysicalInnerIndexNLJoin` as well ? I am fine with tracking this as a separate PR; up to you.

libgpopt/src/operators/CPhysicalJoin.cpp
```diff
@@ -362,22 +363,26 @@ CPhysicalJoin::PdsDerive
...
-		// 'R left join S on R.a = S.a left join T on T.a = S.a'
-		// (R, S, T are distributed by a). Without equiv hashed spec,
-		// Orca will assume (R left join S) is hash distributed by R.a,
-		// ignores the facts that it is also distributed by S.a, and
-		// request a redistribute motion on S.a, which is redundant.
-		// With the equiv hashed spec set for inner distribution, Orca
-		// will know that the equivclass has the same distribution
-		// spec during distribution spec matching.
+		// (except CPhysicalInnerNLJoin) or outer join, set the outer's
+		// distribution spec as inner's equiv hashed distribution spec,
```
dhanashreek89: Why distribution matching spec goes into an infinite loop when we include `CPhysicalInnerNLJoin` in the check ?
It would be helpful If we could get a reasoning and elaborate why `CPhysicalInnerNLJoin` is an exception here.

khannaekta: Thanks for pointing it out. CPhysicalInnerNLJoin is an exception since we never initialize equivalent hashed distribution(m_pdshashedEquiv) ever for inner (`CPhysicalInnerNLJoin::PdsRequired()`). So for R InnerNLJoin S we set S's(m_pdshashedEquiv) to R's distribution spec and vice-versa for the commutative alternative. This leads to a an infinite-recursion in `CDistributionSpecHashed::FSatisfies()`. 
Added the same comment to the file too.

And regarding extended for `CPhysicalInnerIndexNLJoin`, I would suggest we track it as a separate story as the implementations differ.

PR #334, opened 2018-02-20
Avoid starting timer unnecessarily

bhuvnesh2703:
CTimerUser default constructor calls Restart() method which calls getrusage
which is a costly system call. We use the timer to print optimization
statistics if the trace flag EopttracePrintOptimizationStatistics is
set. Since the default constructor calls Restart(), even though
optimization statistics trace is off, we still pay cost to get the
default container. This patch ensures that default constructor of
CTimerUser does not call restart, if required Restart should be called
on CTimerUser object.

In ORCA, there is currently one search stage and the time out threshold
is set to ULONG_MAX which essentially means a timeout after more than a
month, so essentially no timeout is forced. So, there is no need to set
the timer in CSearchStage::RestartTimer which calls Restart() when the
time threshold is set to ULONG_MAX and similarly for
CSearchStage::FTimedOut which checks if the search stage has timed out.

After this optimization, north star query optimization time reduced to 52sec from 62sec (~16% improvement). 
There are other classes like CWallClock where we may have the similar problem as well, will keep them separate than this PR.


libgpopt/include/gpopt/search/CSearchStage.h
```diff
@@ -75,15 +75,21 @@ namespace gpopt
 			virtual
 			~CSearchStage();
 
-			// restart timer
+			// restart timer if time threshold is not default indicating don't timeout
+			// Restart() is a costly method, so avoid calling unnecessarily
 			void RestartTimer()
```
vraghavan78: This function is used in CEngine::MainThreadOptimize();

Should we also `if (fPrintOptStats)` that call?

libgpopt/include/gpopt/search/CSearchStage.h
```diff
@@ -75,15 +75,21 @@ namespace gpopt
...
 			{
-				m_timer.Restart();
+				if (m_ulTimeThreshold != ULONG_MAX)
+					m_timer.Restart();
 			}
 
 			// is search stage timed-out?
+			// if threshold is ULONG_MAX, its the default and we need not time out
```
hsyuan: change it to 
`if threshold is ULONG_MAX, never time out.`

libgpopt/include/gpopt/search/CSearchStage.h
hsyuan: restart timer only when the time threshold is not ULONG_MAX.

bhuvnesh2703:
> This function is used in CEngine::MainThreadOptimize();
> Should we also if (fPrintOptStats) that call?

CSearchStage::RestartTimer() is called in CEngine::MainThreadOptimize(), where the timer is started for the current search stage, by the time search stage is created the time threshold will be already ULONG_MAX, so when RestartTimer is called in MainThreadOptimize it will not start the timer. The purpose of this timer is to kill the threshold after a certain time threshold, so didn't do` if fprintopts`

PR #335, opened 2018-02-21
Raise exception when the aggregate operation has attributes that are …

vraghavan78:
…set returning

In GPDB we do not support aggregates that have attributes that are set
returning. For example: `explain select count(unnest(foo)) FROM tbl;`

In such cases we should stop optimizing and fallback.


vraghavan78:
Couple of things I am still considering:

1. Have a system that check before hand whether a pre-processing step is needed or not.
2. Adding the following as a test in GPDB:

```
CREATE TABLE tbl (foo int[]) DISTRIBUTED RANDOMLY;
INSERT INTO tbl VALUES (array[1,2,3]);
explain count(unnest(foo)) FROM tbl;
```

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -2261,7 +2305,12 @@ CExpressionPreprocessor::PexprPreprocess
 	GPOS_CHECK_ABORT;
 	pexrReorderedScalarCmpChildren->Release();
 
-	return pexprExistWithPredFromINSubq;
+	// (26) check if aggregate function has set returning function
+	CExpression *pexprAggrefOnSet = PexprCheckAggrefOnSetOps(pmp, pexprExistWithPredFromINSubq);
```
khannaekta: Instead of keeping it as the last preprocessing step, shouldn't we move it to the beginning and bail out way earlier. 
Or move it in GPDB DXL translation. Thoughts?

vraghavan78: ur correct we can move it up.

vraghavan78: @khannaekta on second thought ur idea of moving it to DXL to Expr translation is also a valid one

vraghavan78:
@khannaekta I moved it to DXL to Expr translation. While we can further move it to the host system but I do not think the benefit will be that much.

PR #337, opened 2018-02-22
Implement CPhysicalFilter Partition Propagtion Spec

bhuvnesh2703:
Cache the Partition Propagtion Spec generated for CPhysicalFilter.
This help to avoid recomputing of the partition propagation spec if it has already been
computed.

For north star query, total partition propagation request for CPhysicalFilter were 5668 , out of which 5462 were retrieved from cached.
This optimization brings down the optimization of north star query from 52.x sec to 47.y sec (around 9-10% improvement)


libgpopt/include/gpopt/operators/CPhysicalFilter.h
```diff
@@ -33,6 +33,91 @@ namespace gpopt
...
+			// partition propagation request generated by a filter
+			class CPartPropReq : public CRefCount
+			{
+				private:
+
+					// required part propagation
+					CPartitionPropagationSpec *m_pppsRequired;
+
+					// id of outer child
+					ULONG m_ulOuterChild;
```
hsyuan: I would prefer add `Id` to the variable name

libgpopt/src/operators/CPhysicalFilter.cpp
```diff
@@ -281,6 +284,97 @@ CPhysicalFilter::PppsRequired
...
+	ULONG ulChildIndex,
+	DrgPdp *pdrgpdpCtxt,
+	ULONG ulOptReq
+	)
+{
+	CPartPropReq *pppr = PpprCreate(pmp, exprhdl, pppsRequired);
+	if (NULL == pppr)
+	{
+		return PppsRequiredFilter(pmp, exprhdl, pppsRequired, ulChildIndex, pdrgpdpCtxt, ulOptReq);
+	}
```
hsyuan: what does it mean if `pppr` is NULL? I don't get it.

libgpopt/src/operators/CPhysicalFilter.cpp
```diff
@@ -281,6 +284,97 @@ CPhysicalFilter::PppsRequired
...
+
+	CPartitionPropagationSpec *ppps = m_phmpp->PtLookup(pppr);
+	if (NULL == ppps)
+	{
+		ppps = PppsRequiredFilter(pmp, exprhdl, pppsRequired, ulChildIndex, pdrgpdpCtxt, ulOptReq);
+#ifdef GPOS_DEBUG
+		BOOL fSuccess =
+#endif // GPOS_DEBUG
+			m_phmpp->FInsert(pppr, ppps);
+		GPOS_ASSERT(fSuccess);
```
hsyuan: The assert is useless here.

libgpopt/src/operators/CPhysicalFilter.cpp
bhuvnesh2703: PpprCreate can return NULL if exprhdl.Pgexpr() is NULL. However i don't think that should happen. I had this question myself that when exprhdl.Pgexpr() could be NULL, but couldn't find an example. So kept this as it exists at other place. But i will plan to remove it, if no such case.

libgpopt/include/gpopt/operators/CPhysicalFilter.h
bhuvnesh2703: added

libgpopt/src/operators/CPhysicalFilter.cpp
bhuvnesh2703: i removed it, i ran ctest as well didn't find any such instance.

libgpopt/include/gpopt/operators/CPhysicalFilter.h
```diff
@@ -33,6 +33,91 @@ namespace gpopt
 			// private copy ctor
 			CPhysicalFilter(const CPhysicalFilter &);
 
+		protected:
+			// partition propagation request generated by a filter
+			class CPartPropReq : public CRefCount
```
khannaekta: We have `class CPartPropReq` also defined in `CPhysicalJoin`.
Can we create a single class and reuse the code instead of duplicating it in `CPhysicalFilter` and `CPhysicalJoin` ?

bhuvnesh2703: CphysicalFilter and CphysicalJoin has different composition
- CphysicalJoin  (child 1 (outer), child 2 (inner),  childindex (0, 1), scalar child)
- CphysicalFilter (child 1 (outer), child 2(scalar child)
The class CPartPropReq could be moved down to cphysical but that will impose some issues like you will have to initialize some of the non-related variables by some number say ULONG_MAX for child index and then in the hash and equal function of this CPartPropReq you will have to handle the default values vs original values or create different hash functions for different compositions which ideally cphysical shouldn't care.
CPhysicalJoin is essentially a base class for several of the other types of joins and they could reuse the CPartPropReq, so isn't it better to keep them separate (atleast at the lowest level of base class having same composition) ?

vraghavan78:
Closing this PR to perform refactoring first.

PR #338, opened 2018-02-22
WIP: Add preprocessing step to inject limit under union

melanieplageman:
If all of the scalar project elements of a child of a logical union are scalar consts, then we can improve the query performance by adding a limit under that logical union's child.

Currently I have added this to the bottom, as I haven't thought of any benefits of moving it up (i.e. interactions with other pre-processing steps).

It should also apply to some other set operations (not the `ALL` set operations, but most of the others).

It should also be carefully thought out if this belongs as a pre-processing step, since we may have the potential to add unnecessary motion which makes worse plans in some cases.

Also, this doesn't past the ctests right now, as I haven't dealt with the plan diffs. There are also some plan space size changes.

### DIFF
--- a/libgpopt/src/operators/CExpressionPreprocessor.cpp
+++ b/libgpopt/src/operators/CExpressionPreprocessor.cpp
@@ -106,6 +106,81 @@ CExpressionPreprocessor::PexprPruneSuperfluousEquality
 	return GPOS_NEW(pmp) CExpression(pmp, pop, pdrgpexprChildren);
 }
 
+CExpression *
+CExpressionPreprocessor::PexprConvertGetToConst
+	(
+	IMemoryPool *pmp,
+	CExpression *pexpr
+	)
+{
+	if (pexpr->Pop()->Eopid() != COperator::EopLogicalUnion)
+	{
+		(pexpr->Pop())->AddRef();
+		(pexpr->PdrgPexpr())->AddRef();
+		return GPOS_NEW(pmp) CExpression(pmp, pexpr->Pop(), pexpr->PdrgPexpr());
+	}
+	// loop through the children
+	DrgPexpr *logicalUnionChildren = GPOS_NEW(pmp) DrgPexpr(pmp);
+	for (ULONG ul = 0; ul < pexpr->UlArity(); ul++)
+	{
+		CExpression *logicalProject = (*pexpr)[ul];
+		if (logicalProject->Pop()->Eopid() != COperator::EopLogicalProject)
+		{
+			logicalProject->AddRef();
+			logicalUnionChildren->Append(logicalProject);
+			continue;
+		}
+
+		CExpression *logicalGet = (*logicalProject)[0];
+		if (logicalGet->Pop()->Eopid() != COperator::EopLogicalGet)
+		{
+			logicalProject->AddRef();
+			logicalUnionChildren->Append(logicalProject);
+			continue;
+		}
+
+		CExpression *scalarProjectList = (*logicalProject)[1];
+		if (scalarProjectList->Pop()->Eopid() != COperator::EopScalarProjectList)
+		{
+			logicalProject->AddRef();
+			logicalUnionChildren->Append(logicalProject);
+			continue;
+		}
+
+		// loop through project list and return early if any project element's child is not a scalar const
+		const ULONG ulArity = scalarProjectList->UlArity();
+		for (ULONG j = 0; j < ulArity; j++)
+		{
+			CExpression *scalarProjectElem = (*scalarProjectList)[j];
+			GPOS_ASSERT(scalarProjectElem->Pop()->Eopid() == COperator::EopScalarProjectElement);
+			if ((*scalarProjectElem)[0]->Pop()->Eopid() != COperator::EopScalarConst)
+			{
+				logicalProject->AddRef();
+				logicalUnionChildren->Append(logicalProject);
+				continue;
+			}
+		}
+
+		// all children are constant so make a logical limit
+		logicalGet->Pop()->AddRef();
+		logicalGet->PdrgPexpr()->AddRef();
+		CExpression *pexprLogicalGet = GPOS_NEW(pmp) CExpression(pmp, logicalGet->Pop(), logicalGet->PdrgPexpr());
+
+		CExpression *logicalLimit = CUtils::PexprLimit(pmp, pexprLogicalGet, 0, 1);
+
+		scalarProjectList->Pop()->AddRef();
+		scalarProjectList->PdrgPexpr()->AddRef();
+		CExpression *pexprScalarProjectList = GPOS_NEW(pmp) CExpression(pmp, scalarProjectList->Pop(), scalarProjectList->PdrgPexpr());
+
+		CExpression *logicalProjectNewExpr = CUtils::PexprLogicalProject(pmp, logicalLimit, pexprScalarProjectList, false);
+		logicalUnionChildren->Append(logicalProjectNewExpr);
+	}
+
+	pexpr->Pop()->AddRef();
+	CExpression *resultExpr = GPOS_NEW(pmp) CExpression(pmp, pexpr->Pop(), logicalUnionChildren);
+	return resultExpr;
+}
+
 // an existential subquery whose inner expression is a GbAgg
 // with no grouping columns is replaced with a Boolean constant
 //
@@ -2261,7 +2336,11 @@ CExpressionPreprocessor::PexprPreprocess
 	GPOS_CHECK_ABORT;
 	pexrReorderedScalarCmpChildren->Release();
 
-	return pexprExistWithPredFromINSubq;
+	CExpression *pexprUnionGetToConst = PexprConvertGetToConst(pmp, pexprExistWithPredFromINSubq);
+	GPOS_CHECK_ABORT;
+	pexprExistWithPredFromINSubq->Release();
+
+	return pexprUnionGetToConst;
 }
 
 // EOF
--- a/libgpopt/src/operators/CExpressionPreprocessor.cpp
+++ b/libgpopt/src/operators/CExpressionPreprocessor.cpp
@@ -113,6 +113,8 @@ CExpressionPreprocessor::PexprConvertGetToConst
 	CExpression *pexpr
 	)
 {
+	// TODO: handle unions under other stuff
+	// TODO: handle other set operations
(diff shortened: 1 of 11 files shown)


dhanashreek89:
DDL
```
create table foo(a int, b int);
create table bar(a int, b int);
insert into foo select i, i from generate_series(1,20000)i;
insert into bar select i, i from generate_series(1,300000)i;
```

1. Plan before the fix:
```
explain select * from ((select 1 a, 2 b from foo) union (select bar.a, bar.b from bar)) tt limit 10;
                                                   QUERY PLAN
-----------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.00..922.18 rows=4 width=8)
   ->  Gather Motion 3:1  (slice2; segments: 3)  (cost=0.00..922.18 rows=10 width=8)
         ->  Limit  (cost=0.00..922.18 rows=4 width=8)
               ->  HashAggregate  (cost=0.00..922.18 rows=107263 width=8)
                     Group Key: "outer".a, "outer".b
                     ->  Redistribute Motion 3:3  (slice1; segments: 3)  (cost=0.00..895.76 rows=107263 width=8)
                           Hash Key: "outer".a, "outer".b
                           ->  HashAggregate  (cost=0.00..893.08 rows=107263 width=8)
                                 Group Key: "outer".a, "outer".b
                                 ->  Append  (cost=0.00..866.66 rows=107263 width=8)
                                       ->  Result  (cost=0.00..431.21 rows=6667 width=8)
                                             ->  Table Scan on foo  (cost=0.00..431.14 rows=6667 width=1)
                                       ->  Table Scan on bar  (cost=0.00..433.10 rows=100596 width=8)
 Optimizer: PQO version 2.55.2
(14 rows)
```

2. Plan after the fix (Limit added over foo by the new pre-processing step):
```
explain select * from ((select 1 a, 2 b from foo) union (select bar.a, bar.b from bar)) tt limit 10;
                                                             QUERY PLAN
------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.00..935.55 rows=4 width=8)
   ->  Gather Motion 3:1  (slice5; segments: 3)  (cost=0.00..935.55 rows=10 width=8)
         ->  Limit  (cost=0.00..935.55 rows=4 width=8)
               ->  HashAggregate  (cost=0.00..935.55 rows=100597 width=8)
                     Group Key: "outer".a, "outer".b
                     ->  Redistribute Motion 3:3  (slice4; segments: 3)  (cost=0.00..910.78 rows=100597 width=8)
                           Hash Key: "outer".a, "outer".b
                           ->  HashAggregate  (cost=0.00..908.26 rows=100597 width=8)
                                 Group Key: "outer".a, "outer".b
                                 ->  Redistribute Motion 1:3  (slice3)  (cost=0.00..883.49 rows=301789 width=8)
                                       ->  Append  (cost=0.00..877.22 rows=100597 width=8)
                                             ->  Limit  (cost=0.00..431.21 rows=1 width=8)
                                                   ->  Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..431.21 rows=1 width=8)
                                                         ->  Limit  (cost=0.00..431.21 rows=1 width=8)
                                                               ->  Result  (cost=0.00..431.21 rows=6667 width=8)
                                                                     ->  Table Scan on foo  (cost=0.00..431.14 rows=6667 width=1)
                                             ->  Gather Motion 3:1  (slice2; segments: 3)  (cost=0.00..443.60 rows=301788 width=8)
                                                   ->  Table Scan on bar  (cost=0.00..433.10 rows=100596 width=8)
 Optimizer: PQO version 2.55.2
(19 rows)
```

**Performance numbers on the retail build:
Total time with Plan 1 (before the fix):  78.879 ms
Total time with Plan 2 (after the fix)  : 318.060 ms**

Optimization time is almost the same before and after the fix.

The reason why we think that this could be a transformation is because of the Plan 2 generated above where the first child if append is limit over foo, which happens on master and then to perform the append operation, bar is gathered on master. However in Plan 1, append happens locally on all the segments without moving foo and bar. The data is only moved around for de-duplication since it is UNION.

**Legacy planner** on the other hand, does not perform the append operation on master if we explicitly add a limit over foo in the query:
```
explain select * from ((select 1 a, 2 b from foo limit 1) union (select bar.a, bar.b from bar)) tt limit 10;
                                                            QUERY PLAN
----------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=7876.75..7877.15 rows=10 width=8)
   ->  Gather Motion 3:1  (slice4; segments: 3)  (cost=7876.75..7877.15 rows=10 width=8)
         ->  Limit  (cost=7876.75..7876.95 rows=4 width=8)
               ->  HashAggregate  (cost=7876.75..10894.64 rows=100597 width=8)
                     Group Key: "outer".a, "outer".b
                     ->  Redistribute Motion 3:3  (slice3; segments: 3)  (cost=0.00..6367.80 rows=100597 width=8)
                           Hash Key: "outer".a, "outer".b
                           ->  Append  (cost=0.00..6367.80 rows=100597 width=8)
                                 ->  Redistribute Motion 1:3  (slice2; segments: 1)  (cost=0.00..0.04 rows=1 width=0)
                                       Hash Key: "outer".a, "outer".b
                                       ->  Subquery Scan on "*SELECT* 1"  (cost=0.00..0.04 rows=1 width=0)
                                                   ->  Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..0.03 rows=1 width=0)
                                                               ->  Seq Scan on foo  (cost=0.00..224.00 rows=6667 width=0)
                                 ->  Seq Scan on bar  (cost=0.00..3349.88 rows=100596 width=8)
 Optimizer: legacy query optimizer
(17 rows)
```
**Run time with legacy planner : 65.721 ms**


Hence would like to gather thoughts on whether we should add it as a transformation rather than a pre-processing step such that the cost-model will make the final decision ?

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -106,6 +106,70 @@ CExpressionPreprocessor::PexprPruneSuperfluousEquality
 	return GPOS_NEW(pmp) CExpression(pmp, pop, pdrgpexprChildren);
 }
 
+// helper function to determine if all scalar project elements are scalar consts
+BOOL CExpressionPreprocessor::CheckScalarProjectAllConst
```
hsyuan: This should be moved to CUtils.cpp

hsyuan:
"Total time with Plan 1 (before the fix): 78.879 ms
Total time with Plan 2 (after the fix) : 318.060 ms"

Is it overall running time? Isn't this a regression?

hsyuan:
You don't need a transformation, which is an overkill. All you want is put a limit on top of table scan foo, maybe after the plan is generated instead of preprocess step. We know that a going to be a good plan.

hsyuan:
So the expected plan should be:
```sql
explain select * from ((select 1 a, 2 b from foo) union (select bar.a, bar.b from bar)) tt limit 10;
                                                   QUERY PLAN
-----------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.00..922.18 rows=4 width=8)
   ->  Gather Motion 3:1  (slice2; segments: 3)  (cost=0.00..922.18 rows=10 width=8)
         ->  Limit  (cost=0.00..922.18 rows=4 width=8)
               ->  HashAggregate  (cost=0.00..922.18 rows=107263 width=8)
                     Group Key: "outer".a, "outer".b
                     ->  Redistribute Motion 3:3  (slice1; segments: 3)  (cost=0.00..895.76 rows=107263 width=8)
                           Hash Key: "outer".a, "outer".b
                           ->  HashAggregate  (cost=0.00..893.08 rows=107263 width=8)
                                 Group Key: "outer".a, "outer".b
                                 ->  Append  (cost=0.00..866.66 rows=107263 width=8)
                                       ->  Result  (cost=0.00..431.21 rows=6667 width=8)
                                                   ->  Seq Scan on foo  (cost=0.00..224.00 rows=6667 width=0)
                                       ->  Table Scan on bar  (cost=0.00..433.10 rows=100596 width=8)
 Optimizer: PQO version 2.55.2
(14 rows)
```

vraghavan78:
@hsyuan the plan u are stating is a local limit. More I think of this idea, I believe is too nuance to generalize into either a preprocessing step or a transfer.

dhanashreek89:
The expected plan should be similar to the one generated by planner where we redistribute the limit result.
```
explain select * from ((select 1 a, 2 b from foo limit 1) union (select bar.a, bar.b from bar)) tt limit 10;
                                                            QUERY PLAN
----------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=7876.75..7877.15 rows=10 width=8)
   ->  Gather Motion 3:1  (slice4; segments: 3)  (cost=7876.75..7877.15 rows=10 width=8)
         ->  Limit  (cost=7876.75..7876.95 rows=4 width=8)
               ->  HashAggregate  (cost=7876.75..10894.64 rows=100597 width=8)
                     Group Key: "outer".a, "outer".b
                     ->  Redistribute Motion 3:3  (slice3; segments: 3)  (cost=0.00..6367.80 rows=100597 width=8)
                           Hash Key: "outer".a, "outer".b
                           ->  Append  (cost=0.00..6367.80 rows=100597 width=8)
                                 ->  Redistribute Motion 1:3  (slice2; segments: 1)  (cost=0.00..0.04 rows=1 width=0)
                                       Hash Key: "outer".a, "outer".b
                                       ->  Subquery Scan on "*SELECT* 1"  (cost=0.00..0.04 rows=1 width=0)
                                                   ->  Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..0.03 rows=1 width=0)
                                                               ->  Seq Scan on foo  (cost=0.00..224.00 rows=6667 width=0)
                                 ->  Seq Scan on bar  (cost=0.00..3349.88 rows=100596 width=8)
 Optimizer: legacy query optimizer
```

hsyuan:
We don't need to be exact the same with planner plan, even though gather and redistribute will give us only 1 tuple, but we have to start 2 additional slices. If the plan tree is very large, every slice will consume a lot memory, if it only executes part of the plan tree. Additional motion will also add overhead for process communication and tuple transfer. If we have local limit only, even it will give us the same number of duplicate tuples with the number of segments, but since every tuple is process in a single slice in memory, and the number of segments is at most hundreds, it is going to be better plan.

hsyuan:
Oh, but how to put the local limit is a big problem. OK, forget about what I was saying.

melanieplageman:
Closing due to concerns enumerated in comments.

PR #339, opened 2018-02-23
Create GPORCA PR pipeline to run ctests

khannaekta:
This pipeline runs ctests in release build.

### DIFF
--- a/concourse/pr_pipeline.yml
+++ b/concourse/pr_pipeline.yml
@@ -1,0 +1,83 @@
+resource_types:
+  - name: pull_request
+    type: docker-image
+    source:
+      repository: jtarchie/pr
+
+resources:
+  - name: orca_pr
+    type: pull_request
+    source:
+      repo: greenplum-db/gporca
+      access_token: {{gporcabot_access_token}}
+      ignore_paths:
+      - README.md
+      - LICENSE
+      - COPYRIGHT
+      - .gitignore
+
+  - name: xerces_patch
+    type: git
+    source:
+      branch: master
+      paths:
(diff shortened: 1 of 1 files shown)


concourse/pr_pipeline.yml
```diff
@@ -0,0 +1,86 @@
...
+    - task: package_tarball
+      file: xerces_patch/concourse/xerces-c/package_tarball_centos5.yml
+  - aggregate:
+    - put: bin_xerces_centos5
+      params:
+        from: package_tarball_centos5/bin_xerces_centos5.tar.gz
+
+- name: compile_and_test_gporca
+  public: true
+  max_in_flight: 10
```
d: consider lowering the `max_in_flight` to 2?
The reasoning goes like
1. While I'll admit this will be a good problem to have, running 10 `ctest -j8` concurrently will create a spike of load on our server
1. `ctest` will try to "nicely" back off if the system is heavily loaded (say the load average is 16 on an 8-core system), but it takes a while to realize that fact.
1. Again this is more a problem of taste: I'd rather not see wildly varying test run time, and have consistency while we can.

concourse/pr_pipeline.yml
```diff
@@ -0,0 +1,86 @@
+## ======================================================================
+## resources
```
d: Consider removing these comments and the beautiful `===` banners. They don't seem to convey any more information than what the code tells a reader of this pipeline configuration

khannaekta:
@d Review comments addressed. Could you please take a look for any additional comments.

PR #341, opened 2018-02-24
HashJoin rewindability should be identical to its outer child

asubramanya:
Its inner child need not be rewindable for a hash join to be rewindable, unlike
a nested loop join. The logic to compute the derived and required properties
were out of sync.  Specifically the logic to derive the rewindablility for a
hash join erraneously reported unrewindable when the inner child was not
rewindable, even though the outer child was. This wasn't wrong logically, but
it led to extraneous physical spool nodes in the plan, which might cause other
issues for us. One such example is that the materialize might end up in between
a partition selector and a Dynamic table scan leading to wrong results.

This commit fixes that issue by simply returning the outer child's
rewindability while deriving the rewindability.

[#154605891]
Signed-off-by: Abhijit Subramanya <asubramanya@pivotal.io>

### DIFF
--- a/data/dxl/minidump/AddPredsInSubqueries.mdp
+++ b/data/dxl/minidump/AddPredsInSubqueries.mdp
@@ -334,7 +334,7 @@
         </dxl:LogicalGet>
       </dxl:LogicalSelect>
     </dxl:Query>
-    <dxl:Plan Id="0" SpaceSize="414">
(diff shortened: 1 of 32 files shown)


hsyuan:
If the hash side of hashjoin is a motion, no matter the rewindability of outer child, the hashjoin is still not rewindable.

vraghavan78:
Need to re-think this.

PR #343, opened 2018-02-28
Pass optimizer_nestloop_factor to DXL through optimizer config

sambitesh:
The default value of GUC `optimizer_nestloop_factor` in GPDB is 1024. The GUC
is also used in Orca (default value being 1), but it is not passed through to DXL.
This commit fixes this issue.

Now we are accounting for `optimizer_nestloop_factor` while costing  NLJoin (CCostModelGPDB::CostNLJoin).

Earlier we were accounting for it while translating back from DXL in GPDB side in `COptTasks::SetCostModelParams`

Will Bump ORCA version before merging. 


libgpdbcost/src/CCostModelGPDB.cpp
```diff
@@ -1059,6 +1059,10 @@ CCostModelGPDB::CostNLJoin
 	CColRefSet *pcrsUsed = CDrvdPropScalar::Pdpscalar(pexprJoinCond->PdpDerive())->PcrsUsed();
 	const ULONG ulColsUsed = pcrsUsed->CElements();
 
+	// penalize Nested Loop Join by optimizer_nestloop_factor GUC
```
vraghavan78: change comment. fetch the penalization factor for NLJ

vraghavan78:
Can you add a minidump test?

libgpopt/include/gpopt/engine/CHint.h
```diff
@@ -60,6 +63,7 @@ namespace gpopt
 				ULONG ulArrayExpansionThreshold,
 				ULONG ulJoinOrderDPLimit,
 				ULONG ulBroadcastThreshold,
+				ULONG ulNestloopFactor,
 				BOOL fEnforceConstraintsOnDML
 				)
```
hsyuan: I am afraid the CHint class will explode with lots of input arguments in the future. I think we should refactor the constructor to be 0 input arguments with every member to be the default value, at the same time provide getter and setter methods to read/write the value in the GPDB translator side.

server/CMakeLists.txt
```diff
@@ -195,7 +195,7 @@ CEquivClassTest:
 Equiv-HashedDistr-1 Equiv-HashedDistr-2 EquivClassesAndOr EquivClassesLimit IndexNLJoin_Cast_NoMotion;
 
 COuterJoinTest:
-LOJ-Condition-False FOJ-Condition-False FullOuterJoin FullOuterJoin2
+LOJ-Condition-False NestloopFactorGUC FOJ-Condition-False FullOuterJoin FullOuterJoin2
```
khannaekta: NestloopFactorGUC.mdp file missing

khannaekta: On a second thought, we should have a parseHandler test for this instead of a minidump test. Since in the minidump, we only test the plan and this NestLoop factor is not part of the plan.

sambitesh:
I talked to @khannaekta  and we agreed upon adding Parse handler test for OptimizerConfig later since it's not straightforward. SO merging this PR.

sambitesh:
@d For point 1 did you mean to move the nestloop factor from `Chint` to `CostModelConfig` ?

PR #348, opened 2018-03-07
Extract Constraint under Negation correctly

bhuvnesh2703:
Extract Constraint under Negation correctly

While deriving the constraint properties for Aggregates, we attempt to
extract the constraint only on the columns which are used in the
grouping column.  However, in case of negation constraint we cannot
further simplify the constraint.

For instance if the query is of the form:
explain select distinct a from t1 where not (a = b);

For the predicate not (a = b), the constraint is of the form:
NOT ({"a" (0), ranges: (-inf, inf) } AND {"b" (1), ranges: (-inf, inf) }))

While deriving constraint for Agg as the grouping / distinct column is
a, we cannot break the constraint to NOT ({"a" (0), ranges: (-inf, inf) }) as that is equivalent to NOT (a)
Currently, the implementation recurses down the constraints and returns
 NOT ({"a" (0), ranges: (-inf, inf) }) resulting in wrong results.

The returned constraint is further used down during calculation of max
cardinality, which may falsely report that the aggregation will not
return any results and convert the query to a constant table get with a
false filter.

Consider the below case for wrong results:

```sql
create table t1 (a int, b int);
insert into t1 values (1, 1);
insert into t1 values (1, 2);
explain select distinct a from t2 where a is not null and not a = b;
                QUERY PLAN
------------------------------------------
   One-Time Filter: false
 Optimizer: PQO version 2.55.4
```
Input to Expression Preprocessor:

```
Algebrized query:
+--CLogicalGbAgg( Global )
   |--CLogicalSelect
   |  |--CLogicalGet "t1"
   |  +--CScalarBoolOp (EboolopAnd)
   |     |--CScalarBoolOp (EboolopNot)
   |     |  +--CScalarNullTest
   |     |     +--CScalarIdent "a"
   |     +--CScalarBoolOp (EboolopNot)
   |        +--CScalarCmp (=)
   |           |--CScalarIdent "a"
   |           +--CScalarIdent "b"
   +--CScalarProjectList
```

Output of Expression Preprocessor:
```
+--CLogicalConstTableGet Columns: ["a" (0)] Values: []
```

As we see above, preprocessing converts the input query to a constant
table get with false filter which shouldn't be done.
1. During preprocessing, we derive the Property Constraint of
CLogicalGbAgg by calling `CLogicalGbAgg::PpcDeriveConstraint` ->
`PpcDeriveConstraintRestrict`
2. PpcDeriveConstraintRestrict derives constraint only on the input
column set. In case of GbAgg, it derives the constraint on the grouping
column.

```
CLogical::PpcDeriveConstraintRestrict
        (
        IMemoryPool *pmp,
        CExpressionHandle &exprhdl,
        CColRefSet *pcrsOutput ==> grouping column
        )
```
3. In the repro query, a is the grouping column (i.e the distinct
clause), so it tries to determine the constraint on `column a` and
returns the constraint as `NOT (a)`

```
   |     +--CScalarBoolOp (EboolopNot)
   |        +--CScalarCmp (=)
   |           |--CScalarIdent "a" (0)
   |           +--CScalarIdent "b" (1)
```
4. The constraint derived is as : `(NOT {"a" (0), ranges: (-inf, inf)})`

5. The other constraint for a is not NULL Is: `{"a" (0), ranges: (-inf, inf)})`

6. Conjunction of the 2 constraints above is equivalent to an expression which is equivalent to `a is not NULL and a is NULL`.
```
+--CScalarBoolOp (EboolopAnd)
   |--CScalarBoolOp (EboolopNot)
   |  +--CScalarNullTest
   |     +--CScalarIdent "a" (0)
   +--CScalarNullTest
      +--CScalarIdent "a" (0)
```

and the resulting combined constraint becomes `({"a" (0), ranges: })",`

7. Then while deriving the MaxCard property for the global agg, if the
operator produces no rows then MaxCard = 0

```sql
CLogicalGbAgg::Maxcard
..
..
..
        // contradictions produce no rows
        if (CDrvdPropRelational::Pdprel(exprhdl.Pdp())->Ppc()->FContradiction())
        {
                return CMaxCard(0 /*ull*/);
        }
```
8. Since the MaxCard returns 0 rows, `PexprPruneEmptySubtrees(pmp,
pexprWithPreds);` step in expression preprocessing prunes the tree and
replaces that with a const table get with false filter

9. Thus, wrong results.

This PR fixes the issue and returns the complete constraint on the
column a if there is a negation, i.e
NOT ({"a" (0), ranges: (-inf, inf) } AND {"b" (1), ranges: (-inf, inf) }))


libgpopt/src/base/CConstraintNegation.cpp
```diff
@@ -94,7 +94,15 @@ CConstraintNegation::Pcnstr
...
-	return GPOS_NEW(pmp) CConstraintNegation(pmp, m_pcnstr->Pcnstr(pmp, pcr));
+	// donot recurse down the constraint, return the complete constraint
+	// on a given column which may include other columns as well
+	// in case of NOT (negation) we cannot further simplify the constraint.
+	// for instance, conjunction constraint (NOT a=b) is like:
+	//	 NOT ({"a" (0), ranges: (-inf, inf) } AND {"b" (1), ranges: (-inf, inf) }))
+	// recursing down the constraint will give NOT ({"a" (0), ranges: (-inf, inf) })
+	// but that is equivalent to (NOT a) which is not the case.
+	m_pcnstr->AddRef();
+	return GPOS_NEW(pmp) CConstraintNegation(pmp, m_pcnstr);
```
vraghavan78: this seems fishy. I ask for "tofu", and I get "! (String bean and tofu)".

Instead, should we say no constraints here that only apply to "tofu"?

bhuvnesh2703: Consider the case:
```sql
create table t4 (a bool, b bool);
insert into t4 values (false, true);
select distinct a from t4 where a is not null and not (a = b or a);
```
The above should yield 1 record, i.e `false`

So, trying to see what point are you suggesting
1. If we just consider `a is not null` constraint and exclude not (a = b or a) completely ==> This should work.  (Here NOT Constraint is a DISJUNCTION of childs a=b and a)
2. If we consider `a is not null constraint` and `not(a)` ==> This should have the same issue as described in this story.

vraghavan78: I am saying option 1.

vraghavan78: may be i am confused by the AND in the constraint

libgpopt/src/base/CConstraintNegation.cpp
```diff
@@ -94,7 +94,15 @@ CConstraintNegation::Pcnstr
 		return NULL;
 	}
 
-	return GPOS_NEW(pmp) CConstraintNegation(pmp, m_pcnstr->Pcnstr(pmp, pcr));
+	// donot recurse down the constraint, return the complete constraint
+	// on a given column which may include other columns as well
+	// in case of NOT (negation) we cannot further simplify the constraint.
+	// for instance, conjunction constraint (NOT a=b) is like:
+	//	 NOT ({"a" (0), ranges: (-inf, inf) } AND {"b" (1), ranges: (-inf, inf) }))
```
hsyuan: What? I didn't know why Orca represent it like this. What is the benefit of doing this?

bhuvnesh2703:
I believe lets consider at this moment that the representation `NOT {"a" (0), ranges: (-inf, inf) } AND {"b" (1), ranges: (-inf, inf) }` is not appropriate/intuitive and does not represent `NOT (a = b)`. 
However the use of of property constraint is to identify contradiction and if you can safely prune the tree.
If we represent `NOT (a = b)` as `NOT {"a" (0), ranges: (-inf, inf) } = {"b" (1), ranges: (-inf, inf) }` still we need to identify if we can use it to prune the tree, and hence  we cannot deduce it to `NOT {"a" (0), ranges: (-inf, inf) } ` as it loses the context NOT (a = b). I think the representation is an orthogonal issue.

hardikar:
Yeah, seems like we don't really care that `(a=b)` just that individually both `a` and `b` can span from `(-inf, +inf)`. So, in some sense, this is correct. Because the "NOT" needs to apply on the entire sub-expresssion. 

But I wanted to see if this logical applies to other constraint types i.e CConstraintConjuction and CConstraintDisjunction: 

## CConstraintConjunction (AND):

This will return all the conjuncts that contain the `pcr`. That is, a call to Pcnstr("a=b", a) will return only return `a -inf, inf` and will not return b. This is fine as long as each of the conjuncts are treated equally — especially in the case of NOT (a=b) that this PR fixes. I couldn’t find another counter-example where this gives wrong results at the present.

## CConstraintDisjunction (OR):
I see that CConstraintDisjunction does exactly what the this PR is trying to fix. Take the following example : 
```
shardikar=# explain select distinct a from t1 where a is null or (a=b);
                                  QUERY PLAN
------------------------------------------------------------------------------
 Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..431.00 rows=1 width=4)
   ->  GroupAggregate  (cost=0.00..431.00 rows=1 width=4)
         Group Key: a
         ->  Sort  (cost=0.00..431.00 rows=1 width=4)
               Sort Key: a
               ->  Table Scan on t1  (cost=0.00..431.00 rows=1 width=4)
                     Filter: a IS NULL OR a = b
 Optimizer: PQO version 2.55.5
(8 rows)
```
Debugging I see that `a IS NULL OR a = b` is converted to `a is null or (a is not null -inf, + inf   AND b is not null -inf, inf)` which reduces to `(a is null -inf, +inf)` a.k.a `true`. *So how we not get into trouble here? Because ORCA doesn’t reduce it to true*. ORCA apparently only converts contradictions to `false`:

```
                                  QUERY PLAN
------------------------------------------------------------------------------
 Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..431.00 rows=1 width=4)
   ->  GroupAggregate  (cost=0.00..431.00 rows=1 width=4)
         Group Key: a
         ->  Sort  (cost=0.00..431.00 rows=1 width=4)
               Sort Key: a
               ->  Table Scan on t1  (cost=0.00..431.00 rows=1 width=4)
                     Filter: a IS NULL OR (NOT a IS NULL)
 Optimizer: PQO version 2.55.5
```

I noticed some other opportunities that ORCA misses and I will created stories for those. But this PR LGTM.

PR #349, opened 2018-03-07
Error out when Metadata Names contain unicode characters incompatible w.r.t system's

dhanashreek89:
locale

All the metadata objects in ORCA are serialized and deserialized as wide
character strings (CWStringDynamic). The metadata name which is in char*
format; needs to be first converted into a wide character format before
it can be serialized. To perform this function, ORCA uses vswprintf()
library function. This function is sensitive to locale LC_CTYPE. On
Linux platforms, when the locale of the host system is set to "C" and
the metadata name contains unicode string, vswprintf() fails with return
value as -1. ORCA assumes that the failure happen due to insufficient
size of destination buffer and thus performs multiple retries with
increased buffer size. The function fails at each attempt due to
incorrect locale. Eventually ORCA crashes when it runs out of memory.

Repro:
```
gpinitsystem -a -c  clusterConfigFile --locale=C --lc-collate=C --lc-ctype=C --lc-messages=C --lc-monetary=C --lc-numeric=C --lc-time=C
set optimizer=on;
SELECT 1 AS "한글";  // crashes
```

This commit checks the error code when vswprintf() fails and detects the
locale incompatibility. We fall back to planner in such cases.


libgpos/src/common/clibwrapper.cpp
```diff
@@ -486,7 +486,15 @@ gpos::clib::IVswPrintf
 	GPOS_ASSERT(NULL != wszStr);
 	GPOS_ASSERT(NULL != wszFormat);
 
-	return vswprintf(wszStr, ulMaxLen, wszFormat, vaArgs);
+	INT iRes = vswprintf(wszStr, ulMaxLen, wszFormat, vaArgs);
+	if (-1 == iRes && EILSEQ == errno)
```
bhuvnesh2703: are there other error codes which represent failure? In other words, should we check if the return code is successful else for any other error code bail out?

libgpos/server/src/unittest/gpos/string/CWStringTest.cpp
```diff
@@ -181,6 +184,52 @@ CWStringTest::EresUnittest_AppendFormat()
...
+	CWStringDynamic *pstr1 = GPOS_NEW(pmp) CWStringDynamic(pmp);
+
+	GPOS_RESULT eres = GPOS_OK;
+
+	setlocale(LC_CTYPE, "C");
+	GPOS_TRY
+	{
+		pstr1->AppendFormat(GPOS_WSZ_LIT("%s"), (CHAR *)"ÃË", 123);
+
+		eres = GPOS_FAILED;
```
bhuvnesh2703: what happens if the AppendFormat call is successful, does this test fail?

libgpos/src/common/clibwrapper.cpp
dhanashreek89: The man page does not specify any other error number that we can use here for sure.

libgpos/server/src/unittest/gpos/string/CWStringTest.cpp
dhanashreek89: Yes the test fails. For ex, if we run this test without `#ifndef GPOS_DARWIN` on Mac, the test fails because `AppendFormat` is successful.

PR #351, opened 2018-03-10
Avoid placing partition selector outside spool

d:
TODO:
- [x] Have a better title
- [x] Have a better description of this PR
- [x] Comment on `CEnfdDistribution::Epet`
- [x] Comment on `CPhysicalSpool::FValidContext`

## Avoid placing partition selectors outside spool
Not exactly, but I needed a catchy and sufficiently accurate (60%) title.

## Context
ORCA enforces rewindability by adding a `Spool` operator in plans. Spool has two properties:

1. It materializes an otherwise non-rescannable sub plan-tree, making it suitable for a rewindable requirement, e.g. the second (relational) child of a nested loop join, or the second relational child of a merge join.
2. As a corollary of 1, A `Spool` operator effectively "blocks" any partition selection that happens outside of the spool sub-tree from properly propagating into the respective dynamic scans under the spool.

## Example
```
CREATE TEMP TABLE foo (a int, b int);
CREATE TEMP TABLE bar (c int, d int);
DROP TABLE IF EXISTS pt;
CREATE TABLE pt (dk int, partkey int) PARTITION BY range(partkey) (START(0) END(9) EVERY(3));
INSERT INTO pt SELECT i, i%9 FROM generate_series(1, 32) i;
INSERT INTO foo SELECT i%3 + 7, i FROM generate_series(1, 4) i;
INSERT INTO bar SELECT i, i FROM generate_series(1, 8) i;
ANALYZE;

EXPLAIN SELECT * FROM foo INNER JOIN (pt LEFT OUTER JOIN bar ON d = dk) ON a < partkey;
```

We want to avoid plans like the following:
```
+--CPhysicalMotionGather(master)
   +--CPhysicalInnerNLJoin
      |--CPhysicalPartitionSelector
      |  +--CPhysicalMotionBroadcast
      |     +--CPhysicalTableScan "foo" ("foo")
      |--CPhysicalSpool
      |  +--CPhysicalLeftOuterHashJoin
      |     |--CPhysicalDynamicTableScan "pt" ("pt")
      |     |--CPhysicalMotionHashDistribute
      |     |  +--CPhysicalTableScan "bar" ("bar")
      |     +--CScalarCmp (=)
      |        |--CScalarIdent "d" (19)
      |        +--CScalarIdent "dk" (9)
      +--CScalarCmp (<)
         |--CScalarIdent "a" (0)
         +--CScalarIdent "partkey" (10)
```

Because the partition selection above happens dynamically, it's intended to influence the relation delivered by the dynamic scan (`CPhysicalDynamicTableScan "pt"`) and hence the join result (`CPhysicalLeftOuterHashJoin`). However, the `Spool` separating `CPhysicalPartitionSelector` operator from the `CPhysicalDynamicTableScan` breaks that intent: the first scan of "pt" is materialized, voiding any future partition propagation done on rescan. The above plan is invalid: it gives wrong results.

Instead we expect a plan more like the following:

```
+--CPhysicalMotionGather(master)
   +--CPhysicalInnerNLJoin
      |--CPhysicalMotionBroadcast
      |  +--CPhysicalTableScan "foo" ("foo")
      |--CPhysicalSpool
      |  +--CPhysicalLeftOuterHashJoin
      |     |--CPhysicalPartitionSelector
      |     |  +--CPhysicalDynamicTableScan "pt" ("pt")
      |     |--CPhysicalMotionHashDistribute
      |     |  +--CPhysicalTableScan "bar" ("bar")
      |     +--CScalarCmp (=)
      |        |--CScalarIdent "d" (19)
      |        +--CScalarIdent "dk" (9)
      +--CScalarCmp (<)
         |--CScalarIdent "a" (0)
         +--CScalarIdent "partkey" (10)
```

. Notice that the partition selector is in a different place than the first plan? The above plan is valid because both the partition selector and the dynamic scan are both under the same `Spool`. Note that a plan where the dynamic scan is not placed under a `Spool` will also be acceptable, but for illustration purpose I picked the more apt example above.


libgpopt/src/operators/CPhysicalSpool.cpp
```diff
@@ -385,5 +385,26 @@ CPhysicalSpool::EpetRewindability
...
+	GPOS_ASSERT(1 == pdrgpocChild->UlLength());
+
+	COptimizationContext *pocChild = (*pdrgpocChild)[0];
+	CCostContext *pccBest = pocChild->PccBest();
+	GPOS_ASSERT(NULL != pccBest);
+
+	CDrvdPropPlan *pdpplanChild = pccBest->Pdpplan();
+	if (pdpplanChild->Ppim()->FContainsUnresolved())
+	{
+		return false;
```
dhanashreek89: a comment here would be helpful. For ex:
`prohibit adding a Spool on top of unresolved partition consumers`

libgpopt/src/operators/CPhysicalMotion.cpp
```diff
@@ -51,15 +51,6 @@ CPhysicalMotion::FValidContext
 		return false;
 	}
 
-	CExpressionHandle exprhdl(pmp);
```
bhuvnesh2703: good move to `CEnfdDistribution::Epet`, better to validate early than here.

libgpopt/src/operators/CPhysicalSpool.cpp
bhuvnesh2703: question, we cannot move it to AppendEnforcers like this PR does for Motion as some of the other group expression in a group may satisfy this, so we don't want to completely ban spool?

d:
@bhuvnesh2703 Can you help point me to where the formatting anomaly is?

libgpopt/src/operators/CPhysicalSpool.cpp
```diff
@@ -385,5 +385,26 @@ CPhysicalSpool::EpetRewindability
 }
 
 
+BOOL
+CPhysicalSpool::FValidContext(
```
bchaudhary: Here (

d: will do!

d: just done

libgpopt/src/operators/CPhysicalSpool.cpp
d: just pushed two new commits that add comments on `CPhysicalSpool::FValidContext` and `CEnfdDistribution::Epet`

d:
@dhanashreek89 and @bhuvnesh2703 check out the new comments!

bhuvnesh2703:
LGTM

libgpopt/src/operators/CPhysicalMotion.cpp
```diff
@@ -32,7 +32,7 @@ using namespace gpopt;
 BOOL
 CPhysicalMotion::FValidContext
 	(
-	IMemoryPool *pmp,
+	IMemoryPool *pmp __attribute__((unused)),
```
oarap: We have unused parameters everywhere and this is a new way of doing it. I would request a chore to apply this convention in the entire code base.

libgpopt/src/operators/CPhysicalSpool.cpp
```diff
@@ -385,5 +385,56 @@ CPhysicalSpool::EpetRewindability
 }
 
 
+BOOL
+CPhysicalSpool::FValidContext
+	(
+	IMemoryPool *,
```
oarap: ` __attribute__((unused))`

libgpopt/src/operators/CPhysicalMotion.cpp
d: I don't have any strong opinions to the attribute here.

On a tangential note, I *do* have a strong opinion that all functions in our header files should have _named_ arguments. There are two ways (out of scope for this PR) to achieve that:

1. attaching the "unused" attribute to arguments when the definition of the function is in the header
1. just move trivial (to the point of not using the arguments) definitions of methods out of header files

libgpopt/src/operators/CPhysicalSpool.cpp
d: ?

libgpopt/src/operators/CPhysicalMotion.cpp
melanieplageman: +1 to 2

d:
Finally tidied up all the commits and bumped version. Changes since last Wednesday:

1. This is sort of the "staging" push, and ORCA version is bumped to 2.55.13
1. @bhuvnesh2703 the `dxl:comment` commit is separate from the main bug fix for posterity
1. @oarap `__attribute__((unused))` removed and I dropped the name for the unused argument in `CPhysicalSpool::FValidConext`
1. @hardikar I kept the outer ref refactor separate for posterity

Thoughts?

hardikar:
* Filed a chore for point 3.
* I'm ok with point 4. To be honest, I think the 4 commits you have here should be kept different. That is, do not squash 'em.

PR #352, opened 2018-03-12
Fallback to legacy query optimizer for queries over replicated table

pengzhout:


libnaucrates/src/md/IMDRelation.cpp
```diff
@@ -40,6 +41,10 @@ IMDRelation::PstrDistrPolicy
 		case EreldistrRandom:
 			return CDXLTokens::PstrToken(EdxltokenRelDistrRandom);
 		default:
+			GPOS_RAISE(
```
vraghavan78: i guess. but instead why not do it inside GPDB?

bhuvnesh2703: May be look at `CTranslatorRelcacheToDXL::Pmdrel()`, it calls `ereldistribution = Ereldistribution(pgppolicy);`. You could change the caller i.e `Pmdrel()` to handle Sentinel distribution policy and raise an exception. 
In debug build, currently it asserts in `Ereldistribution` so it falls back earlier than the place where you have added the exception.
In retail build, it goes to create a new object of `CMDRelationGPDB` and raises the exception added by you.
bC & EK

pengzhout: Ahh, found the place to fallback inside GPDB:

diff --git a/src/backend/gpopt/translate/CTranslatorRelcacheToDXL.cpp b/src/backend/gpopt/translate/CTranslatorRelcacheToDXL.cpp
index 0393c5e..3359033 100644
--- a/src/backend/gpopt/translate/CTranslatorRelcacheToDXL.cpp
+++ b/src/backend/gpopt/translate/CTranslatorRelcacheToDXL.cpp
@@ -922,7 +922,7 @@ CTranslatorRelcacheToDXL::Ereldistribution
                return IMDRelation::EreldistrMasterOnly;
        }

-       GPOS_ASSERT(!"Unrecognized distribution policy");
+       GPOS_RAISE(gpdxl::ExmaMD, ExmiDXLUnrecognizedType, GPOS_WSZ_LIT("unrecognized distribution policy"));
        return IMDRelation::EreldistrSentinel;
 }

Am I right? @bhuvnesh2703

pengzhout: @vraghavan78 @bhuvnesh2703 thanks, will close this one.

PR #355, opened 2018-03-26
Add unit test and re-organize code for serialization of COptimizerConfig

melanieplageman:
Previously, there was no unit test for the parsing or the serialization of a `COptimizerConfig` object. This commit adds this test.

This commit also introduces some re-organization of production code. The re-organization was motivated by the desire to clean up the call hierarchy for the serialization of the `COptimizerConfig` class.
The entry point for serializing a plan into DXL is `COptimizer::PdxlnOptimize`. If the minidump traceflag was set, it will initialize a minidump and serialize the various components needed.  Serialization of the `COptimizerConfig` is done at the end as part of `CMinidumperUtils::Finalize`, while adding the Error Context.

In the course of adding this test, it became clear that, though `CDXLUtils::SerializeOptimizerConfig was being called from CSerializeOptimizerConfig::Serialize`, our "production" code, in this case, it was bending over backwards to accommodate the need to add traceflags to the same XML serializer while still implementing the `CSerializable` interface for the `Serialize` function.
As such, we have moved the logic of this function into a `Serialize` function called directly on a `COptimizerConfig` object, which is called in production from `CSerializableOptimizerConfig::Serialize` and in test from `CParseHandlerTestUtils::SerializeOptimizerConfig`.
It should be noted that this test does not activate the parse handler for a `COptimizerConfig` directly and instead relies on the fact that a `COptimizerConfig` object is nested only one level below the top-level DXLMessage tag. Thus, this function calls `CDXLUtils::PoptimizerConfigParseDXL` which calls `PphdxlParseDXL` and uses the poconf member.


libgpopt/src/minidump/CSerializableOptimizerConfig.cpp
```diff
@@ -121,11 +76,12 @@ CSerializableOptimizerConfig::Serialize
...
 	)
 {
-	CDXLUtils::SerializeOptimizerConfig(m_pmp, oos, m_poconf, false /*fIndent*/);
-	SerializeTraceflags(oos);
+	CXMLSerializer xmlser(m_pmp, oos, false /*Indent*/);
 
-	// Footer
-	oos << CDXLSections::m_wszOptimizerConfigFooter;
+	// Copy traceflags from global state
+	CBitSet *pbs = CTask::PtskSelf()->Ptskctxt()->PbsCopyTraceFlags(m_pmp);
```
vraghavan78: do u need to copy it just to release it?

libnaucrates/src/CDXLUtils.cpp
```diff
@@ -568,6 +569,7 @@ CDXLUtils::PoptimizerConfigParseDXL
 	CParseHandlerDXL *pphdxl = PphdxlParseDXL(pmp, szDXL, szXSDPath);
 	CAutoP<CParseHandlerDXL> a_pphdxl(pphdxl);
 
+	pphdxl->Pbs();
 	// collect optimizer conf from dxl parse handler
```
dhanashreek89: Why is this needed? are you looking to extract and use the `traceflags` here ?

server/include/unittest/dxl/CParseHandlerTestUtils.h
```diff
@@ -0,0 +1,71 @@
...
+					m_szDXL = CDXLUtils::SzRead(Pmp(), szDXLFileName);
+
+					GPOS_CHECK_ABORT;
+
+					return m_szDXL.Rgt();
+				}
+		};
+
+	};
+
```
dhanashreek89: Why do we need a separate class here? Instead, could these helper routines be a part of `CParseHandlerOptimizerConfigSerializeTest` ?
I do not have a strong preference one way or the other, I am just curious.

data/dxl/parse_tests/OptimizerConfig.xml
```diff
@@ -1,14 +1 @@
...
-      <dxl:Hint MinNumOfPartsToRequireSortOnInsert="2147483647" JoinArityForAssociativityCommutativity="7" ArrayExpansionThreshold="25" JoinOrderDynamicProgThreshold="10" BroadcastThreshold="10000000" EnforceConstraintsOnDML="false"/>
-      <dxl:TraceFlags Value="102120,103001,103014,103015,103022,104003,104004,104005,105000"/>
-    </dxl:OptimizerConfig>
-  </dxl:Thread>
-</dxl:DXLMessage>
\ No newline at end of file
+<?xml version="1.0" encoding="UTF-8"?><dxl:DXLMessage xmlns:dxl="http://greenplum.com/dxl/2010/12/"><dxl:OptimizerConfig><dxl:EnumeratorConfig Id="0" PlanSamples="0" CostThreshold="0"/><dxl:StatisticsConfig DampingFactorFilter="0.750000" DampingFactorJoin="0.010000" DampingFactorGroupBy="0.750000"/><dxl:CTEConfig CTEInliningCutoff="0"/><dxl:WindowOids RowNumber="7000" Rank="7001"/><dxl:CostModelConfig CostModelType="1" SegmentsForCosting="3"/><dxl:Hint MinNumOfPartsToRequireSortOnInsert="2147483647" JoinArityForAssociativityCommutativity="7" ArrayExpansionThreshold="25" JoinOrderDynamicProgThreshold="10" BroadcastThreshold="10000000" EnforceConstraintsOnDML="false"/><dxl:TraceFlags Value=""/></dxl:OptimizerConfig></dxl:DXLMessage>
```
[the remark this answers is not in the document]
melanieplageman: So, when we serialize the `COptimizerConfig`, we make an xml serializer, `CXMLSerializer`, which requires us to specify whether or not to handle indentation. Even though indentation and whitespace are for readability and not significant in parsing XML, we are doing are test assertion with a literal string comparison, so whitespace would affect our result. We elected not to include indentation to avoid the chance of whitespace/indentation-related error when comparing expected with actual results--especially since we are adding a header and footer for the test which would not be added in production for the `COptimizerConfig`.

server/include/unittest/dxl/CParseHandlerOptimizerConfigSerializeTest.h
```diff
@@ -0,0 +1,20 @@
+//	Greenplum Database
```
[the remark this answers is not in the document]
melanieplageman: We can do this, however, my preference is to add a new file for each object under test. In this case, we only added the positive test case -- standard input and everything goes right -- however, ideally, we would add multiple tests for serialization of a COptimizerConfig and then the test file would have all of those tests in it but no tests for other objects.

libnaucrates/src/CDXLUtils.cpp
melanieplageman: The parsing method for parsing `COptimizerConfig` requires traceflags to be present in the serialized `COptimizerConfig`

server/include/unittest/dxl/CParseHandlerTestUtils.h
melanieplageman: I agree that `SerializeOptimizerConfig` doesn't need to be in here and can instead be in the file for those tests. The only reason I made a new file for `CParseHandler` test utils was because I anticipated other tests being added maybe wanting to use the test fixture which would be generic for a `CParseHandler`. However, after giving it further thought, I realize this test fixture is pretty much only relevant for serialization and not for parsing and that if someone adds an test file for serialization of a different object and wants to use the test fixture, he/she can move it at that time. Consequently, I have renamed it to `SerializeTestFixture` and moved it into the COptimizerConfig serialize test file. Thanks for the feedback!

libgpopt/src/minidump/CSerializableOptimizerConfig.cpp
melanieplageman: Currently, the only CTaskContext method for accessing the traceflags makes a copy. I'm not sure what the intent was. I can dig into it more and see if we should add an accessor for the traceflags bitset.

PR #356, opened 2018-03-28
Serialize nestloop factor

sambitesh:
Previously we did not serialize the cost parameter nestloop factor and hence it
was missing from the minidump files. When tests are run through ctest, they are
unable to read the nestloop factor from the minidump files and pick up the
default value of nestloop factor which is 1. This causes a lot of the tests to
fail. Hence we would like to serialize the nestloop factor so that when a
minidump file is added for testing purposes, it contains the desired nestloop
factor. This commit serializes nestloop factor inside CostParams and also adds
a unit test to check the serialization and deserialization of the CostParams.


server/src/unittest/dxl/CParseHandlerCostModelTest.cpp
```diff
@@ -0,0 +1,195 @@
...
+	ICostModel *pcm = pphcm->Pcm();
+
+	GPOS_RTL_ASSERT(ICostModel::EcmtGPDBCalibrated == pcm->Ecmt());
+	GPOS_RTL_ASSERT(3 == pcm->UlHosts());
+
+	CAutoRef<CCostModelParamsGPDB> pcpExpected(GPOS_NEW(pmp) CCostModelParamsGPDB(pmp));
+	pcpExpected->SetParam(CCostModelParamsGPDB::EcpNLJFactor, 1024.0, 1023.0, 1025.0);
+	GPOS_RTL_ASSERT(FEquals(pcpExpected.Pt(), pcm->Pcp()));
+
+	const CHAR *szExpectedString = "<dxl:CostModelConfig CostModelType=\"1\" SegmentsForCosting=\"3\">"
```
melanieplageman: Could you add a comment that would indicate that, for the current cost model, we want to test parsing it and make an assertion on the object which results from doing so *and * test serializing it back and make an assertion on the string that results from that, whereas with the legacy cost model, we are only testing parsing it? Maybe you could either split the current cost model parse and serialize tests up or name the test accordingly.

server/src/unittest/dxl/CParseHandlerCostModelTest.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+	fixture.Parse((const XMLByte *)a_szDXL.Rgt(), strlen(a_szDXL.Rgt()));
+
+	ICostModel *pcm = pphcm->Pcm();
+
+	GPOS_RTL_ASSERT(ICostModel::EcmtGPDBCalibrated == pcm->Ecmt());
+	GPOS_RTL_ASSERT(3 == pcm->UlHosts());
+
+	CAutoRef<CCostModelParamsGPDB> pcpExpected(GPOS_NEW(pmp) CCostModelParamsGPDB(pmp));
+	pcpExpected->SetParam(CCostModelParamsGPDB::EcpNLJFactor, 1024.0, 1023.0, 1025.0);
```
melanieplageman: I think I'm missing something. I don't understand why you have to manually set params here. Weren't those in your input file? Shouldn't parse set them for you somewhere when you parse the input file?

vraghavan78:
@melanieplageman please lets not add tests for legacy cost model. We need to kill the legacy cost model so no point testing it.

d:
Venky demanded a review. I'm re-starting

libgpdbcost/src/CCostModelParamsGPDB.cpp
```diff
@@ -455,5 +455,27 @@ CCostModelParamsGPDB::OsPrint
 	return os;
 }
 
+BOOL CCostModelParamsGPDB::FEquals(ICostModelParams *pcm) const
```
bhuvnesh2703: Formatting, we use like this..
BOOL 
CCostModelParamsGPDB::FEquals
                (
                ICostModelParams *pcm
                ) const

libgpdbcost/src/CCostModelParamsGPDB.cpp
```diff
@@ -455,5 +455,27 @@ CCostModelParamsGPDB::OsPrint
...
+	{
+		if (!m_rgpcp[ul]->FEquals(pcmgOther->m_rgpcp[ul]))
+			return false;
+	}
+
+	return true;
+}
+
+const CHAR *
+CCostModelParamsGPDB::SzNameLookup(ULONG ulId) const
```
bhuvnesh2703: formatting

libnaucrates/include/naucrates/dxl/CCostModelConfigSerializer.h
```diff
@@ -0,0 +1,25 @@
+//	Greenplum Database
+//	Copyright (C) 2018 Pivotal Software, Inc.
+
+#ifndef GPDXL_CCostModelConfigSerializer_H
+#define GPDXL_CCostModelConfigSerializer_H
+
+#include <gpopt/cost/ICostModel.h>
+#include "gpos/memory/IMemoryPool.h"
+#include "gpos/common/CAutoP.h"
```
bhuvnesh2703: Do you need CAutoP and IMemoryPool?

libnaucrates/src/CCostModelConfigSerializer.cpp
```diff
@@ -0,0 +1,47 @@
...
+#include "naucrates/dxl/xml/dxltokens.h"
+#include "gpdbcost/CCostModelParamsGPDB.h"
+
+#include "gpos/common/CAutoRef.h"
+#include "naucrates/dxl/xml/CXMLSerializer.h"
+
+using namespace gpdxl;
+using gpos::CAutoRef;
+
+void CCostModelConfigSerializer::Serialize(CXMLSerializer &xmlser) const
```
bhuvnesh2703: format

server/src/unittest/dxl/CParseHandlerCostModelTest.cpp
```diff
@@ -0,0 +1,195 @@
...
+	class Fixture
+	{
+		private:
+			CAutoMemoryPool m_amp;
+			gpos::CAutoP<CDXLMemoryManager> m_apmm;
+			std::auto_ptr<SAX2XMLReader> m_apxmlreader;
+			gpos::CAutoP<CParseHandlerManager> m_apphm;
+			gpos::CAutoP<CParseHandlerCostModel> m_apphCostModel;
+			gpos::CAutoP<CWStringDynamic> m_serializedString;
+			gpos::CAutoP<COstreamString> m_os;
```
d: this `COStream` object doesn't need to be wrapped in an `AutoP`: it can simply be on the stack.

server/src/unittest/dxl/CParseHandlerCostModelTest.cpp
```diff
@@ -0,0 +1,195 @@
...
+{
+	class Fixture
+	{
+		private:
+			CAutoMemoryPool m_amp;
+			gpos::CAutoP<CDXLMemoryManager> m_apmm;
+			std::auto_ptr<SAX2XMLReader> m_apxmlreader;
+			gpos::CAutoP<CParseHandlerManager> m_apphm;
+			gpos::CAutoP<CParseHandlerCostModel> m_apphCostModel;
+			gpos::CAutoP<CWStringDynamic> m_serializedString;
```
d: This `CWString` object need not be guarded by an `AutoP`, it can simply be embedded in the fixture (stack allocated)

server/src/unittest/dxl/CParseHandlerCostModelTest.cpp
```diff
@@ -0,0 +1,195 @@
...
+			gpos::CAutoP<COstreamString> m_os;
+			gpos::CAutoP<gpdxl::CXMLSerializer> m_apxmlser;
```
d: This `CXMLSerializer` object can be stack allocated. No need to wrap in `AutoP`

server/src/unittest/dxl/CParseHandlerCostModelTest.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+	const CHAR *szExpectedString = "<dxl:CostModelConfig CostModelType=\"1\" SegmentsForCosting=\"3\">"
+													"<dxl:CostParams>"
+													"<dxl:CostParam Name=\"NLJFactor\" Value=\"1024.000000\" LowerBound=\"1023.000000\" UpperBound=\"1025.000000\"/>"
+													"</dxl:CostParams>"
+													"</dxl:CostModelConfig>";
+
+	CAutoP<CCostModelConfigSerializer> cmcSerializer(GPOS_NEW(pmp) CCostModelConfigSerializer(pcm));
+	cmcSerializer->Serialize(*fixture.Pxmlser());
+	gpos::CAutoP<CWStringDynamic> apExpected(GPOS_NEW(pmp) CWStringDynamic(pmp));
```
d: This assertion clearly doesn't fit in this test case, consider putting the serialization test in a separate test case

d:
pushed the rebase and 3 new commits, check 'em out?

PR #359, opened 2018-04-03
Don't stop testing when a minidump test fails

dhanashreek89:
Previously, when a test is run as:
`$ ./server/gporca_test -U <>`
we stopped at the first failing minidump, be it a plan space change or
actual plan diff and did not continue with the subsequent minidumps.

This commit changes the behaviour such that we run through all the
minidump tests and do not stop at the first failure. The space size
change or plan change will be displayed in case of each failure. The
advantage is, we can get a quicker and consolidated feedback without
re-running the test after fixing each failing minidump.


vraghavan78:
look at CI failure. Unused param

/tmp/build/f8ad1a16/orca_src/server/src/unittest/CTestUtils.cpp:3734: warning: unused parameter 'ulTests'

PR #360, opened 2018-04-03
Card estimation based on NDV for equijoin

bhuvnesh2703:
Add cardinality estimation based on NDV for preds of form f(a) = b

Previously, predicates of the form f(a) = b were considered unsupported
and the cardinality for a join having such a predicate was
estimated using a constant scale factor. This resulted in gross
overestimation, and, in some cases, choosing an inferior plan.

With this commit, predicates of this form on an equijoin with exactly
one scalar ident on either side will be considered supported. This
commit introduces a new comparison type which will be used for
predicates like this and for which we estimate cardinality according to
the following formula: max(NDV(a), NDV(b)).


vraghavan78:
@hsyuan 

Let col(a) has 10 NDVs and col(b) has 100 NDVs and the following assumptions hold for the join condition
* predicate is and equality of the type f(a) = f(b) or f(a) = b or a = f(b)
* functions ensure that one value in a matches only one value in b

Then the number of ndv value of a and b after the join is the min(NDV(a), NDV(b)).

Scale factor the join condition is however max(NDV(a), NDV(b))

libnaucrates/src/statistics/CHistogram.cpp
```diff
@@ -2017,5 +2039,27 @@ CHistogram::PhistDefaultBoolColStats
...
+const
+{
+	CDouble pdScaleFactor = CScaleFactorUtils::DDefaultScaleFactorJoin;
+	
+	pdScaleFactor = std::max
+	(
+	 std::max(CHistogram::DMinDistinct.DVal(), dDistinctValuesOuter.DVal()),
+	 std::max(CHistogram::DMinDistinct.DVal(), dDistinctValuesInner.DVal())
+	 );
+	CDouble dCartesianProduct = dRows1 * dRows2;
```
vraghavan78: scale factor of the cartesian product is 1 right?

libnaucrates/src/statistics/CHistogram.cpp
```diff
@@ -2017,5 +2039,27 @@ CHistogram::PhistDefaultBoolColStats
 						);
 }
 
+// set scale factor for an equijoin based on NDV
+CDouble
+CHistogram::DEqualityNDVJoinScaleFactor
+(
+ CDouble dDistinctValuesOuter,
```
vraghavan78: too big and cannot make the difference dNdvOuter and dNdvInner

libnaucrates/src/statistics/CHistogram.cpp
```diff
@@ -2017,5 +2039,27 @@ CHistogram::PhistDefaultBoolColStats
...
+ CDouble dDistinctValuesInner,
+ CDouble dRows1,
+ CDouble dRows2
```
vraghavan78: we don't need this I think

libnaucrates/src/statistics/CStatsPredUtils.cpp
```diff
@@ -371,6 +371,24 @@ CStatsPredUtils::FCmpColsIgnoreCast
 
 	if (NULL == *ppcrLeft || NULL == *ppcrRight)
 	{
+		// if the scalar cmp is of equality type, we may not have been able to extract
+		// the column referenes of scalar ident if they had any other expression than cast
+		// on top of them.
+		// in such cases, check if there is still a possibility to extract scalar ident,
+		// if there is more than one column reference on either side, this is unsupported
+		// If supported, mark the comparision as NDV-based
+		if (*pescmpt == CStatsPred::EstatscmptEq)
```
vraghavan78: we should have a story for the following:

* f(a) != b
* f(a) IDF b
* f(a) INDF b

Will file more stories.

libnaucrates/src/statistics/CHistogram.cpp
```diff
@@ -2017,5 +2039,27 @@ CHistogram::PhistDefaultBoolColStats
...
+	CDouble pdScaleFactor = CScaleFactorUtils::DDefaultScaleFactorJoin;
+	
+	pdScaleFactor = std::max
+	(
+	 std::max(CHistogram::DMinDistinct.DVal(), dDistinctValuesOuter.DVal()),
+	 std::max(CHistogram::DMinDistinct.DVal(), dDistinctValuesInner.DVal())
+	 );
+	CDouble dCartesianProduct = dRows1 * dRows2;
+	// bound scale factor by cross product
+	return std::min(pdScaleFactor.DVal(), dCartesianProduct.DVal());
```
vraghavan78: std::max

vraghavan78:
@hsyuan i see what you said now, you are correct

bhuvnesh2703:
Fixed the comments, @hsyuan thanks

libnaucrates/include/naucrates/statistics/CStatsPred.h
```diff
@@ -52,6 +52,7 @@ namespace gpnaucrates
 				EstatscmptINDF,	// is not distinct from
 				EstatscmptLike,	// LIKE predicate comparison
 				EstatscmptNotLike,	// NOT LIKE predicate comparison
+				EstatscmptEqNDV, // NDV comparision for equality predicate
```
vraghavan78: NDV based comparison .... please add a comment saying this is for conditions such f(a) = b or f(a) = f(b), etc.

libnaucrates/src/statistics/CHistogram.cpp
```diff
@@ -2017,5 +2039,23 @@ CHistogram::PhistDefaultBoolColStats
...
+ )
+const
+{
+	CDouble pdScaleFactor = std::max
+	(
+	 std::max(CHistogram::DMinDistinct.DVal(), dNdvOuter.DVal()),
+	 std::max(CHistogram::DMinDistinct.DVal(), dNdvInner.DVal())
+	 );
+
+	// bound scale factor by cross product i.e 1
```
vraghavan78: lower bound ....

server/src/unittest/dxl/statistics/CJoinCardinalityNDVBasedEqPredTest.cpp
```diff
@@ -0,0 +1,303 @@
...
+
+	GPOS_RESULT
+	CJoinCardinalityNDVBasedEqPredTest::EresUnittest()
+	{
+		CUnittest rgut[] =
+				{
+						GPOS_UNITTEST_FUNC(EresUnittest_NDVEqCardEstimation),
+						GPOS_UNITTEST_FUNC(EresUnittest_NDVCardEstimationNotApplicableMultipleIdents),
+						GPOS_UNITTEST_FUNC(EresUnittest_NDVCardEstimationNotApplicableInequalityRange)
+				};
```
vraghavan78: @bhuvnesh2703 @melanieplageman we need more tests, especially covering the case that @hsyuan and l caught in the review.

libgpopt/src/base/CCastUtils.cpp
```diff
@@ -337,4 +337,18 @@ CCastUtils::PexprCast
...
+	(
+			CExpression *pexpr
+	)
+{
+	CDrvdPropScalar *pdrvdPropScalar = CDrvdPropScalar::Pdpscalar(pexpr->PdpDerive());
+	if (pdrvdPropScalar->PcrsUsed()->CElements() == 1)
+		return pdrvdPropScalar->PcrsUsed()->PcrFirst();
+	
+	return NULL;
+}
```
dhanashreek89: `CCastUtils` does not look like a correct place for this function as we are not checking/dealing with a cast here.

libnaucrates/src/statistics/CHistogram.cpp
```diff
@@ -2017,5 +2039,21 @@ CHistogram::PhistDefaultBoolColStats
...
+	)
+	const
+{
+	CDouble pdScaleFactor = std::max
+	 (
+	 std::max(CHistogram::DMinDistinct.DVal(), DDistinct().DVal()),
+	 std::max(CHistogram::DMinDistinct.DVal(), dNdvInner.DVal())
+	 );
+
+	return pdScaleFactor.DVal();
```
bhuvnesh2703: we should just return the scale factor here instead of having a lower bound of 1. 
- As, we will never come to this code while calculating the NDV for a join if there is atleast a side of the join with empty stats (we bail out early as the histograms are empty.) 
- If there is stats, either with 1 value (may be NULL or some other value) the min distinct will be 1, so the lower bound is by default 1 based on values.

server/src/unittest/dxl/statistics/CJoinCardinalityNDVBasedEqPredTest.cpp
bhuvnesh2703: i just mentioned the comment at https://github.com/greenplum-db/gporca/pull/360/files#diff-f2591b98f1b54ef61c747ed309099b3fR2056, i think we are good with the test case. pls review the comment

libnaucrates/src/statistics/CStatsPredUtils.cpp
dhanashreek89: s/comparision/comparison

libnaucrates/src/statistics/CHistogram.cpp
dhanashreek89: Do we need to cap the scalefactor to cross product, as we do in `CHistogram.cpp:710` ?

bhuvnesh2703: there we take minimum of those 2, `(*pdScaleFactor).DVal(), dCartesianProduct.DVal()`, but our intention here is to take the max of min distinct of either of the outer or inner. So, i don't think we need to cap it here.

dhanashreek89: Do we need to use this function for computing the scale factor of predicates of the form `f(a) != b`?
Specifically, in `DInEqualityJoinScaleFactor()`, the scale factor of inequality is computed from it's equality and we might need to use `CStatsPred:: EstatscmptEqNDV` there appropriately.

bhuvnesh2703: also even i believe that `*pdScaleFactor = std::min((*pdScaleFactor).DVal(), dCartesianProduct.DVal());` in `CHistogram::PhistJoinNormalized` is confusing. In what case could the min distinct which is calculated by could be more than cartesian product.

Case1:
If there is no stats => then the histogram is empty we bail out earlier and return the cartesian product

Case2: 
If there is min 1 distinct value => in that case as well it will be lower than cartesian product
hist1 NDV = 1 (total rows = 10)
hist2 NDV = 1  (total rows = 1)
scalefactor = 1, and cartesian = 10.
Case3:
If all the values are distinct still it will be less than cartesian product
hist1 NDV = 100
hist2 NDV = 100
so scalefactor =  100, and cartesian product 10,000
```
	*pdScaleFactor = std::max
						(
						std::max(DMinDistinct.DVal(), DDistinct().DVal()),
						std::max(DMinDistinct.DVal(), phistOther->DDistinct().DVal())
						);
```

Unless the stats are screwed for some reason, scalefactor will be lower than cartesian product. Do you think this justifies?

bhuvnesh2703: Currently `f(a) != b` is unsupported, we do NDV based comparison only for `=` via this PR.

data/dxl/statistics/Join-Statistics-NDVBasedCardEstimation-EqPred-Input.xml
```diff
@@ -0,0 +1,1222 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+create table foo (b int, c int);
+create table bar (d int);
+insert into bar select i, i from generate_series(1000,20000)i;
+insert into foo select i from generate_series(1,10000)i;
+
```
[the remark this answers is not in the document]
bhuvnesh2703: the xml here is not of the query but it contains only the stats, so didn't include the query.
The test which uses this files includes the query form (predicate example) marking the use of these buckets.

libnaucrates/src/statistics/CStatsPredUtils.cpp
bhuvnesh2703: fixed

PR #362, opened 2018-04-03
Only check dynamic programming threshold for ExpandNAryJoinDP xform

sambitesh:
`CXformUtils::ExfpExpandJoinOrder` is common to both transformations -
`CXformExpandNAryJoinMinCard` and `CXformExpandNAryJoinDP`. So we moved out
checking for dynamic programming threshold out of it. Before this, in case
number of relations exceeded the dynamic programming threshold, ORCA would not
even go for `CXformExpandNAryJoinMinCard`.


vraghavan78:
@d it is WIP since more tests need to be added. I asked him to open it for folks who want to look an option.

vraghavan78:
@d not a WIP anymore.

vraghavan78:
The minidump file, has no SQL to re-generate it later on.

hsyuan:
+1

Issue #358, opened 2018-03-29
GPOS clobbers standard integer macros INT_MIN, INT_MAX, ULONG_MAX, ULLONG_MAX, etc


d:
## Issue
The C programming language standard [^c99] and [POSIX standard (SUS v1 onward, 1995)](http://pubs.opengroup.org/onlinepubs/7908799/xsh/limits.h.html) [^sus_v2] [^posix_01] defines the standard header `<limits.h>` (or `<climits>` for C++) where various standard integral type limits are defined.

GPOS, however, [deliberately clobbers at least 4 of those standard integral limit macros (using `#undef`)](https://github.com/greenplum-db/gporca/blob/v2.55.13/libgpos/include/gpos/types.h#L70-L73). This causes several problems:

1. While working on issue #356 the pair discovered that including Xerces C++ headers after GPOS headers would lead to a compilation failure:

    ```
    /usr/local/include/xercesc/validators/common/Grammar.hpp:68: error: 'INT' was not declared in this scope
    /usr/local/include/xercesc/validators/common/Grammar.hpp:68: error: 'ULONG' was not declared in this scope
    /usr/local/include/xercesc/validators/common/Grammar.hpp:68: error: a function call cannot appear in a constant-expression
    /usr/local/include/xercesc/validators/common/Grammar.hpp:69: error: 'INT' was not declared in this scope
    /usr/local/include/xercesc/validators/common/Grammar.hpp:69: error: 'ULONG' was not declared in this scope
    /usr/local/include/xercesc/validators/common/Grammar.hpp:69: error: a function call cannot appear in a constant-expression
    ```
1. The worse thing that can happen than a compilation failure is a successful compilation: we only get surprised at runtime. Libraries we use (e.g. Xerces C++) may use these standard macros in their headers. The libraries we link to will be compiled with the *actual* standard values of those macros. By overriding `INT_MAX` and `ULONG_MAX`, for example, ORCA may assume one "orca value" of `UINT_MAX` (which in many implementations is defined in terms in `INT_MAX`) from Xerces, but will link to Xerces library code that uses a different value of `UINT_MAX`.

1. Longer feedback loop: this sort of falls into the bucket of "bad practice", which means that you _don't necessarily_ get a failure on the platform and toolchain that you develop (say, Clang on Darwin), but will fail when you run a "more serious" test (say GCC on Linux).

1. This also impedes the best-practice that header files are conventionally re-orderable (with judicious exceptions)

## Question for the team

[...]

PR #365, opened 2018-04-09
Avoid collision w std macros use enum

asubramanya:


libgpos/include/gpos/types.h
```diff
@@ -96,21 +94,17 @@ namespace gpos
...
-#undef INT_MAX
-#endif // INT_MAX
-#define INT_MAX			((INT) (ULONG_MAX >> 1))
-
-#ifdef INT_MIN
-#undef INT_MIN
-#endif // INT_MIN
-#define INT_MIN			(-INT_MAX - 1)
 
 #ifdef LINT_MAX
```
hardikar: Any reason why didn't remove LINT_MAX, LINT_MIN, SINT_MAX & SINT_MIN?

d: 1. `INT_MAX`, `INT_MIN`, `ULONG_MAX` and `ULLONG_MAX` are standard macros that we are clobbering, so the minimal fix to enable us to safely `#include` xerces after GPOS headers are to stop clobbering those

1. We could, if we have a consensus, expand the scope to fixing all of those integral limits. I leave it up to the team to decide

asubramanya:
@hardikar Updated the PR to change SINT and LINT MAX and MIN values to enums as well.

libgpos/include/gpos/types.h
```diff
@@ -66,24 +65,23 @@ namespace gpos
 
 	typedef uint32_t ULONG;
```
melanieplageman: Could we elaborate on the above comment a little bit? 
- Why is this needed? 
- More importantly, how and when should the API user use it. (i.e. whenever writing code in Orca? only under certain circumstances?)

Also, in the PR or commit message, could you put the reason or motivation for using our own types and for changing them to enums? The namespace prefix is more self-explanatory.

asubramanya: I tried looking up the history to find out more about why we created our own types but did not find anything. I will update the commit message with more details from the issue https://github.com/greenplum-db/gporca/issues/358 
before committing it to master.

asubramanya:
@hardikar Corresponding GPDB PR are here: https://github.com/greenplum-db/gpdb/pull/4839

libgpos/include/gpos/memory/CCacheAccessor.h
```diff
@@ -161,9 +161,9 @@ namespace gpos
 				GPOS_ASSERT(NULL == m_pmp);
 
 				// construct a memory pool for cache entry
-				m_pmp = CMemoryPoolManager::Pmpm()->PmpCreate(
-					CMemoryPoolManager::EatTracker, true /*fThreadSafe*/,
-					gpos::ullong_max);
+                m_pmp = CMemoryPoolManager::Pmpm()->PmpCreate(
```
d: use tabs here

libgpos/src/memory/CCacheFactory.cpp
```diff
@@ -70,8 +70,8 @@ CCacheFactory::EresInit()
 	GPOS_RESULT eres = GPOS_OK;
 
 	// create cache factory memory pool
-	IMemoryPool *pmp = CMemoryPoolManager::Pmpm()->PmpCreate(
-		CMemoryPoolManager::EatTracker, true /*fThreadSafe*/, gpos::ullong_max);
+    IMemoryPool *pmp = CMemoryPoolManager::Pmpm()->PmpCreate(
+        CMemoryPoolManager::EatTracker, true /*fThreadSafe*/, gpos::ullong_max);
```
d: why did we change from tab to space here?

libnaucrates/src/statistics/CStatsPredUtils.cpp
```diff
@@ -558,9 +558,9 @@ CStatsPredUtils::AddSupportedStatsFilters
 
 	if (COperator::EopScalarConst == pexprPred->Pop()->Eopid())
 	{
-		pdrgpstatspred->Append(GPOS_NEW(pmp) CStatsPredUnsupported(
-			gpos::ulong_max, CStatsPred::EstatscmptOther,
-			CHistogram::DNeutralScaleFactor));
+        pdrgpstatspred->Append(GPOS_NEW(pmp) CStatsPredUnsupported(
+            gpos::ulong_max, CStatsPred::EstatscmptOther,
```
d: I don't think you meant it when you changed the tabs to spaces here.

libgpos/include/gpos/types.h
```diff
@@ -96,41 +94,36 @@ namespace gpos
...
-#ifdef SINT_MAX
-#undef SINT_MAX
-#endif // SINT_MAX
-#define SINT_MAX		((SINT) (USINT_MAX >> 1))
-
-#ifdef SINT_MIN
-#undef SINT_MIN
-#endif // SINT_MIN
-#define SINT_MIN		(-SINT_MAX - 1)
+	enum
```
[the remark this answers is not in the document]
d: I doubt it: I vaguely recall the compiler determines the *underlying type* of each enumeration by its enumerators. Which means that you often don't want to mix unsigned enumerators with signed ones -- especially in our case. I need to dig into this a bit more. Sit tight @hardikar

d:
Re-posting just so that it doesn't get buried in the history

@hardikar wrote:
> Just curious - can these be combined into one enum { } declaration?

My take: I doubt it. I vaguely recall the compiler determines the *underlying type* of each enumeration by its enumerators. Which means that you often don't want to mix unsigned enumerators with signed ones -- especially in our case.

The fact that we are using unnamed enumerations doesn't mean that semantically all constants (enumerators) here have the same type. Quite the opposite, you'd generally want `gpos::int_min` to have a different type from `gpos::ulong_max`.

d:
Just talked to @hardikar and @dhanashreek89 . Here's a demonstration of how [mixing enumerators is surprising](https://wandbox.org/permlink/0qB3msgZRqqcgKuH)

```c++
#include <climits>
#include <iostream>

namespace jesse {
enum {
  unnamed_enum_int_min = INT_MIN,
  unnamed_enum_ullong_max = ULLONG_MAX,
};
}

using std::cout;

int main() {
  cout << "unnamed_enum_int_min:\t" << jesse::unnamed_enum_int_min << '\t'
       << "unnamed_enum_ullong_max:\t" << jesse::unnamed_enum_ullong_max
       << '\n';
  return 0;
}

```

PR #368, opened 2018-04-14
WIP: Support collation for const

melanieplageman:
In GPDB after the Postgres 9.1 merge, we will support collation. Therefore, we must plumb the collation oids through Orca. This PR does this for constants.
The corresponding translator changes are split across two commits in a branch off of the gpdb-postgres-merge repo (greenplum-db/gpdb-postgres-merge@53333e4) and a second which has not yet landed there. All merged in changes will be squashed into a single merge commit, so, it will be necessary to reference the merge commit for iteration_8 of the gpdb/postgres merge if looking for the sister commit to this.

Below are the details of what is changed here. I am looking for feedback on the decision to overload certain these functions to support older versions of GPDB and on the choice to have an uninitialized member of instances of these classes for those constructed with the backwards-compatible constructor.

Add collation to the following operators to support collation in Const and overload constructors for versions of GPDB for which collation is not supported:
`CDatumGenericGPDB`
`CDXLDatum`
`CDXLDatumGeneric`
`CDXLDatumStatsDoubleMappable`
`CDXLDatumStatsLintMappable`

  These are callers that were updated that use the CDXLDatum constructor:
  `CDXLDatumInt2`
  `CDXLDatumInt4`
  `CDXLDatumInt8`
  `CDXLDatumBool`
  `CDXLDatumOid`

Overload the following functions so that versions of GPDB with no collation are supported:
`CMDTypeGenericGPDB::PdxldatumStatsDoubleMappable`
`CMDTypeGenericGPDB::PdxldatumStatsLintMappable`
`CMDTypeGenericGPDB::Pdxldatum`

### DIFF
--- a/data/dxl/minidump/collation_implicit_ddl_explicit_pred_const.mdp
+++ b/data/dxl/minidump/collation_implicit_ddl_explicit_pred_const.mdp
@@ -1,0 +1,683 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<dxl:DXLMessage xmlns:dxl="http://greenplum.com/dxl/2010/12/">
+  <dxl:Thread Id="0">
+    <dxl:OptimizerConfig>
+      <dxl:EnumeratorConfig Id="0" PlanSamples="0" CostThreshold="0"/>
+      <dxl:StatisticsConfig DampingFactorFilter="0.750000" DampingFactorJoin="0.010000" DampingFactorGroupBy="0.750000"/>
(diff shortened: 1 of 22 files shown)


melanieplageman:
Closing this PR in favor of combined collation work on this branch https://github.com/greenplum-db/gporca/tree/orca_collation_support which will result in a single commit to master and corresponding translator changes for which the current plan is to incorporate them into the 9.1 merge commit.

PR #369, opened 2018-05-02
Add missing Xforms that generate IndexedNLJ to PbsIndexJoinXforms

vraghavan78:
PbsIndexJoinXforms returns a set of xforms that generate indexed nested loop join. The xforms added where not added to this function when they where created, this causes xforms to be fired when the user wants to disable all xforms that generate indexed nested loop join


vraghavan78:
Added.

hardikar:
This needs a patch version bump before merging.

PR #370, opened 2018-05-07
Exit early during constraint derivation for unsupport comparison oper…

vraghavan78:
…ators

GPDB OSS issue: https://github.com/greenplum-db/gpdb/issues/4938

For the following query:

```sql
create table like_error (some_text text);
select * from like_error where some_text LIKE ANY (ARRAY['%text1%']);
```

In ORCA's constraint derivation, the LIKE predicate is not a supported operation. We should therefore no do further processing for all such unsupported operations and move on to the next supported predicate.

Prior to this fix, we would crash. This issue has now been resolved.


d:
Leak

vraghavan78:
@d fixed the leak. Thanks! Please take a look.

PR #371, opened 2018-05-08
Handle the case when unnesting subquery returns NULL

oarap:
When traceflag 103021 is set orca enforces a plan with the subplan.
Therefore, unnesting subquery will return NULL. This PR handles this
case.


khannaekta:
Can we also add a test for this?

PR #387, opened 2018-07-26
Enable static partition elimination when predicate has IDF and implicit

vraghavan78:
cast operation

Before this fix, we ignored scalar comparison of the type:

x::implicit-cast IDF constant-value
x IDF constant-value::implicit-cast
x::implicit-cast IDF constant-value::implicit-cast

ORCA does constraint derivation for INT4, INT8, INT2 and OID columns.
This constraint derivation is then used to compute partition filters.

By ignoring casts we did not use the predicte to do static partition
elimination.


vraghavan78:
Found a bug. Closing will reopen once it is fixed.

PR #388, opened 2018-07-27
Enable partition elimination for IDF comparison with cast

vraghavan78:
Before this fix, we ignored scalar comparison of the type:

x::implicit-cast IDF constant-value
x IDF constant-value::implicit-cast
x::implicit-cast IDF constant-value::implicit-cast

ORCA does constraint derivation for INT4, INT8, INT2 and OID columns.
This constraint derivation is then used to compute partition filters.

By ignoring casts we did not use the predicte to do static partition
elimination.


vraghavan78:
After little more thought:

```
When a cast is present, we cannot use the "CConstraintInterval" framework because it only maintains the pcr and not the entire LHS expression. 
```

:( closing this PR again

PR #391, opened 2018-08-01
Force Redistribute Motion on Insert on Randomly Distributed Tables

bhuvnesh2703:
If the trace flag EopttraceForceRedistributeOnInsertOnRandomDistrTables
is set, force a redistribute motion before executing insert on randomly
distributed table. This will ensure random data distribution across all
the segments, otherwise the data may be inserted into only 1 segment or a
few depending on the data distribution fed to Insert node.

Plan and data distribution after the commit.
```
create table foo (a int, b int) distributed randomly;
create table bar (a int, b int) distributed by (b);
insert into bar select i, i from generate_series(1,20)i;
create table baz (a int, b int) distributed randomly;
insert into baz select i, i from generate_series(1,20)i;
bchaudhary=# explain insert into foo values(1,1), (1,2), (1,3), (2,3), (23,234), (2342,2342432);
                                       QUERY PLAN
-----------------------------------------------------------------------------------------
 Insert  (cost=0.00..0.09 rows=2 width=8)
   ->  Redistribute Motion 3:3  (slice1; segments: 3)  (cost=0.00..0.00 rows=2 width=12)
         ->  Result  (cost=0.00..0.00 rows=2 width=12)
               ->  Result  (cost=0.00..0.00 rows=2 width=8)
                     ->  Values Scan on "Values"  (cost=0.00..0.00 rows=2 width=8)
 Optimizer: PQO version 2.67.0
(7 rows)

bchaudhary=# insert into foo values(1,1), (1,2), (1,3), (2,3), (23,234), (2342,2342432);
INSERT 0 6
Time: 100.191 ms
bchaudhary=# select count(*), gp_segment_id from foo group by 2;
 count | gp_segment_id
-------+---------------
     2 |             0
     2 |             1
     2 |             2

Time: 102.034 ms
truncate foo;
bchaudhary=# explain insert into foo select * from bar;
                                              QUERY PLAN
 Insert  (cost=0.00..431.02 rows=1 width=8)
   ->  Redistribute Motion 3:3  (slice2; segments: 3)  (cost=0.00..431.00 rows=1 width=12) ==> additional redistribute motion even though the data is redistributed already.
         ->  Result  (cost=0.00..431.00 rows=1 width=12)
               ->  Redistribute Motion 3:3  (slice1; segments: 3)  (cost=0.00..431.00 rows=1 width=8)
                     ->  Table Scan on bar  (cost=0.00..431.00 rows=1 width=8)
 Optimizer: PQO version 2.67.0
(6 rows)
bchaudhary=# insert into foo select * from bar;
INSERT 0 20
bchaudhary=# select count(*), gp_segment_id from foo group by 2;
 count | gp_segment_id
-------+---------------
     9 |             0
     8 |             1
     3 |             2

Time: 94.081 ms
bchaudhary=# explain insert into foo select * from baz;
                                        QUERY PLAN
 Insert  (cost=0.00..431.02 rows=1 width=8)
   ->  Redistribute Motion 3:3  (slice1; segments: 3)  (cost=0.00..431.00 rows=1 width=12)
         ->  Result  (cost=0.00..431.00 rows=1 width=12)
               ->  Table Scan on baz  (cost=0.00..431.00 rows=1 width=8)
 Optimizer: PQO version 2.67.0
bchaudhary=# insert into foo select * from baz;
INSERT 0 20
Time: 117.243 ms
bchaudhary=# select count(*), gp_segment_id from foo group by 2;
 count | gp_segment_id
-------+---------------
     6 |             0
    10 |             1
     4 |             2
```

Corresponding GPDB PR: https://github.com/greenplum-db/gpdb/pull/5369

### DIFF
--- a/libgpopt/src/translate/CTranslatorExprToDXL.cpp
+++ b/libgpopt/src/translate/CTranslatorExprToDXL.cpp
@@ -5332,6 +5332,45 @@ CTranslatorExprToDXL::PdxlnDML
 
 	CDXLNode *pdxlnChild = Pdxln(pexprChild, pdrgpcrSource, pdrgpdsBaseTables, pulNonGatherMotions, pfDML, false /*fRemap*/, false /*fRoot*/);
 
+	// add a redistribute motion if the target table is randomly distributed and trace flag is set
+	BOOL is_table_randomly_distributed = ptabdesc->Ereldistribution() == IMDRelation::EreldistrRandom;
+	CDXLNode *random_motion_dxl_node = NULL;
+	if (is_table_randomly_distributed && GPOS_FTRACE(EopttraceForceRedistributeOnInsertOnRandomDistrTables))
+	{
+		CDXLPhysicalRandomMotion *random_motion_dxl_op = GPOS_NEW(m_pmp) CDXLPhysicalRandomMotion(m_pmp, false);
+		random_motion_dxl_node = GPOS_NEW(m_pmp) CDXLNode(m_pmp, random_motion_dxl_op);
+
+		CDXLProperties *child_dxl_prop = pdxlnChild->Pdxlprop();
+		child_dxl_prop->AddRef();
+		random_motion_dxl_node->SetProperties(child_dxl_prop);
+
+		// construct project list
+		CDXLNode *child_proj_list_dxl_node = (*pdxlnChild)[0];
+		CDXLNode *proj_list_dxl_node = CTranslatorExprToDXLUtils::PdxlnProjListFromChildProjList(m_pmp, m_pcf, m_phmcrdxln, child_proj_list_dxl_node);
+
+		// set redistribution segment info n:n
+		DrgPi *segment_id_array = GPOS_NEW(m_pmp) DrgPi(m_pmp);
+		ULONG num_of_segments = COptCtxt::PoctxtFromTLS()->Pcm()->UlHosts();
+		for (ULONG i = 0; i < num_of_segments; i++)
+		{
+			segment_id_array->Append(GPOS_NEW(m_pmp) INT(i));
+		}
+		segment_id_array->AddRef();
+		random_motion_dxl_op->SetSegmentInfo(segment_id_array, segment_id_array);
+
+		// construct an empty filter node
+		CDXLNode *filter_dxl_node = PdxlnFilter(NULL /*pdxlnCond*/);
+
+		// construct sort column list
+		CDXLNode *sort_col_list_dxl_node = GPOS_NEW(m_pmp) CDXLNode(m_pmp, GPOS_NEW(m_pmp) CDXLScalarSortColList(m_pmp));
+
+		// add children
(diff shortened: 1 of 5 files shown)


danielgustafsson:
Isn't this what the round robin approach in `cdbhashnokey()` is supposed to solve?

vraghavan78:
You are correct. The problem here is different. They are of two kinds:

* Orca currently introduces a hash filter that ensures that no duplicates
are inserted when u have universally available data like ValueScan. However
this hash filter makes all the tuples to go on one segment. Which causes
skew.
* Say if the SELECT subquery of an insert is randomly distributed, ORCA
enforcers will no insert new redistribute motion on top of it, even if when
the SELECT subquery results are skewed.

Planner for both these cases forces a spraying of tuple with a redistribute
random on top of the SELECT subquery.

Orca with this fix will do the same.

V


ᐧ

danielgustafsson:
> On 1 Aug 2018, at 19:40, Venkatesh Raghavan <notifications@github.com> wrote:
> 
> You are correct. The problem here is different. They are of two kinds:
> 
> * Orca currently introduces a hash filter that ensures that no duplicates
> are inserted when u have universally available data like ValueScan. However
> this hash filter makes all the tuples to go on one segment. Which causes
> skew.
> * Say if the SELECT subquery of an insert is randomly distributed, ORCA
> enforcers will no insert new redistribute motion on top of it, even if when
> the SELECT subquery results are skewed.
> 
> Planner for both these cases forces a spraying of tuple with a redistribute
> random on top of the SELECT subquery.
> 
> Orca with this fix will do the same.

Aha! Thanks a lot for breaking down the issue, now I see why this patch is needed and is an improvement.

libgpopt/src/translate/CTranslatorExprToDXL.cpp
```diff
@@ -5332,6 +5332,45 @@ CTranslatorExprToDXL::PdxlnDML
 
 	CDXLNode *pdxlnChild = Pdxln(pexprChild, pdrgpcrSource, pdrgpdsBaseTables, pulNonGatherMotions, pfDML, false /*fRemap*/, false /*fRoot*/);
 
+	// add a redistribute motion if the target table is randomly distributed and trace flag is set
+	BOOL is_table_randomly_distributed = ptabdesc->Ereldistribution() == IMDRelation::EreldistrRandom;
+	CDXLNode *random_motion_dxl_node = NULL;
+	if (is_table_randomly_distributed && GPOS_FTRACE(EopttraceForceRedistributeOnInsertOnRandomDistrTables))
+	{
+		CDXLPhysicalRandomMotion *random_motion_dxl_op = GPOS_NEW(m_pmp) CDXLPhysicalRandomMotion(m_pmp, false);
```
oarap: I see that a lot of code has been borrowed from `CTranslatorExprToDXL::PdxlnMotion`. This suggests a refactoring.

bhuvnesh2703:
Closing this PR, as created another PR: https://github.com/greenplum-db/gporca/pull/393

PR #393, opened 2018-08-07
Add Post Process step to insert random motion if required

bhuvnesh2703:
In order to ensure the randomness of the data inserted into a randomly
distributed table, the child of the DML (CTAS or Insert) operator must provide Random
Distribution Spec, however there are cases where the input tree provides random
distribution spec but it will not gaurantee random distribution of data in the
target table.

For instance,
1. If data is inserted into a randomly distributed table from
another randomly distributed table, optimization framework will not add a
redistribute motion as the requested random spec by DML node will match the derived
spec of the child. However, from the perspective of the target table on which
insert is performed, the random spec of the child does not gaurantee
randomess.
Consider a query like below:
```
create table t1_random(a int, b int) distributed randomly;
explain select * from t1_random where gp_segment_id = 1;
Physical plan:
+--CPhysicalDML (Insert, "t1_random") ==> Matches Random Spec of the child
   +--CPhysicalComputeScalar
      |--CPhysicalFilter
      |  |--CPhysicalTableScan "t1_random" ==> Provides Random Spec
      |  +--CScalarCmp (=)
      |     |--CScalarIdent
      |     +--CScalarConst
      +--CScalarProjectList
         +--CScalarProjectElement
            +--CScalarConst
",
                                        QUERY PLAN
 Insert  (cost=0.00..431.02 rows=1 width=8)
   ->  Result  (cost=0.00..431.00 rows=1 width=12)
         ->  Table Scan on t1_random1  (cost=0.00..431.00 rows=1 width=8)
               Filter: gp_segment_id = 1
```
In the above query, target table t1_random is randomly distributed, and the
source table is randomly distributed. Thus, DML (Insert) operator will not
enforce a redistribute motion, however due to the filter gp_segment_id=1, data
will only be fetch from 1 segment and inserted into 1 segment of the target
table, resulting in skewed results.

2. If data is inserted into a randomly distributed table from a child with
universal spec (like Values or ConstTableGet), which means that the universal
child could be executed anywhere in the cluster with a Result node
having hash filter on top of it to avoid duplicates (This result node
provides random distribution). Hash filter will ensure that data from
only one segment is inserted into the target table, but that will again
lead to skewness of data from the perspective of target table.

```
explain insert into t1_random values (1,2);
Physical plan:
+--CPhysicalDML (Insert, "t1_random") ==> Matches Random Spec of child
   +--CPhysicalComputeScalar
      |--CPhysicalMotionRandom   ==> Random Motion Spec with Duplicate Hazard, converted later to result node with hash filter
      |  +--CPhysicalConstTableGet ==> Universal Spec child
      +--CScalarProjectList
         |--CScalarProjectElement
         |  +--CScalarConst (1)
         |--CScalarProjectElement "a" (1)
         |  +--CScalarConst (1)
         +--CScalarProjectElement "b" (2)
            +--CScalarConst (2)
                                       QUERY PLAN
-----------------------------------------------------------------------------------------
 Insert  (cost=0.00..0.00 rows=1 width=12)
   ->  Result  (cost=0.00..0.00 rows=1 width=12)
 Optimizer: PQO version 2.67.0
```

This commit fixes the issue by identifying the spec of the incoming
child for a DML Insert or CTAS, and if a redistribute motion does not
exists, the post processing step will update the plan with a
redistribute motion.

Resulting plans will look like:

Query 1: explain insert into t1_random select * from t1_random where gp_segment_id = 1;
```
Physical plan:
+--CPhysicalDML (Insert, "t1_random")
   +--CPhysicalMotionRandom ==> New Random Motion
      +--CPhysicalComputeScalar
         |--CPhysicalFilter
         |  |--CPhysicalTableScan "t1_random"
         |  +--CScalarCmp (=)
         |     |--CScalarIdent "gp_segment_id"
         |     +--CScalarConst (1)
         +--CScalarProjectList
            +--CScalarProjectElement "ColRef_0009"
               +--CScalarConst (1)
",
                                        QUERY PLAN
 Insert  (cost=0.00..431.02 rows=1 width=8)
   ->  Redistribute Motion 3:3  (slice1; segments: 3)  (cost=0.00..431.00 rows=1 width=12) ==> New Random Motion
         ->  Result  (cost=0.00..431.00 rows=1 width=12)
               ->  Table Scan on t1_random1  (cost=0.00..431.00 rows=1 width=8)
                     Filter: gp_segment_id = 1
 Optimizer: PQO version 2.67.0
```

Query 2: explain insert into t1_random1 values (1,2);
```
Physical plan:
+--CPhysicalDML (Insert, "t1_random"),
   +--CPhysicalMotionRandom ==> New Random Motion
      +--CPhysicalComputeScalar
         |--CPhysicalMotionRandom
         |  +--CPhysicalConstTableGet
         +--CScalarProjectList
            |--CScalarProjectElement
            |  +--CScalarConst (1)
            |--CScalarProjectElement "a"
            |  +--CScalarConst (1)
            +--CScalarProjectElement "b"
               +--CScalarConst (2)
",
                                       QUERY PLAN
-----------------------------------------------------------------------------------------
 Insert  (cost=0.00..0.02 rows=1 width=8)
   ->  Redistribute Motion 3:3  (slice1; segments: 3)  (cost=0.00..0.00 rows=1 width=12) ==> New Random Motion
         ->  Result  (cost=0.00..0.00 rows=1 width=12)
 Optimizer: PQO version 2.67.0
```
Corresponding GPDB PR: https://github.com/greenplum-db/gpdb/pull/5412

### DIFF
--- a/libgpopt/include/gpopt/operators/CExpression.h
+++ b/libgpopt/include/gpopt/operators/CExpression.h
@@ -16,6 +16,7 @@
 #include "gpos/common/CDynamicPtrArray.h"
 
 #include "naucrates/statistics/CStatistics.h"
+#include "gpopt/base/CDrvdPropPlan.h"
 #include "gpopt/cost/CCost.h"
 #include "gpopt/base/CColRef.h"
 #include "gpopt/base/CCostContext.h"
@@ -25,12 +26,13 @@
 #include "gpopt/base/CReqdPropRelational.h"
 #include "gpopt/base/CPrintPrefix.h"
 #include "gpopt/operators/COperator.h"
+#include "gpopt/base/CDrvdPropRelational.h"
 
 
 namespace gpopt
 {
 	// cleanup function for arrays
-	class CExpression;	
+	class CExpression;
 	typedef CDynamicPtrArray<CExpression, CleanupRelease> DrgPexpr;
 
 	// array of arrays of expression pointers
@@ -233,8 +235,7 @@ namespace gpopt
 			// get expression's derived property given its type
 			CDrvdProp *Pdp(const CDrvdProp::EPropType ept) const;
 
-			// get derived statistics object
-			const IStatistics *Pstats() const
+			IStatistics *Pstats() const
 			{
 				return m_pstats;
 			}
@@ -319,6 +320,24 @@ namespace gpopt
 			// rehydrate expression from a given cost context and child expressions
 			static
 			CExpression *PexprRehydrate(IMemoryPool *pmp, CCostContext *pcc, DrgPexpr *pdrgpexpr, CDrvdPropCtxtPlan *pdpctxtplan);
+
+			CDrvdPropRelational *GetDrvdPropRelational() const;
+
+			void SetDrvdPropPlan(CDrvdPropPlan *drvd_prop_plan);
+
+			CDrvdPropPlan *GetDrvdPropPlan() const;
+
+			void SetStats(IStatistics *stats);
+
+			void SetCost(CCost cost);
+
+			void SetReqdPropPlan(CReqdPropPlan *reqd_prop_plan);
+
+			void SetDrvdPropRelational(CDrvdPropRelational *drvd_prop_relational);
+
+			// copy relational, derived, required, statistics and cost of input expression
+			void
+			CopyExpressionProp(CExpression *expr);
 
 
 	}; // class CExpression
--- a/libgpopt/src/operators/CExpression.cpp
+++ b/libgpopt/src/operators/CExpression.cpp
@@ -1536,4 +1536,97 @@ CExpression::FValidPartEnforcers
 	return true;
 }
 
+void
+CExpression::CopyExpressionProp
+	(
+	CExpression *expr
+	)
+{
+	CDrvdPropPlan *drvd_prop_plan = expr->GetDrvdPropPlan();
+	drvd_prop_plan->AddRef();
+	this->SetDrvdPropPlan(drvd_prop_plan);
+
+	CDrvdPropRelational *drvd_prop_relational = expr->GetDrvdPropRelational();
+	drvd_prop_relational->AddRef();
+	this->SetDrvdPropRelational(drvd_prop_relational);
+
+	CReqdPropPlan *reqd_prop_plan = expr->Prpp();
+	reqd_prop_plan->AddRef();
+	this->SetReqdPropPlan(reqd_prop_plan);
+
+	CCost cost = expr->Cost();
+	this->SetCost(cost);
+
+	IStatistics *stats = expr->Pstats();
+	stats->AddRef();
+	this->SetStats(stats);
+}
+
+void
+CExpression::SetDrvdPropPlan
+	(
+	CDrvdPropPlan *drvd_prop_plan
+	)
+{
+	GPOS_ASSERT(NULL != drvd_prop_plan);
+	GPOS_ASSERT(NULL == m_pdpplan);
+	m_pdpplan = drvd_prop_plan;
+}
+
+CDrvdPropPlan *
+CExpression::GetDrvdPropPlan() const
+{
+	return m_pdpplan;
+}
+
+CDrvdPropRelational *
+CExpression::GetDrvdPropRelational() const
+{
+	return m_pdprel;
+}
+
+void
+CExpression::SetDrvdPropRelational
+	(
+	CDrvdPropRelational *drvd_prop_relational
+	)
+{
+	GPOS_ASSERT(NULL == m_pdprel);
+	GPOS_ASSERT(NULL != drvd_prop_relational);
+	m_pdprel = drvd_prop_relational;
+}
+
+void
+CExpression::SetReqdPropPlan
+	(
+	CReqdPropPlan *reqd_prop_plan
+	)
+{
+	GPOS_ASSERT(NULL == m_prpp);
+	GPOS_ASSERT(NULL != reqd_prop_plan);
+	m_prpp = reqd_prop_plan;
+}
+
+void
+CExpression::SetStats
+	(
+	IStatistics *stats
+	)
+{
+	GPOS_ASSERT(NULL == m_pstats);
+	GPOS_ASSERT(NULL != stats);
+	m_pstats = stats;
+}
+
+void
+CExpression::SetCost
+	(
+	CCost cost
+	)
+{
+	GPOS_ASSERT(m_cost == GPOPT_INVALID_COST);
+	m_cost = cost;
+}
+
+
 // EOF
--- a/libgpopt/src/operators/CExpressionPostProcessor.cpp
+++ b/libgpopt/src/operators/CExpressionPostProcessor.cpp
@@ -1,0 +1,129 @@
+//---------------------------------------------------------------------------
+//	Greenplum Database
+//	Copyright (C) 2018 Pivotal Inc.
+//
+//	@filename:
+//		CExpressionPostProcessor.cpp
+//
+//	@doc:
+//		Expression plan tree post processing routines
+//---------------------------------------------------------------------------
+
+#include "gpopt/operators/CExpressionPostProcessor.h"
+#include "gpos/memory/IMemoryPool.h"
+#include "gpos/common/CAutoTimer.h"
+#include "gpopt/operators/CPhysicalDML.h"
+#include "gpopt/base/CDistributionSpecRandom.h"
+#include "gpopt/operators/CPhysicalMotionRandom.h"
+
+
+using namespace gpopt;
+
+CExpressionPostProcessor::CExpressionPostProcessor()
+{}
+
+// add a random motion before dml insert /ctas if the input tree does not have
+// random motion ensuring randomness
+CExpression *
+CExpressionPostProcessor::AddRandomMotionBeforeDMLInsertOrCTAS
+	(
+	IMemoryPool *mp,
+	CExpression *expr
+	)
+{
+	// applicable only for dml insert operation on randomly distributed table
+	if (expr->Pop()->Eopid() == COperator::EopPhysicalDML)
+	{
+		CPhysicalDML *phy_dml_op = CPhysicalDML::PopConvert(expr->Pop());
+		if (CLogicalDML::EdmlInsert == phy_dml_op->Edmlop() &&
+			IMDRelation::EreldistrRandom == phy_dml_op->Ptabdesc()->Ereldistribution() &&
+			IsRandomMotionRequired(expr))
+		{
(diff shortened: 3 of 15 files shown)


libgpopt/src/operators/CExpressionPlanPostProcessor.cpp
```diff
@@ -0,0 +1,123 @@
...
+// random motion ensuring randomness
+CExpression *
+CExpressionPlanPostProcessor::AddRandomMotionBeforeDMLInsertOrCTAS
+	(
+	IMemoryPool *mp,
+	CExpression *expr
+	)
+{
+	// applicable only for dml insert operation on randomly distributed table
+	if (expr->Pop()->Eopid() == COperator::EopPhysicalDML)
```
vraghavan78: `if (COperator::EopPhysicalDML == expr->Pop()->Eopid())`

libgpopt/src/operators/CExpressionPlanPostProcessor.cpp
```diff
@@ -0,0 +1,123 @@
...
+	{
+		CPhysicalDML *phy_dml_op = CPhysicalDML::PopConvert(expr->Pop());
+		if (phy_dml_op->Edmlop() == CLogicalDML::EdmlInsert &&
+			IMDRelation::EreldistrRandom == phy_dml_op->Ptabdesc()->Ereldistribution() &&
+			IsRandomMotionRequired(expr))
+		{
+			// create a new spec and a random motion operator
```
vraghavan78: spec?

libgpopt/src/operators/CExpressionPlanPostProcessor.cpp
```diff
@@ -0,0 +1,123 @@
...
+			CDistributionSpecRandom *new_random_spec = GPOS_NEW(mp) CDistributionSpecRandom();
```
vraghavan78: `random_spec_new` I like that better. @hardikar what is our new convention?

libgpopt/src/operators/CExpressionPlanPostProcessor.cpp
```diff
@@ -0,0 +1,123 @@
...
+			CPhysicalMotionRandom *new_random_motion_op = GPOS_NEW(mp) CPhysicalMotionRandom(mp, new_random_spec);
```
vraghavan78: same here `random_motion_op_new`

libgpopt/src/operators/CExpressionPlanPostProcessor.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+			// create a new random motion expression with the child of dml expression
+			CExpression *sub_tree_expr = (*expr)[0]; // child of dml expression
```
vraghavan78: why no new here. Either have it or not. `sub_tree_expr_new`

libgpopt/src/operators/CExpressionPlanPostProcessor.cpp
```diff
@@ -0,0 +1,123 @@
...
+			sub_tree_expr->AddRef();
+			CExpression *added_random_motion_expr = GPOS_NEW(mp) CExpression(mp, new_random_motion_op, sub_tree_expr);
+			added_random_motion_expr->CopyExpressionProp(sub_tree_expr);
+
+			// create a new dml expression with the new random motion expression
+			phy_dml_op->AddRef();
+			CExpression *new_phy_dml_expr = GPOS_NEW(mp) CExpression(mp, phy_dml_op, added_random_motion_expr);
+			new_phy_dml_expr->CopyExpressionProp(expr);
+
```
vraghavan78: what happens to the old DML node? do we have a leak?

vraghavan78: Ah.. I see you release in the caller.

libgpopt/src/operators/CExpressionPlanPostProcessor.cpp
```diff
@@ -0,0 +1,123 @@
...
+	(
+	CExpression *expr
+	)
+{
+	// protect against stack overflow during recursion
+	GPOS_CHECK_STACK_SIZE;
+	GPOS_ASSERT(NULL != expr);
+
+	// if a random motion exists, exit early as the existing random motion can tell
+	// if a new random motion is required or not
```
vraghavan78: to me this sentence is hard to understand. What is it trying to say?

libgpopt/src/operators/CExpressionPlanPostProcessor.cpp
```diff
@@ -0,0 +1,123 @@
...
+	if (expr->Pop()->Eopid() == COperator::EopPhysicalMotionRandom)
```
vraghavan78: `if ( COperator::EopPhysicalMotionRandom == expr->Pop()->Eopid())`

libgpopt/src/operators/CExpressionPlanPostProcessor.cpp
```diff
@@ -0,0 +1,123 @@
...
+	{
+		return CUtils::FDuplicateHazardMotion(expr);
+	}
+
+	// if the dml insert/ctas is into a randomly distributed table from child which provides random
```
vraghavan78: Consider a dml insert/ctas on a randomly distributed table, and the child of the DML node is provides a random distribution spec. In this scenario, we are guaranteed that there no random motion in the subtree and there is a possibility that the tuples are skewed when the plan is executed. To avoid this we should force a random motion to spray the tuple on all segments.

Is this what you are trying to say? Can you build on the above?

libgpopt/src/optimizer/COptimizer.cpp
```diff
@@ -269,10 +271,14 @@ COptimizer::PdxlnOptimize
 			CExpression *pexprPlan = PexprOptimize(pmp, pqc, pdrgpss);
 			GPOS_CHECK_ABORT;
 
-			PrintQueryOrPlan(pmp, pexprPlan);
+			// post process the plan expression
+			CExpressionPlanPostProcessor expr_plan_post_processor;
+			CExpression *post_processsed_expr_plan = expr_plan_post_processor.PostProcessPlan(pmp, pexprPlan);
```
vraghavan78: `post_processsed_plan` suffices.

bhuvnesh2703:
@vraghavan78 thanks for review.. applied feedback

libgpopt/src/operators/CExpressionPlanPostProcessor.cpp
hardikar: new_random_spec is what I've used mostly.

libgpopt/src/operators/CExpressionPlanPostProcessor.cpp
hardikar: Is it possible to release it here?

libgpopt/include/gpopt/operators/CExpression.h
```diff
@@ -320,6 +321,24 @@ namespace gpopt
...
+			void SetCost(CCost cost);
+
+			void SetReqdPropPlan(CReqdPropPlan *reqd_prop_plan);
+
+			void SetDrvdPropRelational(CDrvdPropRelational *drvd_prop_relational);
+
+			// copy relational, derived, required, statistics and cost of input expression
+			void
+			CopyExpressionProp(CExpression *expr);
+
```
hardikar: 1. I do not like these SetCost, SetStats etc methods added here (especially since they are public). They are only used in `CopyExpressionProp`, which is has access to the private members anyway. So I don't think these getters/setters are really needed.
2. Shouldn't we actually *derive* the properties rather than simply copying them? Or if you want to clone the entire expression, we can just a method to do that!

hardikar:
Thanks for this PR. I MUCH prefer it here than hidden somewhere in the Translator!

libgpopt/src/operators/CExpressionPostProcessor.cpp
```diff
@@ -0,0 +1,129 @@
...
+		{
+			// create a new spec and a random motion operator
+			CDistributionSpecRandom *random_spec_new = GPOS_NEW(mp) CDistributionSpecRandom();
+			CPhysicalMotionRandom *new_random_motion_op = GPOS_NEW(mp) CPhysicalMotionRandom(mp, random_spec_new);
+
+			// create a new random motion expression with the child of dml expression
+			CExpression *sub_tree_expr = (*expr)[0]; // child of dml expression
+			sub_tree_expr->AddRef();
+			CExpression *random_motion_expr_new = GPOS_NEW(mp) CExpression(mp, new_random_motion_op, sub_tree_expr);
+			random_motion_expr_new->CopyExpressionProp(sub_tree_expr);
```
hardikar: Can't we derive it here instead?

libgpopt/src/optimizer/COptimizer.cpp
```diff
@@ -269,10 +271,14 @@ COptimizer::PdxlnOptimize
 			CExpression *pexprPlan = PexprOptimize(pmp, pqc, pdrgpss);
 			GPOS_CHECK_ABORT;
 
-			PrintQueryOrPlan(pmp, pexprPlan);
+			// post process the plan expression
+			CExpressionPostProcessor plan_post_processor;
```
hardikar: If the only purpose of this object is to have a call to PostProcessPlan, let's make that a static method.
Similar to the preprocessor.
```
                         // main driver
			static
			CExpression *PexprPreprocess
```

hardikar: Just noticed it is already a static method. The call should be changed to ` CExpressionPostProcessor:: PostProcessPlan()` instead then.

bhuvnesh2703: oh yes, will make the change.

libgpopt/src/operators/CExpression.cpp
```diff
@@ -1536,4 +1536,97 @@ CExpression::FValidPartEnforcers
...
+{
+	CDrvdPropPlan *drvd_prop_plan = expr->GetDrvdPropPlan();
+	drvd_prop_plan->AddRef();
+	this->SetDrvdPropPlan(drvd_prop_plan);
+
+	CDrvdPropRelational *drvd_prop_relational = expr->GetDrvdPropRelational();
+	drvd_prop_relational->AddRef();
+	this->SetDrvdPropRelational(drvd_prop_relational);
+
+	CReqdPropPlan *reqd_prop_plan = expr->Prpp();
```
vraghavan78: Why do we need this?

bhuvnesh2703:
Closed.
Opened a new PR: https://github.com/greenplum-db/gporca/pull/395

PR #400, opened 2018-08-28
Optimize DML queries by enforcing a non-master gather motion

dhanashreek89:
When a DML query contains operator such as Limit or an Aggregate, such operator requested singleton-master distribution spec which led to a gather motion on master. For instance:
```
create table test1 ( a int, b int);
insert into test1 select
generate_series(1,100000),generate_series(1,100000);
```
```
create table test as select * from test1 limit 10000000;

+--CPhysicalDML (Insert, "fake ctas rel")
	   +--CPhysicalComputeScalar
	      |--CPhysicalMotionRandom
	      |  +--CPhysicalLimit <empty> global
	      |     |--CPhysicalMotionGather(master)
	      |     |  +--CPhysicalTableScan "test1" ("test1")
	      |     |--CScalarConst (0)
	      |     +--CScalarConst (100000)
	      +--CScalarProjectList
		 +--CScalarProjectElement "ColRef_0009" (9)
		    +--CScalarConst (1)
```
This behavior has a potential to cause resource contention on a master node when the workload contains lot of such concurrent queries.  Unlike SELECT statements, DML queries do not have a Singleton distribution requirement, hence by default we start with `CDistributionSpecAny()` requirement at the top level. As DML query can be executed completely without master (with one exception described later), this commit optimizes such queries by enforcing a non-master gather motion whenever a single host execution is needed.

All the derive() and required() distribution spec routines now will produce a singleton segment when optimizing a DML query. This optimization is not performed if -
(1) DML query that contains a master-only table such as below : 
```
set optimizer_enable_master_only_queries = on;
delete from x where exists (select port from gp_segment_configuration limit 10);
```
In such scenarios, we refrain from producing singleton segment distribution spec as the tree containing master-only table can only be executed on a master node.

(2) DML query that contains a function (STABLE or VOLATILE) which performs a SQL access. We can not determine if the SQL access is performed on a master-only table; hence we can not enforce non-master gather safely.

For volatile functions with no SQL access, such as random(), we will produce singleton-segment spec and enforce non-master gather.


libgpopt/include/gpopt/operators/CExpressionHandle.h
```diff
@@ -243,10 +243,10 @@ namespace gpopt
 				return (0 < GetRelationalProperties()->PcrsOuter()->Size());
 			}
 
-			// check if attached expression must execute on the master
-			BOOL FMasterOnly() const
+			// check if attached expression must execute on a single host
+			BOOL SingleHostOnly() const
```
vraghavan78: isn't that singleton? How about `SingletonExecution`?

libgpopt/include/gpopt/operators/CPhysical.h
```diff
@@ -382,14 +382,18 @@ namespace gpopt
 
 			// enforce an operator to be executed on the master
```
vraghavan78: fix comment to master of singleton segment

libgpopt/src/operators/CLogicalSequenceProject.cpp
```diff
@@ -479,7 +479,14 @@ CLogicalSequenceProject::PopRemoveLocalOuterRefs
 		if (NULL == pds)
 		{
 			// eliminate Partition clause
```
vraghavan78: confused by this comment, can you elaborate?

libgpopt/src/operators/CPhysical.cpp
```diff
@@ -392,10 +392,10 @@ CPhysical::PdsMasterOnlyOrReplicated
 {
 	GPOS_ASSERT(2 > ulOptReq);
 
-	// if expression has to execute on master then we need a gather
-	if (exprhdl.FMasterOnly())
+	// if expression has to execute a single host then we need a gather
```
vraghavan78: a gather *motion*

libgpopt/src/operators/CPhysicalHashJoin.cpp
```diff
@@ -409,12 +409,20 @@ CPhysicalHashJoin::PdsRequiredSingleton
 
 	if (CDistributionSpec::EdtUniversal == pdsFirst->Edt())
 	{
-		// first child is universal, request second child to execute on the master to avoid duplicates
-		return GPOS_NEW(mp) CDistributionSpecSingleton(CDistributionSpecSingleton::EstMaster);
+		// first child is universal, request second child to execute on a single host to avoid duplicates
+		return GenerateSingletonSpec(mp);
+	}
+
+	if(COptCtxt::PoctxtFromTLS()->DMLQueryWithoutMasterOnlyTables() &&
```
vraghavan78: if (

libgpopt/src/operators/CPhysicalHashJoin.cpp
```diff
@@ -409,12 +409,20 @@ CPhysicalHashJoin::PdsRequiredSingleton
...
+	   CDistributionSpec::EdtReplicated == pdsFirst->Edt())
+	{
+		// query is DML with no master-only tables
+		// and the first child is replicated, require singleton-segment on the second child
```
vraghavan78: This seems suboptimal, what if the second child is hashed, we will now gather?

vraghavan78: Can you think about what could go wrong (performance wise) with this logic ?

dhanashreek89: I believe that was the case even before this commit. `PdsRequiredSingleton()` is responsible for generating a singleton execution alternative for HJ. Hence an alternative with gather motion on both the children will be produced (with an exception of replicated or master-only child). Please see below alternative without this fix:
```
create table foo (a int) distributed by(a);
create table bar (a int) distributed by(a);
set optimizer_plan_id=1;

Physical plan:
+--CPhysicalInnerHashJoin   rows:1   width:68  rebinds:1   cost:862.000537
   |--CPhysicalMotionGather(master)   rows:1   width:34  rebinds:1   cost:431.000071
   |  +--CPhysicalTableScan "bar" ("bar")   rows:1   width:34  rebinds:1   cost:431.000019
   |--CPhysicalMotionGather(master)   rows:1   width:34  rebinds:1   cost:431.000071
   |  +--CPhysicalTableScan "foo" ("foo")   rows:1   width:34  rebinds:1   cost:431.000019
   +--CScalarCmp (=)
      |--CScalarIdent "a" (0)
      +--CScalarIdent "a" (8)
```
If the second child is hashed, we need to gather it to a single host for singleton execution. Only difference after this commit, is, we will gather the hashed tree on a one segment rather than master.
Thoughts ?

server/src/unittest/gpopt/minidump/CDMLTest.cpp
```diff
@@ -74,6 +74,19 @@ const CHAR *rgszDMLFileNames[] =
 	"../data/dxl/minidump/UpdateZeroRows.mdp",
 	"../data/dxl/minidump/InsertNoEnforceConstraints.mdp",
 	"../data/dxl/minidump/UpdateNoEnforceConstraints.mdp",
+	"../data/dxl/minidump/Insert-With-HJ-CTE-Agg.mdp",
```
hardikar: Do these queries have counterparts in ICG? I ask because this change needs correctness tests as well. Especially for cases of volatile functions, TVFs etc.

hardikar:
Also I forgot to mention, Planner randomly decides which segment to use when it does the equivalent of Singleton(segment)... I think we should do the same. Otherwise we're just moving the execution from master to seg0.

bhuvnesh2703:
> Just leaving a note here of what we discussed. Ideally a single plan would do either Singleton(master) or Singleton(segment) and not mix and match. To achieve this maybe we should no long distinguish between the two during the optimization and let the translator decide it during translation?

I believe translation should just to translation, and not decide where the execution should happen. 
Thoughts on what advantage would be to decide in translator?

libgpopt/src/operators/CPhysicalComputeScalar.cpp
```diff
@@ -399,6 +399,10 @@ CPhysicalComputeScalar::PdsDerive
 	if (CDistributionSpec::EdtUniversal == pds->Edt() && 
 		IMDFunction::EfsVolatile == exprhdl.GetDrvdScalarProps(1 /*child_index*/)->Pfp()->Efs())
 	{
+		if (COptCtxt::PoctxtFromTLS()->DMLQueryWithoutMasterOnlyTables())
```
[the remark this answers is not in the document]
dhanashreek89: We can not since this function derives a `strict` singleton.

libgpopt/src/operators/CPhysical.cpp
```diff
@@ -435,8 +435,8 @@ CPhysical::PdsUnary
 	GPOS_ASSERT(0 == child_index);
 	GPOS_ASSERT(2 > ulOptReq);
 
-	// check if master-only/replicated distribution needs to be requested
-	CDistributionSpec *pds = PdsMasterOnlyOrReplicated(mp, exprhdl, pdsRequired, child_index, ulOptReq);
+	// check if singleton/replicated distribution needs to be requested
```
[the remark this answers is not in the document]
dhanashreek89: We `request` singleton/replicated bases on the expression properties. I renamed the function as `PdsRequireSingletonOrReplicated()`. 
I agree that we are inconsistent in terms of naming derive/require functions. I feel we should rename `PdsUnary()` to something like `PdsRequiredForUnaryOperator()`. Thoughts ?

dhanashreek89:
Corresponding GPDB PR https://github.com/greenplum-db/gpdb/pull/5759 that:
- Adds tests to ensure correctness
- randomly picks a segment to gather on
- Adds a guc to turn disable the traceflag

dhanashreek89:
@bhuvnesh2703 @hardikar @vraghavan78 Feedback applied. please take another look. Thanks,

PR #409, opened 2018-10-02
Remove Broadcast for Inner Nested Loop Join

hardikar:
This PR is a reopening of https://github.com/greenplum-db/gporca/pull/332; create to address gporca  issue: #309.

There are couple of plan changes, most of it seems reasonable to me. The plan size changes are expected with the way ORCA does plan enumeration. In short, ORCA deduplicate GroupExpressions in a group & Optimization contexts, but does no deduplicate optimization requests (optreqs). So if an operator generates the same optimization context for different optreqs it counts that as different plan - even though it produces the same plan. So the increase in the number of plans at this point does not correspond to the optimization time - there is a story in our backlog to fix this.

Performance:
I ran tpcds performance test on 5X_STABLE using the earlier version of this code. The performance was similar to 5X_STABLE proper, with no plan changes.

Also, the plan is to extend this to IndexNLJ in a future PR.

Corresponding GPDB PR : https://github.com/greenplum-db/gpdb/pull/5923


hardikar:
Corresponding ORCA PR : https://github.com/greenplum-db/gpdb/pull/5923

libgpopt/src/operators/CPhysicalInnerNLJoin.cpp
```diff
@@ -108,6 +111,50 @@ CPhysicalInnerNLJoin::PdsRequired
 
 	if (GPOS_FTRACE(EopttraceDisableReplicateInnerNLJOuterChild) || 0 == ulOptReq)
 	{
+		if (1 == child_index)
+		{
+			// compute a matching distribution based on derived distribution of outer child
+			CDistributionSpec *pdsOuter = CDrvdPropPlan::Pdpplan((*pdrgpdpCtxt)[0])->Pds();
```
bhuvnesh2703: Could we have the part here common with `CPhysicalScan::PdshashedDeriveWithOuterRefs` in a util function?

PR #414, opened 2018-10-11
 Remove broadcasts for Inner NLJ with INDF join predicates

hardikar:
This commit extends the feature in commit 26ad730 to also support INDF.

Corr. GPDB PR : https://github.com/greenplum-db/gpdb/pull/5970


libgpopt/src/base/CUtils.cpp
```diff
@@ -5160,15 +5160,24 @@ CUtils::PexprMatchEqualitySide
 	for (ULONG ul = 0; ul < ulSize; ul++)
 	{
 		CExpression *pexprPred = (*pdrgpexpr)[ul];
-		if (!CPredicateUtils::IsEqualityOp(pexprPred))
+		CExpression *pexprPredOuter, *pexprPredInner;
```
vraghavan78: please fix the comment and function name. It is misleading.

hardikar: You mean `PexprMatchEqualitySide`?

bhuvnesh2703:
Updated the DDL in commit `d9e69af`

PR #417, opened 2018-10-12
Fix cross join delay condition in Greedy algorithm

asubramanya:
Earlier, we were relying on the number of edges to identify if we have
potential opportunities to create joins avoiding cross joins.  However, we
should actually iterate over all the edges to find overlapping edges. If any
edge exists which is overlapping with the current join, then we should avoid
creating a cross join with other component which donot have an overlapping edge.


libgpopt/src/xforms/CJoinOrderGreedy.cpp
```diff
@@ -215,11 +215,37 @@ CJoinOrderGreedy::PexprExpand()
 
 			// combine component with current result and derive stats
 			CJoinOrder::SComponent *pcompTemp = PcompCombine(m_pcompResult, pcompCurrent);
-			if(CUtils::FCrossJoin(pcompTemp->m_pexpr) && 1 < m_ulNumUsedEdges)
```
hardikar: Please rename m_ulNumUsedEdges. It is confusing.

libgpopt/include/gpopt/xforms/CJoinOrderGreedy.h
```diff
@@ -35,7 +35,7 @@ namespace gpopt
 			// result component
 			SComponent *m_pcompResult;
 
-			ULONG m_ulNumUsedEdges;
+			ULONG m_ulNumEdgesRemaining;
```
vraghavan78: comment please

libgpopt/src/xforms/CJoinOrderGreedy.cpp
```diff
@@ -185,65 +184,67 @@ CJoinOrderGreedy::PexprExpand()
...
-		m_pcompResult->Release();
-		m_pcompResult = pcompBestResult;
+		else
+		{
+			// only cross joins are available. pick the unused component which will
+			// result in minimal cardinality
+			best_comp_idx = PickBestJoin(unused_components_set);
+			unused_components_set->ExchangeClear(best_comp_idx);
+		}
+		candidate_comp_set->Release();
```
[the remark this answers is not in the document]
asubramanya: Added assert.

libgpopt/src/xforms/CJoinOrder.cpp
```diff
@@ -301,8 +306,10 @@ CJoinOrder::ComputeEdgeCover()
 			CExpression *pexprComp = m_rgpcomp[ulComp]->m_pexpr;
 			CColRefSet *pcrsOutput = CDrvdPropRelational::GetRelationalProperties(pexprComp->PdpDerive())->PcrsOutput();
 
+
```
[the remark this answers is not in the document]
asubramanya: Updated comment.

libgpopt/src/xforms/CJoinOrderGreedy.cpp
```diff
@@ -185,65 +184,68 @@ CJoinOrderGreedy::PexprExpand()
...
 
 	if(NULL != m_pcompResult)
 	{
-		ulCoveredComps = 2;
+		// found atleast one non cross join
 		MarkUsedEdges();
 	}
 	else
 	{
+		// if every join combination is a cross join
```
hardikar:
```suggestion
		// every join combination is a cross join
```

libgpopt/src/xforms/CJoinOrderGreedy.cpp
```diff
@@ -185,65 +184,68 @@ CJoinOrderGreedy::PexprExpand()
...
 		m_pcompResult = GPOS_NEW(m_mp) SComponent(m_mp, NULL /*pexpr*/);
 	}
-
-	while (ulCoveredComps < m_ulComps)
+	
+	// create a bitset for all the unused components
+	CBitSet *unused_components_set = GPOS_NEW(m_mp) CBitSet(m_mp);
```
hardikar: Any reason why we can't use `m_ulNumEdgesRemaining`? If there is, add a comment here as to why.

libgpopt/src/xforms/CJoinOrderGreedy.cpp
```diff
@@ -185,65 +184,68 @@ CJoinOrderGreedy::PexprExpand()
...
-				continue;
-			}
+			unused_components_set->ExchangeSet(ul);
+		}
+	}
+	
+	while (unused_components_set->Size() > 0)
+	{
+		// get a set of components which can be joined with m_pcompResult
+		CBitSet *candidate_comp_set = GetJoinCandidateComponents();
```
hardikar: Rename suggestion: GetJoinCandidateComponents -> GetAdjacentComponentsToJoinCandidate. Or GetJoinCandidateAdjacentComponents.

libgpopt/src/xforms/CJoinOrderGreedy.cpp
iyerr3: `m_ulNumEdgesRemaining` was replaced with a set of unused components. Hence we don't need that member variable anymore and has been removed in the latest commit.

PR #419, opened 2018-10-17
Update pipeline file to publish ORCA 2X for GPDB4.3 builds

asubramanya:


concourse/pipeline.yml
```diff
@@ -202,112 +206,128 @@ jobs:
...
+      TASK_OS_VERSION: 6
+      TARGET_OS: centos
+      TARGET_OS_VERSION: 5
+  - task: compile_gpdb
+    file: gpdb_src/ci/concourse/compile_gpdb.yml
+    image: centos-gpdb-dev-5
+    params:
+      TARGET_OS: centos
+      TARGET_OS_VERSION: 5
+      BLD_TARGETS: clients connectivity loaders
```
bhuvnesh2703: can we just keep it empty? because it will increase the compile time and we don't really use clients, connectivity and loaders installer in icg i believe.
Also, they get tested in the main prod pipeline.

asubramanya: will do.

PR #421, opened 2018-10-25
Fix greedy bestjoin

bhuvnesh2703:
Fix Bug in Greedy Join Algo while picking best join

The goal of `PickBestJoin` is to find the best component out of all the
available component which gives the minimum cardinality with the existing
`m_pcompResult`. However, `m_pcompResult` was getting updated multiple 
times for the join candidate set.
`m_pcompResult` should only be updated once after it has iterated over all 
the candidate component and found the best  join combination.

Also, have enabled Greedy Algorithm in exhaustive mode as well.


libgpopt/src/xforms/CJoinOrderGreedy.cpp
```diff
@@ -265,36 +265,41 @@ CJoinOrderGreedy::PickBestJoin
...
 		// pick the component which will give the lowest cardinality
 		if (NULL == pcompBestComponent || dRows < dMinRows)
 		{
 			dMinRows = dRows;
 			best_comp_idx = iter.Bit();
 			pcompBestComponent = pcompCurrent;
-			m_pcompResult->Release();
-			m_pcompResult = pcompTemp;
-			m_pcompResult->AddRef();
+			pcompTemp->AddRef();
```
hardikar:
```
pcompTemp->AddRef();
CRefCount::SafeRelease(pcompBest);
```
This should probably go outside of the loop.

libgpopt/src/xforms/CXformExpandNAryJoinGreedy.cpp
```diff
@@ -69,21 +69,6 @@ CXformExpandNAryJoinGreedy::Exfp
 	)
 	const
 {
-	COptimizerConfig *poconf = COptCtxt::PoctxtFromTLS()->GetOptimizerConfig();
```
hardikar: Please separate this out into a different commit. This is a feature on its own.

libgpopt/src/xforms/CJoinOrderGreedy.cpp
bhuvnesh2703: As discussed this can't be moved outside.

libgpopt/src/xforms/CXformExpandNAryJoinGreedy.cpp
bhuvnesh2703: Made separate commits

libgpopt/src/xforms/CJoinOrderGreedy.cpp
iyerr3: Another option is to compute the best `pCompBestComponent` and then call `pcompCombine` a final time outside the loop to update `m_pcompResult`. This way the objective of the loop is to compute the best component and then we add that into our result. Not sure about the cost of making an extra `pcomCombine` call - if it's an expensive operation then we can stick with our current approach of maintaining `pcompBest`.

PR #423, opened 2018-11-08
 Add new interface for sharing printer functions

hardikar:
Contains 2 commits - they can be review independently (See individual comments for details).

This commit should have not impact on ORCA retail build.

### DIFF
--- a/libgpopt/src/base/CQueryContext.cpp
+++ b/libgpopt/src/base/CQueryContext.cpp
@@ -262,16 +262,6 @@ CQueryContext::PqcGenerate
 	return GPOS_NEW(mp) CQueryContext(mp, pexprResult, prpp, colref_array, pdrgpmdname, fDeriveStats);
 }
 
-#ifdef GPOS_DEBUG
-
-//---------------------------------------------------------------------------
-//	@function:
-//		CQueryContext::OsPrint
-//
-//	@doc:
-//		Debug print
-//
-//---------------------------------------------------------------------------
 IOstream &
 CQueryContext::OsPrint
 	(
@@ -281,15 +271,6 @@ CQueryContext::OsPrint
 {
 	return os << *m_pexpr << std::endl << *m_prpp;
 }
-
-void
-CQueryContext::DbgPrint() const
-{
-	CAutoTrace at(m_mp);
-	(void) this->OsPrint(at.Os());
-}
-#endif // GPOS_DEBUG
-
 
 //---------------------------------------------------------------------------
 //	@function:
--- a/libgpos/include/gpos/common/IPrinter.h
+++ b/libgpos/include/gpos/common/IPrinter.h
@@ -46,6 +46,7 @@ namespace gpos
 
 #ifdef GPOS_DEBUG
 			// debug print; for interactive debugging sessions only
+			// use `setting set escape-non-printables false` in lldb to see the string fully formatted
 			const WCHAR *DbgPrint() const;
 #endif  // GPOS_DEBUG
 
--- a/libgpos/include/gpos/common/IPrinter.h
+++ b/libgpos/include/gpos/common/IPrinter.h
@@ -1,0 +1,58 @@
+//---------------------------------------------------------------------------
+//	Greenplum Database
+//	Copyright (C) 2018 Pivotal Inc.
+//
+//	Interface for printing functions
+//
+//	TODO: Implement OsPrint() & DbgPrint() for all classes by inheriting from
+//	this class.
+//---------------------------------------------------------------------------
+#ifndef GPOS_IPrinter_H
+#define GPOS_IPrinter_H
+
+#include "gpos/types.h"
+#include "gpos/io/IOstream.h"
+
+namespace gpos
+{
+	class CWStringDynamic;
+
+	class IPrinter
+	{
+		private:
+
+			// private copy ctor
+			IPrinter(const IPrinter&);
+
+#ifdef GPOS_DEBUG
+			static
+			CWStringDynamic *m_msg_holder;
+#endif  // GPOS_DEBUG
+
+		public:
+
+			// ctor
+			IPrinter()
+			{}
+
+			// dtor
+			virtual
+			~IPrinter()
+			{}
+
+			// print driver
+			virtual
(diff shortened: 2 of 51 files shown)


kanishka-azimi:
I was casually skimming this PR as a gporca outsider. I read the commit messages. I'm curious about the overall context for when you realized this was needed.

libgpos/include/gpos/common/IPrinter.h
```diff
@@ -0,0 +1,58 @@
...
+	class CWStringDynamic;
+
+	class IPrinter
+	{
+		private:
+
+			// private copy ctor
+			IPrinter(const IPrinter&);
+
+#ifdef GPOS_DEBUG
```
bhuvnesh2703: do we need these GPOS_DEBUG ifdef? can't we have it available even for retail builds?

libgpos/include/gpos/common/IPrinter.h
```diff
@@ -0,0 +1,58 @@
...
+			// dtor
+			virtual
+			~IPrinter()
+			{}
+
+			// print driver
+			virtual
+			IOstream &OsPrint(IOstream &os) const = 0;
+
+#ifdef GPOS_DEBUG
```
bhuvnesh2703: same here?

hardikar:
@kanishka-azimi
Hey - I'm not sure I understand what you mean. This PR was to help us do 2 things:
1. Organize the DbgPrint() code in one place and reduce the amount of copy-pasted implementations thrown around the code.
2. ORCA has a good way of testing/debugging queries without the need for a running database by loading the mdp file format. But sometimes, a database is needed - e..g when we're checking for something database specific or when we need to use the executor for constant evaluation. In those cases DbgPrint() was pretty useless since the output would get buffered in an in-memory string that is dumped (only on error) at the end of the execution. This slowed down the debugging process a lot! This PR returns a string instead ~ so it is immediately visible in the debugging session.

libgpos/include/gpos/common/IPrinter.h
hardikar: Right - so I just wanted to reduce the risk. This PR is incomplete in some sense - not all DbgPrint()/OsPrint() methods have been moved. So I wanted to keep the change in functionality (especially w.r.t leaked memory in the string) only available in debug builds - because that's where it matters anyway.

kanishka-azimi:
Exactly what I was looking for, thanks.

libgpos/include/gpos/common/IPrinter.h
```diff
@@ -0,0 +1,58 @@
...
+			// debug print; for interactive debugging sessions only
```
hardikar:
```suggestion
			// debug print; for interactive debugging sessions only
			// use `setting set escape-non-printables false` in lldb to see the string fully formatted
```

hardikar:
@bhuvnesh2703 Added your comment.

libgpopt/src/base/CQueryContext.cpp
```diff
@@ -262,16 +262,6 @@ CQueryContext::PqcGenerate
...
-//---------------------------------------------------------------------------
-//	@function:
-//		CQueryContext::OsPrint
-//
-//	@doc:
-//		Debug print
-//
-//---------------------------------------------------------------------------
 IOstream &
 CQueryContext::OsPrint
```
chrishajas: Was it intentional to remove this from the debug only build?

zellerh:
This adds multiple inheritance, and @bhuvnesh2703 mentioned that this would be the first time we use multiple inheritance. In my humble opinion, multiple inheritance just doesn't work in C++, so if we could avoid it that would be great. My main issue with it is that casting a pointer to another class no longer works reliably, if we have forward class declarations.

As far as I know, the following may or may not work:

```
class A; // A is a class that inherits from multiple base classes

MyClass *x = some_function_call();

A *y = (A *) x;
```

Without multiple inheritance, the cast to A * doesn't change the actual address we point to, but if we have multiple inheritance, doing such a cast sometimes requires adding a constant to the pointer. However, if we just declared A as in the example above, without sourcing the header file for A, the compiler won't know whether adding a constant is needed or not. Unfortunately, there is no warning, the compiler may just do the wrong thing and the program will crash or corrupt memory.

My apologies if this is already well-known to you. Also, I haven't tested this, it's just what I heard from others many years ago.

hardikar:
@zellerh 
Right, but isn't that issue with really with multiple inheritance where both bases classes have data members? Since IPrinter is an interface with no data members, the address issue shouldn't matter, because there is no pointer modification needed at all. However, there is the risk someone might do that in the future.
I'd love any suggestion to implement this in any other way. I initially intended to have CColRef inherit IPrinter, but that's tricky because we can't use OsPrint everywhere yet (as described already). The only other alternative I can think of is to change the DbgPrint() implementations where they are on master - which would avoid the risk completely.

zellerh:
Hi @hardikar, it could be that this is restricted to cases where there are data members and it might be ok if the second base class doesn't have any data members. Googling this issue, I could not find a clear example that proves the particular case used here is an issue. Somehow I thought that virtual function pointers are also causing problems, and may also in some cases cause this issue, but I couldn't find an example.

@bhuvnesh2703 mentioned that the original problem had to do with const-ness of the base class. If that's the case, couldn't we just cast the const-ness of the base class away when we need to modify a data member? Does the version of C++ we use support the "mutable" keyword? If so, that might be another option. Sorry, I probably don't understand what the real issue is with the alternatives to multiple inheritance.

libgpos/src/common/IPrinter.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+// Manages the memory for debug printer string
+// NB: This is allocated the first time DbgPrint() is called; but never free'd
+// till the server is shut down. But since DbgPrint() is intended for use in
+// debug sessions only, this doesn't matter...
+CWStringDynamic *
+IPrinter::m_msg_holder = NULL;
+
+const WCHAR *
+IPrinter::DbgPrint() const
```
hardikar: Note here that it cannot be called recursively!

bhuvnesh2703:
Closing it for rethought.

PR #426, opened 2018-11-21
Make CXformSplitDQA always generate a two-stage agg

sambitesh:
When optimizer_force_three_stage_scalar_dqa is on. We expect the below:

- We expect ORCA to pick up three stage DQA.
- If three stage DQA is not picked because it couldn't be optimized, then pick
  at least two stage DQA.  because in case of 2 Stage Agg, one of the agg will
  be on segment and other will be on master after the gather in case of select
  queries.

Consider the case below:
```
create table foo (a int);
explain select count(distinct a) from foo;
```
Previously, this would give us a one stage agg because a three stage couldn't
be optimized and no 2-stage option was generated by CXformSplitDQA.  But now,
we make sure to generate both 2-stage aggs and 3-stage aggs in CXformSplitDQA.

For `explain select count(distinct a) from foo;`

Earlier Plan
```
			QUERY PLAN
----------------------------------------------------------------------------------------
 Aggregate  (cost=0.00..433.48 rows=1 width=8)
   ->  Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..433.43 rows=99984 width=4)
         ->  Table Scan on foo  (cost=0.00..431.70 rows=33328 width=4)
```

Present plan
```
			QUERY PLAN
------------------------------------------------------------------------------------
 Aggregate  (cost=0.00..431.96 rows=1 width=8)
   ->  Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..431.96 rows=1 width=8)
         ->  Aggregate  (cost=0.00..431.96 rows=1 width=8)
               ->  Table Scan on foo  (cost=0.00..431.70 rows=33331 width=4)
```
We also had to make sure that, when optimizer_force_three_stage_scalar_dqa is
on, pick 3-stage aggs.
For this we first tracked the 2-stage aggs and 3-stage aggs created by CXformSplitDQA.
In CConstContext::FbetterThan, we made sure 3-stage aggs from CXformSplitDQA
are marked as more optimal than 2-stage aggs from CXformSplitDQA.


libgpopt/src/xforms/CXformUtils.cpp
```diff
@@ -1850,6 +1850,21 @@ CXformUtils::FSplitAggXform
 		CXform::ExfSplitGbAggDedup == exfid;
 }
 
+BOOL
+CXformUtils::FSplitDQAAggXform
```
vraghavan78: FSplitDQAXform

libgpopt/src/xforms/CXformUtils.cpp
```diff
@@ -1850,6 +1850,21 @@ CXformUtils::FSplitAggXform
...
+	(
+	CExpression *pexprAgg
+	)
+{
+	CGroupExpression *pgexprOrigin = pexprAgg->Pgexpr();
+	if(NULL != pgexprOrigin)
```
vraghavan78: can it be NULL and why?

libgpopt/src/xforms/CXformSplitDQA.cpp
```diff
@@ -773,10 +785,17 @@ CXformSplitDQA::PexprMultiLevelAggregation
 	}
 
 	pdrgpcrLastStage->AddRef();
```
vraghavan78: name inconsistency, why LastStage and not ThreeStage. Pick one.

vraghavan78: may be ok

libgpopt/src/xforms/CXformSplitDQA.cpp
```diff
@@ -736,7 +747,8 @@ CXformSplitDQA::PexprMultiLevelAggregation
 									pdrgpcrLocal,
 									COperator::EgbaggtypeIntermediate,
 									false, /* fGeneratesDuplicates */
-									pdrgpcrArgDQA
+									pdrgpcrArgDQA,
+									true /* m_isAggThreeStage */
```
vraghavan78: how is this true. This is a two stage aggregate correct?

libgpopt/src/xforms/CXformSplitDQA.cpp
```diff
@@ -174,26 +174,24 @@ CXformSplitDQA::Transform
...
-				phmexprcr,
-				pdrgpcrArgDQA
-				);
-
-		pxfres->Add(pexprAlt2);
-	}
-
-	if (fScalarDQA && !fForce3StageScalarDQA) {
+	
+	// local/global for both scalar and non-scalar aggregates.
```
vraghavan78: change comment please very confusing.

libgpopt/src/xforms/CXformSplitDQA.cpp
```diff
@@ -174,26 +174,24 @@ CXformSplitDQA::Transform
...
+			phmexprcr,
+			pdrgpcrArgDQA
+			);
+
+	pxfres->Add(pexprAlt2);
+	
+
+	if (fScalarDQA)
+	{
 		// if only want 3-stage DQA then skip this 2-stage option for scalar DQA.
```
vraghavan78: remove extra line

libgpopt/src/xforms/CXformSplitDQA.cpp
```diff
@@ -174,26 +174,24 @@ CXformSplitDQA::Transform
...
```
vraghavan78: above comment confusing me

libgpopt/src/xforms/CXformSplitDQA.cpp
```diff
@@ -773,10 +785,17 @@ CXformSplitDQA::PexprMultiLevelAggregation
 	}
```
vraghavan78: so we exit early but label all the aggregates as three stage this is wrong!!!

libgpopt/src/operators/CPhysicalStreamAggDeduplicate.cpp
```diff
@@ -37,7 +37,18 @@ CPhysicalStreamAggDeduplicate::CPhysicalStreamAggDeduplicate
...
+	(
+	 mp,
+	 colref_array,
+	 pdrgpcrMinimal,
+	 egbaggtype,
+	 fGeneratesDuplicates,
+	 NULL /*pdrgpcrGbMinusDistinct*/,
+	 fMultiStage,
+	 false /* m_isAggFromSplitDQA */,
+	 false /* m_isAggThreeStage */
```
vraghavan78: I thought the logical operator should provide this, right? Why this is initialisation here?

libgpopt/src/operators/CPhysicalHashAggDeduplicate.cpp
```diff
@@ -39,7 +39,18 @@ CPhysicalHashAggDeduplicate::CPhysicalHashAggDeduplicate
...
+		(
+		 mp,
+		 colref_array,
+		 pdrgpcrMinimal,
+		 egbaggtype,
+		 fGeneratesDuplicates,
+		 NULL /*pdrgpcrGbMinusDistinct*/,
+		 fMultiStage,
+		 false /* m_isAggFromSplitDQA */,
+		 false /* m_isAggThreeStage */
```
vraghavan78: why this is initialisation? Shouldn't the logical operator provide u this?

libgpopt/src/base/CCostContext.cpp
```diff
@@ -476,7 +502,37 @@ CCostContext::FBetterThan
...
+CCostContext::FAggTwoStageSplitDQA
+	(
+	const CCostContext *pcc
+	)
+	const
+{
+	if(CUtils::FPhysicalAgg(pcc->Pgexpr()->Pop()))
+	{
+		CPhysicalAgg *popAgg = CPhysicalAgg::PopConvert(pcc->Pgexpr()->Pop());
+		return (popAgg->GetIsAggFromSplitDQA() && !popAgg->GetIsAggThreeStage());
```
vraghavan78: question, for non-scalar aggregate like `select count(distinct a) from foo group by b;` what happens. The plan will be a three stage aggregate, and a generated by CXformSplitDQA. 

The trace flag `EopttraceForceThreeStageScalarDQA` only applies to scalar DQA and not general DQA.

vraghavan78:
I am have reservations about the code as it stands:

* there is a lot of assumptions when the 3-stage flag is set to true, 
* we do not track that the 3-stage agg was trying to optimise scalar dqa
* physical aggregate constructors should not initialise things, they should rely on the logical operator to set things

The core idea is there, just needs some clean up.

vraghavan78:
We should instead have the ENUM stating single stage, two stage and three stage. Then the function IsMultiStage will return true for 2/3 stage.

In addition we should have a flag that states if the aggregate operators from a scalar dqa.

That way you don’t need three values but 2 instead.

libgpopt/src/base/CCostContext.cpp
chrishajas: The GUC name is confusing. The existing code in master affected both Scalar and Non-scalar DQAs using the flag `EopttraceForceThreeStageScalarDQA`



Refer to :

https://github.com/greenplum-db/gporca/blob/6e567f29621a651e266eb3593ed26e7d782a6e66/libgpopt/src/xforms/CXformSplitDQA.cpp#L178

https://github.com/greenplum-db/gporca/blob/6e567f29621a651e266eb3593ed26e7d782a6e66/libgpopt/src/xforms/CXformSplitDQA.cpp#L196


Hence in our fix, we were handling both scalar and non-scalar DQA too. We'll change it to only handle scalar DQAs.

libgpopt/src/xforms/CXformSplitDQA.cpp
chrishajas: If `fSplit2LevelsOnly` then the three stage aggregates are not created. Refer to below lines.

https://github.com/greenplum-db/gporca/blob/6e567f29621a651e266eb3593ed26e7d782a6e66/libgpopt/src/xforms/CXformSplitDQA.cpp#L710

https://github.com/greenplum-db/gporca/blob/6e567f29621a651e266eb3593ed26e7d782a6e66/libgpopt/src/xforms/CXformSplitDQA.cpp#L728

libgpopt/src/xforms/CXformUtils.cpp
chrishajas: We can't think of a scenario where this can be NULL. However, since m_pgexpr is null in `CXformTest`, the test failed with a segmentation fault.

We referred to `CXformUtils::FMultiStageAgg`, it had a null check for `m_pgpexpr` too. So we added a check here.

https://github.com/greenplum-db/gporca/blob/6e567f29621a651e266eb3593ed26e7d782a6e66/libgpopt/src/xforms/CXformUtils.cpp#L1877

bhuvnesh2703: i think the root group expression will have null origin..

bhuvnesh2703: pgexprOrigin should be not null when it comes here..

libgpopt/src/base/CCostContext.cpp
```diff
@@ -410,6 +411,33 @@ CCostContext::FBetterThan
...
+	if (GPOS_FTRACE(EopttraceForceThreeStageScalarDQA))
+	{
+		if (CUtils::FPhysicalAgg(Pgexpr()->Pop()) && CUtils::FPhysicalAgg(pcc->Pgexpr()->Pop()))
+		{
+			// we are only interested in aggs generated by CXformSplitDQA. If the trace flag is turned on
+			// we want to favor 3-stage agg over 2-stage scalar DQA agg. So here we mark 2-stage scalar DQA as less optimal.
+			// single stage agg do not get optimized when multi-stage aggs are present. Refer to COptimizationContext::FOptimizeAgg
+			// so even if mark 2-stage scalar DQA aggs as less optimal cost context, it will never lead to us picking a single stage agg.
+			// note: we are only concerned about 2-stage scalar DQA and NOT 2-stage non-scalar DQA.
+			if (IsTwoStageScalarDQA(this) && IsSplitDQA(pcc) && !IsTwoStageScalarDQA(pcc))
```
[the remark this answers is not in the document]
chrishajas: Since we are explicitly checking for two-stage scalar aggs and three-stage scalar aggs, we don't need an assert here.

libgpopt/src/xforms/CXformSplitDQA.cpp
```diff
@@ -167,38 +170,28 @@ CXformSplitDQA::Transform
...
+		// this transform is useful for cases where distinct column is same as distributed column.
+		// for a query like "select count(distinct a) from bar;"
+		// we generate a two stage agg where the aggregate operator gives us the distinct values
+
+		//		+--CLogicalGbAgg( Global ) Grp Cols: [][Global]
+		//		|--CLogicalGbAgg( Local ) Grp Cols: ["a" (0)][Local],
+		//		|  |--CLogicalGet "bar" ("bar"),
+		//		|  +--CScalarProjectList
+		//		+--CScalarProjectList
+		//		+--CScalarProjectElement "count" (9)
```
iyerr3: Minor indentation for the ProjectElement

libgpopt/src/xforms/CXformSplitDQA.cpp
```diff
@@ -167,38 +170,28 @@ CXformSplitDQA::Transform
...
-	if (fScalarDQA && !fForce3StageScalarDQA) {
-		// if only want 3-stage DQA then skip this 2-stage option for scalar DQA.
-
-		// special case for 'scalar DQA' only, transform to 2-stage aggregate.
-		// It's beneficial for distinct column same as distributed column.
-		CExpression *pexprAlt3 = PexprSplitHelper
+	if (fScalarDQA)
+	{
+		// generate two-stage agg for scalar DQA case
+		// this transform is useful for cases where distinct column is same as distributed column.
```
iyerr3: Please add more info about this plan getting rejected downstream when distinct column != distributed column (as we discussed offline). Also please add explanation that the `CScalarProjectList` for the `Local`  agg below is empty on purpose.

sambitesh:
Merged

zellerh:
Looks good to me as well, based on my limited understanding.

PR #427, opened 2018-12-01
Push filter to outer child of Semi Join

asubramanya:
If a subquery had a predicate on the outer table, that predicate would not get
pushed to the outer child of the semi join. This was primarily because we were
not normalizing the expression after generating the CLogicalLeftSemiJoin
alternative. This patch fixes the issue by normalizing the expression while
decorrelating the subquery.

Plan before the patch:
```
explain select * from bar where exists (select 1 from foo where foo.a = bar.a and bar.a > 1);
                                  QUERY PLAN
-------------------------------------------------------------------------------
 Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..862.00 rows=1 width=4)
   ->  Hash Semi Join  (cost=0.00..862.00 rows=1 width=4)
         Hash Cond: (bar.a = foo.a)
         Join Filter: (bar.a > 1)
         ->  Table Scan on bar  (cost=0.00..431.00 rows=1 width=4)
               ->  Result  (cost=0.00..431.00 rows=1 width=4)
                           Filter: (a > 1)
```

Plan after the patch:
```
explain select * from bar1 bar where exists (select 1 from foo where foo.a = bar.a and bar.a > 1);
                                  QUERY PLAN
-------------------------------------------------------------------------------
 Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..862.00 rows=1 width=4)
   ->  Hash Semi Join  (cost=0.00..862.00 rows=1 width=4)
         Hash Cond: (bar.a = foo.a)
         ->  Table Scan on bar  (cost=0.00..431.00 rows=1 width=4)
               Filter: (a > 1)
               ->  Result  (cost=0.00..431.00 rows=1 width=4)
                           Filter: (a > 1)
```


libgpopt/src/operators/CNormalizer.cpp
```diff
@@ -47,6 +47,7 @@ const CNormalizer::SPushThru CNormalizer::m_rgpt[] =
 	{COperator::COperator::EopLogicalLeftAntiSemiApply, PushThruJoin},
 	{COperator::COperator::EopLogicalLeftAntiSemiApplyNotIn, PushThruJoin},
 	{COperator::COperator::EopLogicalLeftAntiSemiCorrelatedApplyNotIn, PushThruJoin},
+	{COperator::EopLogicalLeftSemiJoin, PushThruJoin},
```
bhuvnesh2703: can you also fix the above `COperator::COperator::` double qualification?

asubramanya: fixed fouble qualification

vraghavan78:
Nice work!

hardikar:
LGTM.
I wonder if this logic applies to all the other transforms. I'm asking if we should normalize all the results return by any transform in `CGroupExpression::Transform()`.

bhuvnesh2703:
The purpose of Normalize is the push down the filters if possible, the list of operators for which it supports push down is listed in the CNormalizer.cpp.

> It seems that we run PexprNormalize() in quite a few different places, and there's possibly/probably other places we're missing. Is PexprNormalize() expensive? Theoretically, is this something we would want to run after every transform if it took zero time?

So, definitely, i think it could be applied at other places too. However, before we apply it to every other alternative, we should find out queries where it is not happening currently. It should not be a zero time, but not a very heavy function compared to the benefits it will reap by pushing it down. If it turns out, that we are not pushing predicates at lot of places, then we should plan to refactor the code to perform it once. However, we are yet to find such cases. One of them we found was when we recently put a fix for converting IN to Exists and had to put Normalize there. I will create a story to see if we can move all the normalize to the end of the Expression Preprocessing. So, hopefully that will address this concern, and it will require some analysis / refactoring, so i think we should keep it as a separate task.

> Additionally, since this code is in Decorrelate(), it will also end up normalizing expressions returned from other transforms such as CXformInnerApply2InnerJoin. Is this intentional?

The alternative which it supports is listed in CNormalizer.cpp, so if the tree produced has supported operators it will attempt to push them down. But will not impact the operators which does not have PushThru methods defined for them.

There were a few queries which @asubramanya @iyerr3 found where we could generate additional predicates, however, that is a different story. That fix goes in how we generate implied predicates in `PexprInferPredicates`

iyerr3:
Merged with dd1e56be27c83ad103e8a9f461d3ce03254edbbd

PR #429, opened 2018-12-06
 Removing unnecessary motion while updating a table

sambitesh:
We achieve this by fixing the `PdsDerive` of `CPhysicalSplit`.

Consider the example below.

```
CREATE TABLE test
(
 a CHARACTER VARYING(10),
 b CHARACTER VARYING(10)
)
DISTRIBUTED BY (b);

EXPLAIN UPDATE test SET a = b;
```

Earlier we would get a redundant redistribute motion like below:

```
 Update  (cost=0.00..431.07 rows=1 width=1)
   ->  Result  (cost=0.00..431.00 rows=1 width=26)
         ->  Redistribute Motion 3:3  (slice1; segments: 3)  (cost=0.00..431.00 rows=1 width=22)
               Hash Key: public.test.b
               ->  Result  (cost=0.00..431.00 rows=1 width=22)
                     ->  Split  (cost=0.00..431.00 rows=1 width=22)
                           ->  Table Scan on test  (cost=0.00..431.00 rows=1 width=28)
```
After our fix, the redundant motion is gone

```
 Update  (cost=0.00..431.07 rows=1 width=1)
   ->  Result  (cost=0.00..431.00 rows=1 width=26)
         ->  Split  (cost=0.00..431.00 rows=1 width=22)
               ->  Table Scan on test  (cost=0.00..431.00 rows=1 width=28)
```


libgpopt/src/operators/CPhysicalSplit.cpp
```diff
@@ -356,16 +356,36 @@ CPhysicalSplit::PdsDerive
...
 		CColRef *new_colref = (*m_pdrgpcrInsert)[ul];
 		if (pcrOld != new_colref)
 		{
-			pcrsModified->Include(pcrOld);
-			pcrsModified->Include(new_colref);
+			// if delete column and insert column belong to same table
+			// then we care only care about whether the delete column is same as distribution column
+			// of the outer child of CPhysicalSplit. If the insert column is in outer child's
+			// derived hash distribution column/s, then data is alread colocated and we don't need a motion.
+			if(pcrsDelete->FMember(new_colref))
```
bhuvnesh2703: is this understanding correct?

create table s (a int, b int) distributed by (a);
Consider 0, 1 represent a, b respectively.
Case1:
update s set a = b; 

So, `insert array = {0,1}
delete array = {1,1};`

Here in pcrsModified, you will include `0`.
pcrsHashed will be based on `0`

So they are not disjoint, and we will return a random spec, this random spec will not satisfy the spec requested by CPhysicalDML which should ask the child to be distributed by column a, and we will see a redistribute motion in this case.

For Case2:
update s set b = a; 
So, `insert array = {0,1}
delete array = {0,0};`
pcrsModified, you will include `1`.
pcrsHashed will be based on `0`
So they are disjoint, and we will return pdsHashed, this will satisfy the spec requested by CPhysicalDML (which should ask the child to be distributed by column a), and we will not see a redistribute motion.

Also, in the comment then `then we care only care about whether the delete column is same as distribution column of the outer child of CPhysicalSplit.` is correct, but it does not indicate why you include only `pcrOld`. Is it because, that if you include the colref `1` (consider case 1) then pcrsModified will not be disjoint with pcrsHashed and you will return randomspec, which will cause a redistribute motion to be inserted because CPhysicalDML must have requested Hash Distribution. Also, why is this different then the case where you have queries were the insert column is coming from some other table?

libgpopt/src/operators/CPhysicalSplit.cpp
```diff
@@ -356,16 +356,36 @@ CPhysicalSplit::PdsDerive
...
 	for (ULONG ul = 0; ul < num_cols; ul++)
 	{
 		CColRef *pcrOld = (*m_pdrgpcrDelete)[ul];
 		CColRef *new_colref = (*m_pdrgpcrInsert)[ul];
 		if (pcrOld != new_colref)
 		{
-			pcrsModified->Include(pcrOld);
-			pcrsModified->Include(new_colref);
+			// if delete column and insert column belong to same table
+			// then we care only care about whether the delete column is same as distribution column
```
bhuvnesh2703: `care only care `

libgpopt/src/operators/CPhysicalSplit.cpp
```diff
@@ -356,16 +356,36 @@ CPhysicalSplit::PdsDerive
...
+			// of the outer child of CPhysicalSplit. If the insert column is in outer child's
+			// derived hash distribution column/s, then data is alread colocated and we don't need a motion.
```
bhuvnesh2703: `alread`

bhuvnesh2703:
I believe, if we revert the optimizer=off setting in test added in https://github.com/greenplum-db/gpdb/pull/6245/files, that should work now?

libgpopt/src/operators/CPhysicalSplit.cpp
```diff
@@ -356,16 +356,36 @@ CPhysicalSplit::PdsDerive
...
+	}
+
 	for (ULONG ul = 0; ul < num_cols; ul++)
 	{
 		CColRef *pcrOld = (*m_pdrgpcrDelete)[ul];
 		CColRef *new_colref = (*m_pdrgpcrInsert)[ul];
 		if (pcrOld != new_colref)
 		{
-			pcrsModified->Include(pcrOld);
-			pcrsModified->Include(new_colref);
```
zellerh: My guess would have been that a one-line fix might have worked, by just deleting line 368 from the old code? I would have thought that we don't care where the new value comes from, all we care about is that it changed? So, whether we say "set a=b" or "set a=10" doesn't really matter, in both cases what matters is only that we changed column a?

libgpopt/src/operators/CPhysicalSplit.cpp
```diff
@@ -356,16 +356,36 @@ CPhysicalSplit::PdsDerive
...
+			if(pcrsDelete->FMember(new_colref))
+			{
+				pcrsModified->Include(pcrOld);
+			}
+			// if delete column and insert column belong to different tables, then we have to make sure
+			// none of them match the derived distribution column of outer child of CPhysicalSplit.
+			else
+			{
+				pcrsModified->Include(pcrOld);
+				pcrsModified->Include(new_colref);
```
zellerh: If new_colref is not a member of pcrsDelete, can it really be a member of pcrsHashed?

libgpopt/src/operators/CPhysicalSplit.cpp
sambitesh: `Also, why is this different then the case where you have queries were the insert column is coming from some other table?`

For the update to take place we need to make sure that both the column to be deleted and column from which we are inserting is colocated. Hence we make sure to take care the derived distribution of child should not match either the column being deleted and the column from where we are inserting. 

This issue becomes moot when both are from same table.

libgpopt/src/operators/CPhysicalSplit.cpp
sambitesh: Take the case below

```
create table r (a float4, b float4) distributed by (a);
create table s (a float4, b float4) distributed by (a);
insert into r select generate_series(1, 10000), generate_series(1, 10000) * 3;
insert into s select generate_series(1, 100), generate_series(1, 100) * 4;
explain update s set a = r.a from r where r.b = s.b;
```

This will give us a plan like below

```
QUERY PLAN
------------------------------------------------------------------------------------------------------------------------------
 Update  (cost=0.00..869.58 rows=34 width=1)
   ->  Result  (cost=0.00..862.81 rows=67 width=26)
         ->  Redistribute Motion 3:3  (slice2; segments: 3)  (cost=0.00..862.81 rows=67 width=22)
               Hash Key: r.a
               ->  Result  (cost=0.00..862.80 rows=67 width=22)
                     ->  Split  (cost=0.00..862.80 rows=67 width=22)
                           ->  Hash Join  (cost=0.00..862.80 rows=34 width=22)
                                 Hash Cond: r.b = public.s.b
                                 ->  Table Scan on r  (cost=0.00..431.07 rows=3334 width=8)
                                 ->  Hash  (cost=431.03..431.03 rows=100 width=18)
                                       ->  Broadcast Motion 3:3  (slice1; segments: 3)  (cost=0.00..431.03 rows=100 width=18)
                                             ->  Table Scan on s  (cost=0.00..431.00 rows=34 width=18)
 Optimizer status: PQO version 3.14.0
(13 rows)
```

Here the `CPhysicalSplit` has a join below it. So pcrsHashed will actually be r.a in this case which is also the `new_colref` in this case.

libgpopt/src/operators/CPhysicalSplit.cpp
sambitesh: That was our first hunch but it doesn't work in the example I have cited below.

libgpopt/src/operators/CPhysicalSplit.cpp
sambitesh: I will update the comment to address this question

libgpopt/src/operators/CPhysicalSplit.cpp
zellerh: Thanks!

libgpopt/src/operators/CPhysicalSplit.cpp
```diff
@@ -356,26 +356,66 @@ CPhysicalSplit::PdsDerive
...
+			// the decision to insert a motion should depend only on the delete column.
+			if(pcrsDelete->FMember(new_colref))
+			{
+				pcrsModified->Include(pcrOld);
+			}
+			// if delete column and insert column belong to different tables, in that case we should
+			// check both of them against the distribution spec of the split's outer child because
+			else
+			{
+				pcrsModified->Include(pcrOld);
```
bhuvnesh2703: i believe we will not need pcrOld here.
Because if they are coming from different tables, in that case, this will be always disjoint with the pdsHashed for the split's outer child. Most likely, the decision will be always based on the new_colref for different tables. But i don't mind letting it be here.

libgpopt/src/operators/CPhysicalSplit.cpp
```diff
@@ -356,26 +356,68 @@ CPhysicalSplit::PdsDerive
...
+	for (ULONG ul = 0; ul < num_cols; ul++)
+	{
+		CColRef *pcrOld = (*m_pdrgpcrDelete)[ul];
+		pcrsDelete->Include(pcrOld);
+	}
+
 	for (ULONG ul = 0; ul < num_cols; ul++)
 	{
 		CColRef *pcrOld = (*m_pdrgpcrDelete)[ul];
 		CColRef *new_colref = (*m_pdrgpcrInsert)[ul];
```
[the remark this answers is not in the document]
sambitesh: Done !

sambitesh:
Applied all feedbacks.

PR #431, opened 2018-12-13
Avoid skew in motions for intermediate aggregate

zellerh:
When three stage aggs are produced, the optimizer considers two
alternatives while adding motions. One is with a motion on top of the
intermediate agg which reduces the skew, and the other is without a
motion but possible producing skew based on the data. This patch takes a
conservative approach and disables the second alternative so that the
plan produced will always have an extra motion and reduce the possible
skew.

This patch also introduces a guc `optimizer_enable_agg_skew_avoidance`
which can be used to turn on or off the behavior. This guc is set to on
by default which means that the second alternative will be disabled.


pivotal-issuemaster:
@zellerh Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=greenplum-db/gporca&pullRequestId=431)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=greenplum-db/gporca&pullRequestId=431) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@zellerh Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=greenplum-db/gporca&pullRequestId=431)!

bhuvnesh2703:
Also, please include the corresponding GPDB PR with the new GUC in the description.

zellerh:
GPDB PR: https://github.com/greenplum-db/gpdb/pull/6482

PR #433, opened 2018-12-20
Use default cardinality for stats when comparing different time-related types

chrishajas:
Previously, when comparing different time-related types such as
timestamp and timestamp with time zone, the cardinality estimate would
be inaccurate since it is using the binary representation of one type to
select buckets in the histogram of the other type without converting. We
would end up selecting the wrong plan node as a result in some cases.

Now, we mark time-related types as non-comparable and estimate the
cardinality as the default (40%) of the rows whenever comparing two
different time-related types.

Plan before patch:
```
explain select * from foo where ts = to_timestamp('99991231'::text, 'YYYYMMDD'::text);

                                          QUERY PLAN
----------------------------------------------------------------------------------------------
 Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..6.00 rows=1 width=336)
   ->  Index Scan using foo_pkey on foo  (cost=0.00..6.00 rows=1 width=336)
         Index Cond: ts = '9999-12-31 00:00:00-08'::timestamp with time zone
 Settings:  optimizer=on
 Optimizer status: PQO version 3.13.0
(5 rows)
```

Plan after patch:
```
explain select * from foo where ts = to_timestamp('99991231'::text, 'YYYYMMDD'::text);

                                        QUERY PLAN
 Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..342741.25 rows=206753204 width=336)
   ->  Table Scan on foo  (cost=0.00..83853.16 rows=68917735 width=336)
         Filter: ts = '9999-12-31 00:00:00-08'::timestamp with time zone
 Optimizer status: PQO version 3.13.0
(4 rows)
```


libnaucrates/src/base/IDatumStatisticsMappable.cpp
```diff
@@ -237,6 +238,13 @@ IDatumStatisticsMappable::StatsAreComparable
 
 	const IDatumStatisticsMappable *datum_cast = dynamic_cast<const IDatumStatisticsMappable*>(datum);
 
+	BOOL is_types_match = this->MDId()->Equals(datum_cast->MDId());
+	BOOL is_time_comparison = CMDTypeGenericGPDB::IsTimeRelatedType(this->MDId())
+			&& CMDTypeGenericGPDB::IsTimeRelatedType(datum_cast->MDId());
+	// the statistics for different time related types can't be directly compared, eg: timestamp vs timestamp with time zone.
+	// to prevent innaccurate statistics, mark as non-comparable
```
asubramanya: s/innaccurate/inaccurate/

libnaucrates/src/statistics/CStatsPredUtils.cpp
```diff
@@ -593,29 +593,31 @@ CStatsPredUtils::AddSupportedStatsFilters
 		GPOS_ASSERT(CStatsPredUtils::EptSentinel != ept);
 
 		// array extract function mapping
```
asubramanya: I think this comment can be removed.

libnaucrates/src/statistics/CStatsPredUtils.cpp
```diff
@@ -593,29 +593,31 @@ CStatsPredUtils::AddSupportedStatsFilters
...
-		SScStatsfilterMapping stats_filter_translators[] =
```
asubramanya: Seems like SScStatsfilterMapping is not being used anywhere else. Can we remove it?

libnaucrates/src/statistics/CStatsPredUtils.cpp
```diff
@@ -593,29 +593,31 @@ CStatsPredUtils::AddSupportedStatsFilters
...
-		{
-			{CStatsPredUtils::EptDisj, &CStatsPredUtils::CreateStatsPredDisj},
-			{CStatsPredUtils::EptScIdent, &CStatsPredUtils::GetStatsPredFromBoolExpr},
-			{CStatsPredUtils::EptLike, &CStatsPredUtils::GetStatsPredLike},
-			{CStatsPredUtils::EptPoint, &CStatsPredUtils::GetStatsPredPoint},
-			{CStatsPredUtils::EptConj, &CStatsPredUtils::CreateStatsPredConj},
-			{CStatsPredUtils::EptNullTest, &CStatsPredUtils::GetStatsPredNullTest},
-		};
-
-		FuncPtrStatsFilterFromPredExpr *function_ptr = &CStatsPredUtils::CreateStatsPredUnsupported;
```
asubramanya: I guess FuncPtrStatsFilterFromPredExpr can be removed too.

PR #435, opened 2018-12-27
Fix for cardinality estimation of left outer joins

zellerh:
The cardinality of a left outer join is estimated as the cardinality of
an inner join plus the cardinality of a left anti-semijoin (LASJ).

The cardinality of the LASJ is computed by subtracting the inner child
histogram from the outer child, leaving only buckets of the outer table
that have no counterparts in the inner table. This is computed by method
CHistogram::MakeLASJHistogram().

There was a bug in this method where we stopped too early in some
cases, leading to an over-estimation of the rows in the LASJ.

We also fixed a couple of small issues in CBucket::CompareLowerBounds(),
a typo and two cases where we compared the boundary values but not
the included/excluded flags.

Several mdp files needed changes, all of these now return lower
cardinality estimates for left outer joins.

Note that we may still overestimate some outer join cardinalities,
since we use an NDV-based estimate for the inner join and this
estimate sometimes exceeds the number of rows indicated by the
intersection of inner and outer histogram. When that is the case,
then adding the difference of outer - inner histograms causes
us to double-count the rows in the LASJ. We can consider changing
that in a future fix, if it becomes an issue with plan quality.

### DIFF
--- a/libnaucrates/src/statistics/CBucket.cpp
+++ b/libnaucrates/src/statistics/CBucket.cpp
@@ -604,7 +604,7 @@ CBucket::CompareUpperBounds
 	CPoint *point2 = bucket2->GetUpperBound();
 
 	BOOL is_closed_point1 = bucket1->IsUpperClosed();
-	BOOL is_closed_point2 = bucket1->IsUpperClosed();
+	BOOL is_closed_point2 = bucket2->IsUpperClosed();
 
 	if (point1->Equals(point2))
 	{
--- a/libnaucrates/src/statistics/CBucket.cpp
+++ b/libnaucrates/src/statistics/CBucket.cpp
@@ -516,7 +516,7 @@ CBucket::CompareLowerBounds
 	CPoint *point2 = bucket2->GetLowerBound();
 
 	BOOL is_closed_point1 = bucket1->IsLowerClosed();
-	BOOL is_closed_point2 = bucket1->IsLowerClosed();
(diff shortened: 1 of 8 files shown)


libnaucrates/src/statistics/CBucket.cpp
```diff
@@ -516,7 +516,7 @@ CBucket::CompareLowerBounds
 	CPoint *point2 = bucket2->GetLowerBound();
 
 	BOOL is_closed_point1 = bucket1->IsLowerClosed();
-	BOOL is_closed_point2 = bucket1->IsLowerClosed();
+	BOOL is_closed_point2 = bucket2->IsLowerClosed();
```
vraghavan78: please fix the same mistake in CBucket::CompareUpperBounds

zellerh: Thanks for catching that!

zellerh:
Closing this PR, it causes performance regressions, we also need to fix how we compute the LASJ cardinality when the histogram boundaries of inner and outer table are approximately the same, as in this simple repro, which will return a wrong cardinality estimate with our fix:

```
create table cs(a int not null, b int);
create table cr(a int not null, b int);

insert into cs select x, x from (select * from generate_series(1, 100000)) T(x);
insert into cr select 10*y, 10*y from (select * from generate_series(1, 10000)) T(y);

analyze cs;
analyze cr;

explain
select *
from cs left join cr on cs.a=cr.a
where cr.a is null;
```

PR #441, opened 2019-01-11
Fix MaxCard for full outer join

bhuvnesh2703:
Max card for full outer join should be the the maximum of the max cards
of the child. We can't collapse the full outer join to a constant table get
with false filter if any of the child has non-zero max card.

For the below query, we can't collapse the full outer join to a const table get
with false filter, as the left child of the full outer join has max card >0
```
explain
WITH cte1 AS (SELECT a::int FROM (VALUES (1),(2)) AS x(a)), -- left child
cte2 AS (SELECT b::int FROM (VALUES (2)) AS x(b) LIMIT 0) --right child
SELECT a, b FROM cte1 FULL JOIN cte2 ON 1 = 1;
                QUERY PLAN
-------------------------------------------
     One-Time Filter: false
```

After the fix, plan we be:
```
                                             QUERY PLAN
 Sequence  (cost=0.00..2648064.15 rows=2 width=8)
   ->  Shared Scan (share slice:id -1:2)  (cost=0.00..0.00 rows=1 width=1)
         ->  Materialize  (cost=0.00..0.00 rows=1 width=1)
               ->  Values Scan on "Values"  (cost=0.00..0.00 rows=1 width=4)
   ->  Sequence  (cost=0.00..2648064.15 rows=2 width=8)
         ->  Shared Scan (share slice:id -1:3)  (cost=0.00..0.00 rows=0 width=1)
               ->  Materialize  (cost=0.00..0.00 rows=0 width=1)
                     ->  Result  (cost=0.00..0.00 rows=0 width=4)
                           One-Time Filter: false
         ->  Append  (cost=0.00..2648064.15 rows=2 width=8)
               ->  Nested Loop Left Join  (cost=0.00..1324032.13 rows=1 width=8)
                     Join Filter: true
                     ->  Shared Scan (share slice:id -1:2)  (cost=0.00..431.00 rows=1 width=4)
                     ->  Shared Scan (share slice:id -1:3)  (cost=0.00..431.00 rows=0 width=4)
               ->  Result  (cost=0.00..1324032.02 rows=1 width=8)
                     ->  Nested Loop Left Anti Semi Join  (cost=0.00..1324032.02 rows=1 width=4)
                           Join Filter: true
                           ->  Shared Scan (share slice:id -1:3)  (cost=0.00..431.00 rows=0 width=4)
                           ->  Shared Scan (share slice:id -1:2)  (cost=0.00..431.00 rows=1 width=1)
 Optimizer status: PQO version 3.18.0
```

In Expression Preprocessing, during the step `PexprPruneEmptySubtrees`,
operators which had 0 max card were converted to a constant table get with
false filter due to which we were getting wrong results.  Now, we have fixed the
derivation of max card for full outer join, which earlier was derived as 0
incorrectly.


libgpopt/src/operators/CLogicalFullOuterJoin.cpp
```diff
@@ -55,7 +55,16 @@ CLogicalFullOuterJoin::Maxcard
...
+
+	CMaxCard left_child_maxcard = exprhdl.GetRelationalProperties(0)->Maxcard();
+	CMaxCard right_child_maxcard = exprhdl.GetRelationalProperties(1)->Maxcard();
+
+	if (left_child_maxcard <= right_child_maxcard)
+	{
+		return right_child_maxcard;
+	}
+
+	return left_child_maxcard;
```
vraghavan78: If both left and right are both non-zero then the output should be left_child_maxcard * right_child_maxcard just like in `MaxcardDef(exprhdl)`

bhuvnesh2703:
> recent changes has made the pipeline red.
fixed

PR #442, opened 2019-01-15
Revert commit to estimate cardinality for text columns based on NDV (5a32636)

asubramanya:
The commit 5a326369ba926349147f4bb6746ebd92d73ef024 causes underestimation of
cardinality in some cases leading to a sub optimal plan being picked.  Whenever
there are predicates of the form f(a) = b, the optimizer cannot estimate the
cardinality of the filter accurately since it doesn't know how the function
f(a) works. It should revert to using the default estimate in such cases.


libnaucrates/src/statistics/CHistogram.cpp
```diff
@@ -651,16 +651,6 @@ CHistogram::MakeJoinHistogramNormalize
...
 	GPOS_ASSERT(NULL != other_histogram);
 
-	if (CStatsPred::EstatscmptEqNDV == stats_cmp_type)
-	{
-		*scale_factor = std::max
-								 (
-								  std::max(CHistogram::MinDistinct.Get(), GetNumDistinct().Get()),
-								  std::max(CHistogram::MinDistinct.Get(), other_histogram->GetNumDistinct().Get())
-								  );
-		return MakeNDVBasedJoinHistogramEqualityFilter(mp, other_histogram);
```
hardikar: Can we get rid of `MakeNDVBasedJoinHistogramEqualityFilter` function?

hardikar:
Our intention is here to only revert for Filters and not touch joins, am I right? If so, LGTM.

vraghavan78:
@hardikar this is for Joins

asubramanya:
@hardikar Venky is right. This is for joins and not for filters. 
While reverting we did leave the refactoring of `MakeNDVBasedJoinHistogramEqualityFilter` though because we felt it made the code cleaner.

PR #444, opened 2019-01-25
Fix wrong results from LOJ reordering

hardikar:
This bug occurs in CXformExpandNAryJoinGreedy &
CXformExpandNAryJoinMinCard. These transforms reorder joins (including
LOJs) to produce a plan with a lower cost. In that process, predicates
need to be rearranged with the reordered tables. In the case of INNER
JOINS, it doesn’t matter if a predicate in included in as a join filter
or simply kept as a regular filter on the join results.  However, this
difference matters for OUTER joins, since a join condition mismatch
still produces results from one side. Thus is necessary to differentiate
between conditions that are part of the OUTER join condition and that
which a simple filter, since they change the semantics of an OUTER JOIN.

For example, in reordering the following query to bring the LOJ up, only
predicate A must be included in the resulting LOJ's join condition. Any
other conjuncts must be captured as a (join) filter.

Query:
```
+--CLogicalNAryJoin
   |--CLogicalLeftOuterJoin
   |  |--CLogicalGet "t1"
   |  |--CLogicalGet "t2"
   |  +-- <PREDICATE A: t1.a = t2.a>
   |--CLogicalGet "t3"
   +--CScalarBoolOp (EboolopAnd)
      |-- <PREDICATE B: t1.b = t3.b>
      +-- <PREDICATE C: t2.a IS NULL OR (t1.c = t3.c)>
```
Incorrect:
```
+--CLogicalLeftOuterJoin
   |--CLogicalInnerJoin
   |  |--CLogicalGet "t1"
   |  |--CLogicalGet "t3"
   |  +-- <PREDICATE B: t1.b = t3.b>
   |--CLogicalGet "t2"
   +--CScalarBoolOp (EboolopAnd)
      |-- <PREDICATE A: t1.a = t2.a>
      +-- <PREDICATE C: t2.a IS NULL OR (t1.c = t3.c)>
```
Correct:
```
+--CLogicalSelect
   |--CLogicalLeftOuterJoin
   |  |--CLogicalInnerJoin
   |  |  |--CLogicalGet "t1"
   |  |  |--CLogicalGet "t3"
   |  |  +-- <PREDICATE B: t1.b = t3.b>
   |  |--CLogicalGet "t2"
   |  +-- <PREDICATE A: t1.a = t2.a>
   +-- <PREDICATE C: t2.a IS NULL OR (t1.c = t3.c)>
```
GPDB PR: https://github.com/greenplum-db/gpdb/pull/6824


libgpopt/src/xforms/CJoinOrder.cpp
```diff
@@ -272,6 +275,8 @@ CJoinOrder::CJoinOrder
 	INT loj_id = 0;
 	INT comp_num = 0;
 
+	ULONG num_inner_join_conjuncts = pdrgpexprConj->Size();
```
vraghavan78: const

libgpopt/src/xforms/CJoinOrder.cpp
```diff
@@ -434,8 +443,17 @@ CJoinOrder::PcompCombine
 		// first call to this function, we create a Select node
 		parent_loj_id = comp2->ParentLojId();
 		position = comp2->Position();
-		pexpr = CUtils::PexprCollapseSelect(m_mp, pexprChild2, pexprScalar);
-		pexprScalar->Release();
+
+		GPOS_ASSERT(loj_conjuncts->Size() == 0);
+		//GPOS_ASSERT(filter_conjuncts->Size() == 0);
```
vraghavan78: remove

server/CMakeLists.txt
```diff
@@ -218,7 +218,7 @@ LeftJoin-UnsupportedFilter-Cardinality LeftOuter2InnerUnionAllAntiSemiJoin
 LOJ_IDF_no_convert_outer_ref_predicate_with_NULL
 LOJ_IDF_no_convert_outer_ref_predicate_with_inner_ref LOJ_convert_to_inner_with_inner_predicate_IDF
 LOJ_dont_convert_to_inner_with_inner_predicate_IDF_const LOJ_dont_convert_to_inner_with_outer_predicate_INDF
-LOJ_dont_convert_to_inner_with_outer_predicate_INDF_const LOJ-With-Agg;
+LOJ_dont_convert_to_inner_with_outer_predicate_INDF_const LOJ-With-Agg LOJReorderWithComplexPredicate;
```
vraghavan78: can we have test with a nary join with multiple LOJ children that gets merged?

libgpopt/src/xforms/CJoinOrder.cpp
```diff
@@ -434,8 +443,17 @@ CJoinOrder::PcompCombine
...
+
+		// combine all the conjuncts - since we are not creating a LOJ
+		CUtils::AddRefAppend(loj_conjuncts, other_conjuncts);
```
bhuvnesh2703: do we need to combine them, can we just use other_conjuncts?

libgpopt/src/xforms/CJoinOrder.cpp
```diff
@@ -477,7 +504,14 @@ CJoinOrder::PcompCombine
 				// for this join component
 				GPOS_ASSERT(NON_LOJ_DEFAULT_ID < parent_loj_id);
 			}
-			pexpr = CUtils::PexprLogicalJoin<CLogicalInnerJoin>(m_mp, pexprChild1, pexprChild2, pexprScalar);
+
+			GPOS_ASSERT(loj_conjuncts->Size() == 0);
+
+			// combine all the conjuncts - since we are not creating a LOJ
+			CUtils::AddRefAppend(loj_conjuncts, other_conjuncts);
```
bhuvnesh2703: do we need to combine here? use other_conjuncts..

libgpopt/include/gpopt/xforms/CJoinOrder.h
```diff
@@ -58,12 +58,15 @@ namespace gpopt
 				
 				// associated conjunct
 				CExpression *m_pexpr;
+
+				// tracks if the associated join is a LOJ
+				BOOL m_is_loj;
```
chrishajas: Add a comment here saying that we don't *really* need this, since we can figure this out by looking at each component & seeing if they are all LOJ related. Of course keeping this flag makes it easier and faster that to check that continuously.

libgpopt/src/xforms/CJoinOrder.cpp
chrishajas: For this case : `(NULL == pexprChild1)` I don't think its enough to only use other_conjuncts. For the other case, we would probably be OK - I just decided to go with combining for both because that matches earlier behavior & won't really be more expensive anyway.

server/CMakeLists.txt
chrishajas: Yup - adding a bunch of tests.

chrishajas:
@zellerh Yup - planning on doing that in a separate PR coming up soon.

PR #447, opened 2019-02-05
Remove conan yml files

bhuvnesh2703:
Since, we don't plan to use conan, remove the related files.
Bump ORCA to v3.23.2


hardikar:
All good : There are mentions of conan in `concourse/upload_orca_source_bintray.py` & `concourse/upload_orca_source_bintray.yml` also.

kmacoskey:
Lets slow down on this. Once we stop shipping orca to `bintray.com/greenplum-db` by removing `build_orca_using_conan` then there will need to be a more work in both gpdb5 and gpdb6 to align with shifting away from conan. A lot of oss documentation becomes out of date for gpdb5 and gpdb6, such that there needs to be something to replace it. The deb oss release in gpdb5 uses conan to build everything, so it would need to adjust as well.
We need to sync up about this topic to talk about what the long term plan is for gpdb5 and gpdb6. My suggestion would be that we find a way to minimize any impact to gpdb5 and ensure that there won't be any complications in gpdb6.

PR #453, opened 2019-03-12
Add total number of bindings and alternatives generated info

bhuvnesh2703:
For a group expression, the pattern defined by a xform is extracted, and the
xform can be applied to all the possible bindings for that group expression.
When the GUC optimizer_print_optimization_stats is on, the number of times the
xform is called is printed, however, even though there were several bindings
and the xform was applied several times, we record the number of calls to that
xform as 1. This commit adds more debugging information by printing the total
number of bindings and the total number of alternatives generated for a given
group expression.

Previous output:
CXformInnerJoin2IndexGetApply: 1 calls, 10 ms

Now:
CXformInnerJoin2IndexGetApply: 1 calls, 2 total binding, 2 alternatives generated, 10ms

This helps in identifying the total number of patterns which get extracted for
the given group expression which can often be a factor for long running
queries.

Signed-off-by: Ashuka Xue <axue@pivotal.io>


libgpopt/src/engine/CEngine.cpp
```diff
@@ -2501,6 +2521,12 @@ CEngine::FCheckReqdProps
 	return FCheckReqdPartPropagation(popPhysical, prpp->Pepp());
 }
 
+UlongPtrArray *
+CEngine::GetNumberOfBindings()
```
chrishajas: Is this being called anywhere?

libgpopt/src/engine/CEngine.cpp
```diff
@@ -1574,9 +1590,13 @@ CEngine::PrintActivatedXforms
 			CXform *pxform = CXformFactory::Pxff()->Pxf(xsi.TBit());
 			ULONG ulCalls = (ULONG) (*m_pdrgpulpXformCalls)[m_ulCurrSearchStage][pxform->Exfid()];
 			ULONG ulTime = (ULONG) (*m_pdrgpulpXformTimes)[m_ulCurrSearchStage][pxform->Exfid()];
+			ULONG ulBindings = (ULONG) (*m_pdrgpulpXformBindings)[m_ulCurrSearchStage][pxform->Exfid()];
+			ULONG ulResults = (ULONG) (*m_pdrgpulpXformResults)[m_ulCurrSearchStage][pxform->Exfid()];
 			os
 				<< pxform->SzId() << ": "
 				<< ulCalls << " calls, "
+				<< ulBindings << " total bindings, "
```
chrishajas: This may be more difficult to read/quickly parse due to alignment, but I don't think there's a way to align these columns easily.

libgpopt/src/engine/CEngine.cpp
ashuka24: It gets called in the test case in https://github.com/greenplum-db/gporca/pull/454.

libgpopt/src/engine/CEngine.cpp
asubramanya: yeah, we thought about it, we can format it, but in order to properly format it we need the biggest name for the xform and then format everything accordingly, so left it as is (good in looks but not really required) as you can take this output to a csv and scan through it.
Bhuvnesh

PR #454, opened 2019-03-12
Fix extraction of scalar group expression bindings

bhuvnesh2703:
Don't extract all bindings below scalar operators

Consider the below input query tree:
```
+--CLogicalInnerJoin
   |--CLogicalGet "t1"
   |--CLogicalSelect
   |  |--CLogicalGet "t3"
   |  +--CScalarSubqueryAny(=)
   |     |--CLogicalUnion
   |     |  |--CLogicalGet "t4"
   |     |  +--CLogicalGet "t5"
   |     +--CScalarIdent "b"
   +--CScalarCmp (=)
      |--CScalarIdent "b" (1)
      +--CScalarIdent "b" (10)
```
For the Operator CLogicalSelect, there are 2 different alternatives for
scalar condition which are generated by CXformSimplifySelectWithSubquery:
    1. CScalarSubqueryAny(=)["b" (19)]
    2. CScalarCmp (>)

For the Xform, CXformInnerJoinWithInnerSelect2IndexGetApply (and other equivalent xforms),
the pattern is as below:
```
+--TJoin
   |--CPatternLeaf
   |--CLogicalSelect
   |  |--TGet
   |  +--CPatternTree
   +--CPatternTree
```
If we look at the pattern for the scalar condition of CLogicalSelect, it's
defined as a CPatternTree. For the test query, the inner child of the join
(i.e CLogicalSelect), can have 2 different representation for the scalar
condition as denoted below by Alternative 1 and 2.  Anything below the
Scalar operators (CScalarSubqueryAny, CScalarCmp) is irrelevant in
determining the indexes which can be used for CLogicalGet "t3". Also, there
could be several bindings below the scalar operator, for instance, a Union
Operation can also be performed using Agg (Global) on Union ALL, and the
Global Agg could further have Global and Local alternatives.  We should
extract the Scalar Group Expression only once per alternative. In the case
of the test query used, there are 2 alternatives, so the test expects that
there should be 2 bindings for the pattern defined by
CXformInnerJoinWithInnerSelect2IndexGetApply.

Alternative: 1
```
+--CLogicalSelect
   |--CLogicalGet "t3"
   +--CScalarSubqueryAny(=)
      |--CLogicalUnion
      |  |--CLogicalGet "t4"
      |  +--CLogicalGet "t5"
      +--CScalarIdent
```

Alternative: 2 (CScalarSubqueryAny transformed to CScalarCmp > CScalarSubquery)
```
+--CLogicalSelect
   |--CLogicalGet "t3"
   +--CScalarCmp (>)
      |--CScalarSubquery
      ...... (Subtree - removed as its not relevant)
```
Signed-off-by: Ashuka Xue <axue@pivotal.io>

Note: We have added a test in this case which requires binding information. So, this PR is based on the other PR:  https://github.com/greenplum-db/gporca/pull/453.
We plan to merge 453 PR first.


bhuvnesh2703:
> Hmm the ordering also doesn't matter, then? Do we not need the entire expression?

1. Didn't get for what group you are asking for the ordering, but will give a shot. See if it helps.

Let's say in a group we have
Group 1:
CLogicalSelect [2 10]
CLogicalSelect [3 12]

where Group 10 Contains a CScalarSubquery
The Pattern expected by some xform is
```
+--Join
      |--CLogicalSelect
      |      |--CLeafPattern
      |     +--CPatternTree
      |---CLeafPattern
      +--CPatternTree 
```

So, when we extract the expression for CLogicalSelect [2 10], and build the tree for Group 10 (containing CScalarSubquery). Since, Group 10 contains a scalar operator, there will be no other alternative in Group 10, so ordering is not a concern. 
CScalarSubquery may have a child which can have multiple alternatives, but once we have extracted one complete tree under CScalarSubquery, the check added in this PR will mark that CScalarSubquery group as exhaustive and it will not extract any other alternative in its childs, as irrespective which child get's picked up it will have no role to play in identifying the index conditions. When optimization will kick in, it will ask CScalarSubquery to give whichever is the best child.

And for Group 1, it will go ahead and then look for the matching Pattern for CLogicalSelect [3 12].

2. Entire tree is extracted.

If this didn't help to clarify the question, let me know what ordering you are referring to.

hardikar:
@bhuvnesh2703 The explanation did help. Let me clarify my question. My concern with the PR is more about the fact that it is very generally applied. So I want to make sure we're not missing some case(s). 
Let me elaborate: You say
> irrespective which child get's picked up it will have no role to play in identifying the index conditions

1. The changes in this PR doesn't affect IndexJoins only. It affects any other op that uses a `CPatternTree` to extract a subquery. Does the above statement apply for all those cases also?
2. Even in case of IndexJoins, can you elaborate why we must extract the entire tree, if the children getting extracted don't matter?
3. What if the same scalar operator is used in multiple groups? Still OK with extracting only once?

I feel like I'm missing some intuition here.

bhuvnesh2703:
> 1. The changes in this PR doesn't affect IndexJoins only. It affects any other op that uses a CPatternTree to extract a subquery. Does the above statement apply for all those cases also?

Yes, you are right, it affects other also, and it applies to all of them.

> 2. Even in case of IndexJoins, can you elaborate why we must extract the entire tree, if the children getting extracted don't matter?

Let's consider the below tree:
```
2019-03-25 14:39:09:691058 PDT,THD000,TRACE,"Xform: CXformInnerJoinWithInnerSelect2IndexGetApply
Input:
+--CLogicalInnerJoin   origin: [Grp:15, GrpExpr:1]
   |--CLogicalSelect   origin: [Grp:4, GrpExpr:0]
   |--CLogicalSelect   origin: [Grp:13, GrpExpr:0]
   |  |--CLogicalGet "t3" ("t3"), Columns: ["a" (9), "b" (10), "ctid" (11), "xmin" (12), "cmin" (13), "xmax" (14), "cmax" (15), "tableoid" (16), "gp_segment_id" (17)] Key sets: {[2,8]}   origin: [Grp:5, GrpExpr:0]
   |  +--CScalarBoolOp (EboolopAnd)   origin: [Grp:12, GrpExpr:0]
   |     |--CScalarSubqueryAny(=)["b" (19)]   origin: [Grp:10, GrpExpr:0]
   |     |  |--CLogicalUnion Output: ("b" (19)), Input: [("b" (19)), ("b" (28))]   origin: [Grp:8, GrpExpr:0]
   |     |  |  |--CLogicalGet "t4" ("t4"), Columns: ["a" (18), "b" (19), "ctid" (20), "xmin" (21), "cmin" (22), "xmax" (23), "cmax" (24), "tableoid" (25), "gp_segment_id" (26)] Key sets: {[2,8]}   origin: [Grp:6, GrpExpr:0]
   |     |  |  +--CLogicalGet "t5" ("t5"), Columns: ["a" (27), "b" (28), "ctid" (29), "xmin" (30), "cmin" (31), "xmax" (32), "cmax" (33), "tableoid" (34), "gp_segment_id" (35)] Key sets: {[2,8]}   origin: [Grp:7, GrpExpr:0]
   |     |  +--CScalarIdent "b" (10)   origin: [Grp:9, GrpExpr:0]
   |     +--CScalarCmp (=)   origin: [Grp:11, GrpExpr:0]
   |        |--CScalarIdent "b" (10)   origin: [Grp:9, GrpExpr:0]
   |        +--CScalarConst (5)   origin: [Grp:2, GrpExpr:0]
   +--CScalarCmp (=)   origin: [Grp:14, GrpExpr:0]
      |--CScalarIdent "b" (1)   origin: [Grp:1, GrpExpr:0]
      +--CScalarIdent "b" (10)   origin: [Grp:9, GrpExpr:0]

```
For the above tree, the objective of CXformInnerJoinWithInnerSelect2IndexGetApply is to convert the child `CLogicalGet "t3"` of join's inner child `CLogicalSelect   origin: [Grp:13, GrpExpr:0]` into a CLogicalIndexGet if indexes exists. (and InnerJoin to IndexApply). To do that, it has to extract the predicates from the Scalar Child of `CLogicalSelect   origin: [Grp:13, GrpExpr:0]`, to identify what columns the index can be applied (+ push the applicable predicates from the inner join to the inner child). So, you have to extract the whole tree. Below will be the output generated.

```
Output:
Alternatives:
0:
+--CLogicalIndexApply
   |--CLogicalSelect   origin: [Grp:4, GrpExpr:0]
   |--CLogicalSelect
   |  |--CLogicalIndexGet   Index Name: ("t3idx"), Table Name: ("t3"), Columns: ["a" (9), "b" (10), "ctid" (11), "xmin" (12), "cmin" (13), "xmax" (14), "cmax" (15), "tableoid" (16), "gp_segment_id" (17)]
   |  |  +--CScalarBoolOp (EboolopAnd)
   |  |     |--CScalarCmp (=)
   |  |     |  |--CScalarIdent "b" (10)   origin: [Grp:9, GrpExpr:0]
   |  |     |  +--CScalarIdent "b" (1)   origin: [Grp:1, GrpExpr:0]
   |  |     +--CScalarCmp (=)   origin: [Grp:11, GrpExpr:0]
   |  |        |--CScalarIdent "b" (10)   origin: [Grp:9, GrpExpr:0]
   |  |        +--CScalarConst (5)   origin: [Grp:2, GrpExpr:0]
   |  +--CScalarSubqueryAny(=)["b" (19)]   origin: [Grp:10, GrpExpr:0]
   |     |--CLogicalUnion Output: ("b" (19)), Input: [("b" (19)), ("b" (28))]   origin: [Grp:8, GrpExpr:0]
   |     |  |--CLogicalGet "t4" ("t4"), Columns: ["a" (18), "b" (19), "ctid" (20), "xmin" (21), "cmin" (22), "xmax" (23), "cmax" (24), "tableoid" (25), "gp_segment_id" (26)] Key sets: {[2,8]}   origin: [Grp:6, GrpExpr:0]
   |     |  +--CLogicalGet "t5" ("t5"), Columns: ["a" (27), "b" (28), "ctid" (29), "xmin" (30), "cmin" (31), "xmax" (32), "cmax" (33), "tableoid" (34), "gp_segment_id" (35)] Key sets: {[2,8]}   origin: [Grp:7, GrpExpr:0]
   |     +--CScalarIdent "b" (10)   origin: [Grp:9, GrpExpr:0]
   +--CScalarConst (1)
```

> 3. What if the same scalar operator is used in multiple groups? Still OK with extracting only once?

The extraction is driven by the parent group. So, if there are 2 group referring to the same scalar child, they would be extracted 2 times. The logic in the PR applies to the GE in the same group.

libgpopt/src/search/CBinding.cpp
```diff
@@ -167,6 +167,19 @@ CBinding::PexprExtract
...
 
+	// for a scalar operator, there is only one group expression in it's
+	// group. so, if it's been extracted once, there is no need to explore
+	// all the child bindings, as it is not useful for transformations
+	if (NULL != pexprLast && pgexpr->Pgroup()->FScalar())
+	{
+#ifdef GPOS_DEBUG
+		// the last operator and the current group expression will be same
+		// for a scalar operator, as there is only one group expression
+		GPOS_ASSERT(pexprLast->Pop()->Eopid() == pgexpr->Pop()->Eopid());
```
hardikar: Let's add an assertion here that confirms that the scalar group has exactly one group expressions.
Let's also add this explanation in the comments.

bhuvnesh2703: added an assertion for the number of GE and the comment.

PR #462, opened 2019-04-02
Normalize expression before inferring predicates

zellerh:
Previously, we missed some inferred predicates in cases where we
generated an IS NOT NULL predicate on a column c, when we had a
more restrictive filter like c=5 on an ancestor node in the tree.
Since the IS NOT NULL predicate marked column c as processed, we
did not consider the more restrictive predicate.

The fix is to normalize the tree before inferring predicates, so
that we push all predicates down first. The issue below was
introduced as a regression with commit 9eed4f, when we changed
the order in which we infer predicates from top-down to bottom-up.

Repro:

create table cat (ci int, cj int);
create table foo (fi int, fj int);
create table bar (bi int, bj int) partition by range(bj) (start (1) end (10) every (1));

explain
select *
from (SELECT  pos_prod.cj
      FROM (foo pnl
            LEFT JOIN bar sub_prod ON pnl.fj = sub_prod.bj),
           cat pos_prod
      WHERE pos_prod.ci = 1
            AND pos_prod.cj = pnl.fj) t
where t.cj=3;

Plan before fix:
```
                                                        QUERY PLAN
---------------------------------------------------------------------------------------------------------------------------
 Gather Motion 3:1  (slice3; segments: 3)  (cost=0.00..1293.00 rows=1 width=4)
   ->  Hash Join  (cost=0.00..1293.00 rows=1 width=4)
         Hash Cond: foo.fj = cat.cj
         ->  Hash Left Join  (cost=0.00..862.00 rows=1 width=4)
               Hash Cond: foo.fj = bar.bj
                     Filter: fj = 3
                     ->  Broadcast Motion 3:3  (slice1; segments: 3)  (cost=0.00..431.00 rows=1 width=4)
                           ->  Sequence  (cost=0.00..431.00 rows=1 width=4)
                                 ->  Partition Selector for bar (dynamic scan id: 1)  (cost=10.00..100.00 rows=34 width=4)
                                       Partitions selected: 9 (out of 9)
                                 ->  Dynamic Table Scan on bar (dynamic scan id: 1)  (cost=0.00..431.00 rows=1 width=4)
               ->  Broadcast Motion 3:3  (slice2; segments: 3)  (cost=0.00..431.00 rows=1 width=4)
                     ->  Table Scan on cat  (cost=0.00..431.00 rows=1 width=4)
                           Filter: ci = 1 AND cj = 3
 Optimizer status: PQO version 3.32.1
(18 rows)
```

Plan after fix:
```
                                                        QUERY PLAN
---------------------------------------------------------------------------------------------------------------------------
 Gather Motion 3:1  (slice3; segments: 3)  (cost=0.00..1293.00 rows=1 width=4)
   ->  Hash Join  (cost=0.00..1293.00 rows=1 width=4)
         Hash Cond: foo.fj = cat.cj
         ->  Hash Left Join  (cost=0.00..862.00 rows=1 width=4)
               Hash Cond: foo.fj = bar.bj
                     Filter: fj = 3
                     ->  Broadcast Motion 3:3  (slice1; segments: 3)  (cost=0.00..431.00 rows=1 width=4)
                           ->  Sequence  (cost=0.00..431.00 rows=1 width=4)
                                 ->  Partition Selector for bar (dynamic scan id: 1)  (cost=10.00..100.00 rows=34 width=4)
                                       Partitions selected: 1 (out of 9)
                                 ->  Dynamic Table Scan on bar (dynamic scan id: 1)  (cost=0.00..431.00 rows=1 width=4)
                                       Filter: bj = 3
               ->  Broadcast Motion 3:3  (slice2; segments: 3)  (cost=0.00..431.00 rows=1 width=4)
                     ->  Table Scan on cat  (cost=0.00..431.00 rows=1 width=4)
                           Filter: ci = 1 AND cj = 3
 Optimizer status: PQO version 3.32.1
(19 rows)
```

The unit test change is due to us inferring an extra predicate and not recognizing that it is a duplicate. This isn't a bug in the current code, but rather an existing deficiency by not deduplicating predicates, because the inferred predicate looks slightly different from the original (OR predicate vs. IN predicate). We've created another story to fix this.


zellerh:
@hardikar asked to run the following query, to see whether we manage to infer predicates correctly even with CTEs:

```
create table foo(fa int, fb int);
create table bar(ba int, bb int);

explain
with cte1 as (select * from foo, bar)
select *
from cte1
where fa = ba and fa=10;
```

We tried it and fortunately got a good plan with inferred pred ba=10:

```
                                  QUERY PLAN
-------------------------------------------------------------------------------
 Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..862.00 rows=1 width=16)
   ->  Hash Join  (cost=0.00..862.00 rows=1 width=16)
         Hash Cond: foo.fa = bar.ba
               Filter: fa = 10
         ->  Hash  (cost=431.00..431.00 rows=1 width=8)
               ->  Table Scan on bar  (cost=0.00..431.00 rows=1 width=8)
                     Filter: ba = 10
 Optimizer status: PQO version 3.32.3
(9 rows)
```
Tagging @ashuka24 and @sambitesh in case you want to try this test case as well for your current story.

hardikar:
@zellerh 

I also tried this query:
```
shardikar=# explain                                                                                                                                         with cte1 as (select * from foo, bar where fa = 10)                                                                                                         select * from cte1                                                                                                                                                   where fa = ba;
                                  QUERY PLAN
-------------------------------------------------------------------------------
 Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..862.00 rows=1 width=16)
   ->  Hash Join  (cost=0.00..862.00 rows=1 width=16)
         Hash Cond: (foo.fa = bar.ba)
         ->  Seq Scan on foo  (cost=0.00..431.00 rows=1 width=8)
               Filter: ((fa = 10) AND (fa = 10))
         ->  Hash  (cost=431.00..431.00 rows=1 width=8)
               ->  Seq Scan on bar  (cost=0.00..431.00 rows=1 width=8)
                     Filter: (ba = 10)
 Optimizer: PQO version 3.32.3
(9 rows)
```

While the inference is correct, it does produces duplicate predicates. Any idea why this isn't handled? I ask because I believe the 'var = const' is handled by the predicate deduplication logic.

chrishajas:
@hardikar This is a result of CTE inlining. After the preprocessing, the query is 
```

2019-04-04 17:33:47:623274 PDT,THD000,TRACE,"+--CLogicalCTEAnchor (0)
   +--CLogicalSelect
      |--CLogicalCTEConsumer (0), Columns: ["fa" (18), "fb" (19), "ba" (20), "bb" (21)]
      +--CScalarBoolOp (EboolopAnd)
         |--CScalarCmp (=)
         |  |--CScalarIdent "fa" (18)
         |  +--CScalarConst (10)
         |--CScalarCmp (=)
         |  |--CScalarIdent "ba" (20)
         |  +--CScalarConst (10)
         +--CScalarCmp (=)
            |--CScalarIdent "fa" (18)
            +--CScalarIdent "ba" (20)

2019-04-04 17:34:02:750775 PDT,THD000,TRACE,"+--CLogicalNAryJoin
   |--CLogicalSelect
   |  |--CLogicalGet "foo" ("foo"), Columns: ["fa" (0), "fb" (1), "ctid" (2), "xmin" (3), "cmin" (4), "xmax" (5), "cmax" (6), "tableoid" (7), "gp_segment_id" (8)] Key sets: {[2,8]}
   |  +--CScalarCmp (=)
   |     |--CScalarIdent "fa" (0)
   |     +--CScalarConst (10)
   |--CLogicalSelect
   |  |--CLogicalGet "bar" ("bar"), Columns: ["ba" (9), "bb" (10), "ctid" (11), "xmin" (12), "cmin" (13), "xmax" (14), "cmax" (15), "tableoid" (16), "gp_segment_id" (17)] Key sets: {[2,8]}
   |  +--CScalarCmp (=)
   |     |--CScalarIdent "ba" (9)
   |     +--CScalarConst (10)
   +--CScalarCmp (=)
      |--CScalarIdent "fa" (0)
      +--CScalarIdent "ba" (9)
```

So this would need to be fixed during inlining. This issue existed prior to this PR however.

PR #463, opened 2019-04-03
Predicate pushdown in part tables with window func

ashuka24:
Previously, we pushed down a filter predicate that should not have been
pushed down due to the presence of a window function. This was causing
the optimizer to select a NLJ instead of a Hash Join. The pushed down
filter predicate also selected partitions incorrectly which could have
lead to wrong results.

This commit changes CCopyPartFilter to take into account the paritition
columns in the window function when determining which filter predicate
can be pushed down.

Repro:
```
CREATE TABLE p1 (a int, b int) PARTITION BY range(b) (START (1) END (5)
EVERY (1));
CREATE TABLE p2 (a int, b int) PARTITION BY range(b) (START (1) END (5)
EVERY (1));

WITH cte AS (SELECT p1.b, row.number() over() FROM p1, p2 where p1.a =
p2.a) SELECT 1 FROM cte where cte.b > 2;
```

Before the fix: We have an NLJ with wrongly selected partitions:
```
                                                        QUERY PLAN
--------------------------------------------------------------------------------------------------------------------------
 Result  (cost=0.00..1324032.34 rows=1 width=4)
   ->  Result  (cost=0.00..1324032.34 rows=1 width=1)
         Filter: (p1.b > 2)
         ->  WindowAgg  (cost=0.00..1324032.34 rows=1 width=4)
               ->  Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..1324032.34 rows=1 width=4)
                     ->  Nested Loop  (cost=0.00..1324032.34 rows=1 width=4)
                           Join Filter: (p1.a = p2.a)
                                 ->  Partition Selector for p1 (dynamic scan id: 1)  (cost=10.00..100.00 rows=34 width=4)
                                       Partitions selected: 3 (out of 4)
                                 ->  Dynamic Seq Scan on p1 (dynamic scan id: 1)  (cost=0.00..431.00 rows=1 width=8)
                           ->  Sequence  (cost=0.00..431.00 rows=1 width=4)
                                 ->  Partition Selector for p2 (dynamic scan id: 2)  (cost=10.00..100.00 rows=34 width=4)
                                       Partitions selected: 4 (out of 4)
                                 ->  Dynamic Seq Scan on p2 (dynamic scan id: 2)  (cost=0.00..431.00 rows=1 width=4)
 Optimizer: PQO version 3.32.3
(16 rows)
```

After the fix: We have a Hash join with correctly selected partitions
```
                                                           QUERY PLAN
--------------------------------------------------------------------------------------------------------------------------------
 Result  (cost=0.00..862.00 rows=1 width=4)
   ->  Result  (cost=0.00..862.00 rows=1 width=1)
         Filter: (p1.b > 2)
         ->  WindowAgg  (cost=0.00..862.00 rows=1 width=4)
               ->  Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..862.00 rows=1 width=4)
                     ->  Hash Join  (cost=0.00..862.00 rows=1 width=4)
                           Hash Cond: (p1.a = p2.a)
                                 ->  Partition Selector for p1 (dynamic scan id: 1)  (cost=10.00..100.00 rows=34 width=4)
                                       Partitions selected: 4 (out of 4)
                                 ->  Dynamic Seq Scan on p1 (dynamic scan id: 1)  (cost=0.00..431.00 rows=1 width=8)
                                 ->  Sequence  (cost=0.00..431.00 rows=1 width=4)
                                       ->  Partition Selector for p2 (dynamic scan id: 2)  (cost=10.00..100.00 rows=34 width=4)
                                             Partitions selected: 4 (out of 4)
                                       ->  Dynamic Seq Scan on p2 (dynamic scan id: 2)  (cost=0.00..431.00 rows=1 width=4)
 Optimizer: PQO version 3.32.3
(17 rows)
```

Co-authored by: Ashuka Xue <axue@pivotal.io>
Co-authored by: Shreedhar Hardikar <shardikar@pivotal.io>


libgpopt/src/translate/CTranslatorExprToDXL.cpp
```diff
@@ -4890,16 +4890,17 @@ CTranslatorExprToDXL::ConstructLevelFilters4PartitionSelector
...
 															(
 															m_mp,
 															m_pmda,
-															CUtils::PexprScalarIdent(m_mp, pcrPartKey),
+															pexprIdent,
 															pmdidTypePartKey,
 															ulLevel
 															);
-
+				pexprIdent->Release();
```
chrishajas: We also call PdxlnListFilterPartKey in PdxlnScCmpPartKey, do we need to release pexprIdent there too? Should we instead release it inside PdxlnListFilterPartKey?

libgpopt/src/operators/CPhysicalSequenceProject.cpp
```diff
@@ -413,7 +413,20 @@ CPhysicalSequenceProject::PppsRequired
 	GPOS_ASSERT(0 == child_index);
 	GPOS_ASSERT(NULL != pppsRequired);
 
-	return CPhysical::PppsRequiredPushThruUnresolvedUnary(mp, exprhdl, pppsRequired, CPhysical::EppcAllowed);
+	CColRefSet *pcrsPartCols;
+	if (CDistributionSpec::EdtHashed == Pds()->Edt())
+	{
+		pcrsPartCols = CUtils::PcrsExtractColumns(mp, CDistributionSpecHashed::PdsConvert(Pds())->Pdrgpexpr());
```
bhuvnesh2703: the current code only considers the keys used in the starting hash distribution spec, but there could be equivalent hash distribution spec in Pds(), which can have different keys / columns which can impact this logic..

in case for this we will not have equivalent distribution spec, then it should be fine.

hardikar: So, the logic here is similar/related to `CNormalizer::FPushableThruSeqPrjChild()`. Even there, equivalent distribution spec is not considered.

I think we should do the following:
1. Add a comment in this method that mentions the related method.
2. Add an assert in both places asserting that equivalent hash distribution is NULL.
3. Assert in the constructor of CPhysicalSequenceProject that equivalent hash distr is NULL. It is created from a LogicalSequenceProject (which has no idea about equivalent distributions AFAIK. See `CXformImplementSequenceProject::Transform()`.

libgpopt/src/translate/CTranslatorExprToDXL.cpp
hardikar: > do we need to release pexprIdent there too?
Maybe - that code this wierd.

> Should we instead release it inside PdxlnListFilterPartKey?

Nope, it has a read-only usage. I wish ORCA used c++11 so that this was more obvious

Issue #321, opened 2018-01-30
::UlHash() member function should cache the hash value


hsyuan:
Most of logical/physical operators or  are immutable. We can safely cache the hash value to avoid redundant calculating.

PR #465, opened 2019-04-05
Use prime number of buckets to avoid hash collisions

chrishajas:
Previously, the default number of buckets/chains in our hash map and
hash set was 128. In cases when passing in a column represented by a
ulong, a column number "12" would be interpreted as 4 bytes
such as [1,2,\0\0]. The hash function is below:
```
for(ULONG i = 0; i < size; ++i)
	{
		BYTE b = byte_array[i];
		hash = ((hash << 5) ^ (hash >> 27)) ^ b;
	}
```

By shifting the hash left 5 bits, we effectively multiply by 32. Since
we do this twice, we end up creating a hash that is a multiple of 2048
(or possibly larger). However, we choose the bucket by `hash % 128`. Since
our hash is always a multiple of 128, we end up hashing to the same
bucket. This takes a significant amount of time in some cases since we
need to iterate through the entire hash map to check if a column already
exists (and inserts/selects take O(n) instead of O(1)). Changing the
default number of buckets/chains to 127 makes the distribution more
uniform.

After applying this change, the planning time for a query went down by
almost 50%. Specifically, greedy, mincard, and DP planning time went
down by ~80% for this query. Additionally, the 106 way join in ctest
goes from ~55s to 42s on my laptop (and total ctest goes from ~68s to
56s).

Note: Another alternative is to change the hash function and keep the default  number of chains as 128.


hsyuan:
Nice findings. Rather than changing the bucket size, I think we should change the hash function. DEK doesn't perform well as a generic hash function, as the [link](https://research.neustar.biz/2012/08/20/k-minimum-values-sketching-error-hash-functions-and-you/) shows.

AP hash function might be a better one, even not as good as Murmur3.
http://www.partow.net/programming/hashfunctions/#APHashFunction
https://research.neustar.biz/2011/12/29/choosing-a-good-hash-function-part-2/

BTW, the default MEMO bucket size is still not a prime number.
```
src/search/CMemo.cpp
36:#define GPOPT_MEMO_HT_BUCKETS	50000
```

hsyuan:
Another thing we can also optimize is to let each immutable object(scalar, logical, physical operator, metadata object, CMD...) cache its own hash code, like how Java's Integer, Long, String do. There is a issue for this: https://github.com/greenplum-db/gporca/issues/321

Currently, the hash code is recomputed each time function is called, for immutable objects, it is a waste. But I am not sure how much performance gain we will get.

bhuvnesh2703:
Changing it to 127 is less invasive. We can do this in step by step approach, let's first fix the length validate the results..and push it in.
We can then investigate what are the bottlenecks with the current hash function and replace it with a new one. (But via a separate PR)

PR #469, opened 2019-04-19
Avoid unnecessary broadcasts on outer table for index joins

sambitesh:
An index join forces its outer child to match a distribution spec that
is equivalent to that of the inner child. When the inner child derives
its distribution spec (CPhysicalScan::PdsDerive()), we generate both
the actual and the equivalent distribution spec (the equivalent one
being expressed in ColRefs of the outer table).

The code missed one type of index scan: CPhysicalDynamicIndexScan.


scripts/fix_mdps.py
```diff
@@ -100,7 +100,7 @@ def processLogFile(logFileLines):
         elif errorLogMatch:
             print "Error in file %s: %s" % (current_file, lines)
         elif failureMatch:
-            if not re.search(r'.*Plan .* comparison.*', lines):
+            if not re.search(r'Plan [a-z ]*comparison', lines) and not re.search(r'Unittest', lines):
```
bhuvnesh2703: is this intentional?

PR #470, opened 2019-04-25
Reduce projected columns for UNION

zellerh:
When we have a UNION operator that needs to produce only a subset
of the available columns, then also reduce the project lists of the
children of the union. The fix consists of two changes.

First, in method CPhysicalUnionAll::PcrsRequired(), we require only
those columns from the child that are also needed by the parent of
the union operator.

Second, when generating project lists in the expression to DXL transformation,
we use the required plan properties to project only the required columns.
We need to generate n+1 matching project lists, that of the union operator
itself and the project lists of the n children. We first build an array of
integer indexes of the required columns for the union and we then reduce
each of the n+1 lists using this array, ensuring a match.


bhuvnesh2703:
Have we tested the results of the setup which was provided in story, did it come down as expected?

libgpopt/src/translate/CTranslatorExprToDXL.cpp
```diff
@@ -1927,25 +1927,44 @@ CTranslatorExprToDXL::PdxlnAppend
...
+	ULONG num_total_cols = pdrgpcrOutputAll->Size();
+	for (ULONG c=0; c<num_total_cols; c++)
+	{
+		if (reqdCols->FMember((*pdrgpcrOutputAll)[c]))
+		{
+			reqd_col_array->Append((*pdrgpcrOutputAll)[c]);
+		}
+	}
+	ULongPtrArray *reqd_col_positions = pdrgpcrOutputAll->IndexesOfSubsequence(reqd_col_array);
+	CColRefArray *requiredOutput = pdrgpcrOutputAll->ReduceArray(reqd_col_positions);
```
bhuvnesh2703: instead of ReduceArray, do you think CopyColRefsAt is more valid here.. because you are not reducing the original array..

libgpopt/src/operators/CPhysicalUnionAll.cpp
```diff
@@ -206,18 +207,17 @@ const
...
+		IMemoryPool *mp,
 		CExpressionHandle &,//exprhdl,
-		CColRefSet *, //pcrsRequired,
+		CColRefSet *pcrsRequired,
 		ULONG child_index,
 		CDrvdProp2dArray *, // pdrgpdpCtxt
 		ULONG // ulOptReq
 	)
 {
-	CColRefSet *pcrs = PcrsInput(child_index);
```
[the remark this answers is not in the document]
zellerh: Good point, I'll delete it.

libgpopt/src/operators/CPhysicalUnionAll.cpp
```diff
@@ -957,6 +957,36 @@ const
...
+CPhysicalUnionAll::MapOutputColRefsToInput(IMemoryPool *mp,
+										   CColRefSet *out_col_refs,
+										   ULONG child_index)
+{
+	CColRefSet *result = GPOS_NEW(mp) CColRefSet(mp);
+	CColRefArray *all_outcols = m_pdrgpcrOutput;
+	ULONG total_num_cols = all_outcols->Size();
+	CColRefArray *in_colref_array = (*PdrgpdrgpcrInput())[child_index];
+	CColRefSetIter iter(*out_col_refs);
+	while (iter.Advance())
```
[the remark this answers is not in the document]
zellerh: Yes, it would be very nice to avoid this nested loop, but I think for that we would need to add one or two hash maps, because with the current arrays we need the index in the array to translate between ColRefs.

libgpopt/src/translate/CTranslatorExprToDXL.cpp
zellerh: Yes, I agree that the name is misleading. The method is defined in CDynamicPtrArray, so the name can't mention ColRefs. Hope "CreateReducedArray" as a method name is ok.

libgpopt/src/operators/CPhysicalUnionAll.cpp
```diff
@@ -957,6 +946,36 @@ const
...
+	{
+		ULONG out_col_id = iter.Bit();
+		ULONG i = 0;
```
asubramanya: It would probably be more readable to use a boolean flag to check if the out_col_id was found and then assert on that instead of asserting that `i` is equal to `total_num_cols`.

zellerh:
@bhuvnesh2703, here is the result of a performance measurement:

51 columns, 10,000,000 rows per table:

| Table format | num tables in union | table/view      | master, execution time in msec | fix, execution time in msec |
| ------------ | ------------------- | --------------- | ------------------------------ | --------------------------- |
| row          | 1                   | table, 51 cols  | 11307                          | 10863                       |
| row          | 3                   | view, 51 cols   | 11701                          | 11637                       |
| row          | 3                   | view, 3 cols    | 11244                          | 11092                       |
| column       | 1                   | table, 51 cols  | 1981                           | 1944                        |
| column       | 3                   | view, 51 cols   | 18471                          | 2247                        |
| column       | 3                   | view, 3 cols    | 1892                           | 2026                        |

Even though the fix applies to both row and column store, we don't see real savings with row-oriented tables, but we do see an improvement for columnar tables. The fix can only improve the measurements above where we have a 3-way union of tables with 51 columns (the second and fifth measurement). All other results should be the same with and without the fix.

libgpopt/include/gpopt/operators/CPhysicalUnionAll.h
```diff
@@ -36,6 +36,13 @@ namespace gpopt
 			// map given array of scalar ident expressions to positions of UnionAll input columns in the given child;
 			ULongPtrArray *PdrgpulMap(IMemoryPool *mp, CExpressionArray *pdrgpexpr, ULONG child_index) const;
 
+			// map given ColRefArray, expressed in terms of outputs,
+			// into an equivalent ColRefArray, expressed in terms
```
asubramanya: s/ColRefArray/ColRefSet/

bhuvnesh2703:
> Even though the fix applies to both row and column store, we don't see real savings with row-oriented tables, but we do see an improvement for columnar tables. The fix can only improve the measurements above where we have a 3-way union of tables with 51 columns (the second and fifth measurement). All other results should be the same with and without the fix.

Thanks Hans for the update. With Rows tables, sure i think it doesn't make a difference as we still have to read the complete row, but in case of columnar tables just the specific column files. This is a great improvement from 18471ms to 2247ms.
LGTM.

libgpopt/src/operators/CPhysicalUnionAll.cpp
zellerh: Thanks, I like that suggestion.

libgpopt/include/gpopt/operators/CPhysicalUnionAll.h
zellerh: Done.

PR #473, opened 2019-05-01
Fix external table spec union

bhuvnesh2703:
This PR does is separated into 2 commits:

commit 1: Fix Require Distribution Spec for Union with External Scan Childs

If the required spec input to `CPhysicalSerialUnionAll` is not hash
distributed, `CPhysicalSerialUnionAll` requests its 0th child to provide
any distribution spec. For any other child n+0 where n=1, based on the
spec of 0th child, we require specific distribution spec. In case of 0th
child providing External Distribution Spec, it requests Random Spec from
all other child. This add extra Random Motion on all other child
although it's not required.
This commit fixes the issue by requesting Any Distribution Spec from all
the other childs if the outer child provides External Distribution Spec

Consider the setup
```
CREATE EXTERNAL TABLE ext_expenses(name text, date date, amount float4, category text, desc1 text )
LOCATION ('file://host1:5432/data/expense/*.csv')
FORMAT 'CSV' (HEADER);
```

Current Plan:
```
EXPLAIN SELECT max (name) FROM (SELECT name FROM ext_expenses UNION ALL SELECT name FROM ext_expenses UNION ALL SELECT name FROM ext_expenses)f;
                                                     QUERY PLAN
---------------------------------------------------------------------------------------------------------------------
 Aggregate  (cost=0.00..1367.57 rows=1 width=8)
   ->  Gather Motion 3:1  (slice3; segments: 3)  (cost=0.00..1367.57 rows=1 width=8)
         ->  Aggregate  (cost=0.00..1367.57 rows=1 width=8)
               ->  Append  (cost=0.00..1366.67 rows=1000000 width=8)
                     ->  External Scan on ext_expenses  (cost=0.00..442.37 rows=333334 width=8)
                     ->  Redistribute Motion 3:3  (slice1; segments: 3)  (cost=0.00..455.67 rows=333334 width=8)
                           ->  External Scan on ext_expenses ext_expenses_1  (cost=0.00..442.37 rows=333334 width=8)
                     ->  Redistribute Motion 3:3  (slice2; segments: 3)  (cost=0.00..455.67 rows=333334 width=8)
                           ->  External Scan on ext_expenses ext_expenses_2  (cost=0.00..442.37 rows=333334 width=8)
 Optimizer: Pivotal Optimizer (GPORCA) version 3.35.0
```
In the above plan, there is an aggregate (global and local) max on top
of Union / Append Node, we don't need a redistribute motion to be added
on any of the child of the Union but since Union all Request Random, a
motion is added.

Plan after fix:
```
EXPLAIN SELECT max (name) FROM (SELECT name FROM ext_expenses UNION ALL SELECT name FROM ext_expenses UNION ALL SELECT name FROM ext_expenses)f;
                                                   QUERY PLAN
         Aggregate  (cost=0.00..1350.88 rows=1 width=8)
           ->  Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..1350.88 rows=1 width=8)
                 ->  Aggregate  (cost=0.00..1350.88 rows=1 width=8)
                       ->  Append  (cost=0.00..1349.98 rows=1000000 width=8)
                             ->  External Scan on ext_expenses  (cost=0.00..442.37 rows=333334 width=8)
                             ->  External Scan on ext_expenses  (cost=0.00..442.37 rows=333334 width=8)
                             ->  External Scan on ext_expenses  (cost=0.00..442.37 rows=333334 width=8)
         Optimizer status: PQO version 3.35.0
```

commit 2:
Enable CExternalTableTest

Somehow, the call to add_orca_test to include CExternalTableTest was from CMakeLists, so it was never executed.
Re-enabling the test with this commit and updated the plans.


bhuvnesh2703:
Well, i think the logic in CPhysicalSerialUnionAll::PdsRequired is tightly coupled to the spec provided by the outer. Because we check the spec of the outer and based on that we request spec from the other child..
Consider this case where the first child is a regular table and the second child is a external scan, we will still add a redistribute motion on top of all the other childs excluding 0th.

```
bhuvnesh=# explain select * from (select * from t1 union all select * from ext_expenses)f;
                                              QUERY PLAN
 Gather Motion 3:1  (slice2; segments: 3)  (cost=0.00..1056.51 rows=1000001 width=32)
   ->  Append  (cost=0.00..937.26 rows=333334 width=32)
         ->  Table Scan on t1  (cost=0.00..431.00 rows=1 width=32)
         ->  Redistribute Motion 3:3  (slice1; segments: 3)  (cost=0.00..495.59 rows=333334 width=32)
               ->  External Scan on ext_expenses  (cost=0.00..442.37 rows=333334 width=32)
 Optimizer status: PQO version 3.35.0
```
Might need to re-think the fix, or limit the scope of this fix to the existing implementation.

vraghavan78:
> Redistribute Motion 3:3  (slice1; segments: 3)  (cost=0.00..495.59 rows=333334 width=32)
>                ->  External Scan on

This may be sub-optimal, but does it generate wrong results. I do not think so.

V

bhuvnesh2703:
@vraghavan78 agree, it is sub-optimal, but will not generate wrong results.

Some more cases where it's sub-optimal.
create table t1 (a int) distributed by (a);
I believe, the way we request spec from the child of union all is sub-optimal in general and is highly dependent on the outer. consider the below examples:
0th child is Universal, and 1st child is Hash distributed on a.
If you place 0th child first in the union all, it pulls the 1st child on Master. Here t1 is a huge table 4000509...
```
bhuvneshchaudhary=# explain select 1 union all select a from t1;
                                        QUERY PLAN
 Append  (cost=0.00..730.48 rows=1333504 width=4)
   ->  Result  (cost=0.00..0.00 rows=1 width=4)
   ->  Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..714.48 rows=4000509 width=4)
         ->  Table Scan on t1  (cost=0.00..505.81 rows=1333503 width=4)
 Optimizer status: PQO version 3.35.0
```
However, if you change the order, it does it on segments..
```
bhuvneshchaudhary=# explain select a from t1 union all select 1;
                                     QUERY PLAN
------------------------------------------------------------------------------------
 Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..600.54 rows=4000510 width=4)
   ->  Append  (cost=0.00..540.91 rows=1333504 width=4)
         ->  Table Scan on t1  (cost=0.00..505.81 rows=1333503 width=4)
         ->  Result  (cost=0.00..0.00 rows=1 width=4)
 Optimizer status: PQO version 3.35.0
```

Consider the below case were t1 is distributed by a; Since `select 1` is universal, even though the table t1 is distributed by a, still there is a redistribute motion on t1..
explain select a from (select 1 as a union all select a from t1)t group by a;
```
                                                                              QUERY PLAN
------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Gather Motion 3:1  (slice6; segments: 3)  (cost=0.00..739.24 rows=2 width=4)
   ->  GroupAggregate  (cost=0.00..739.24 rows=1 width=4)
         Group By: a
         ->  Sort  (cost=0.00..739.24 rows=1 width=4)
               Sort Key: a
               ->  Redistribute Motion 3:3  (slice5; segments: 3)  (cost=0.00..739.24 rows=1 width=4)
                     Hash Key: a
                     ->  GroupAggregate  (cost=0.00..739.24 rows=1 width=4)
                           Group By: a
                           ->  Sort  (cost=0.00..739.24 rows=1 width=4)
                                 Sort Key: a
                                 ->  Redistribute Motion 1:3  (slice4)  (cost=0.00..739.24 rows=2 width=4)
                                       ->  Append  (cost=0.00..739.24 rows=1 width=4)
                                             ->  GroupAggregate  (cost=0.00..0.00 rows=1 width=4)
                                                   Group By: a
                                                   ->  Sort  (cost=0.00..0.00 rows=1 width=4)
                                                         Sort Key: a
                                                         ->  Result  (cost=0.00..0.00 rows=1 width=4)
                                             ->  Gather Motion 3:1  (slice3; segments: 3)  (cost=0.00..739.24 rows=1 width=4)
                                                   ->  GroupAggregate  (cost=0.00..739.24 rows=1 width=4)
                                                         Group By: a
                                                         ->  Sort  (cost=0.00..739.24 rows=1 width=4)
                                                               Sort Key: a
                                                               ->  Redistribute Motion 3:3  (slice2; segments: 3)  (cost=0.00..739.24 rows=1 width=4)
                                                                     Hash Key: a
                                                                     ->  HashAggregate  (cost=0.00..739.24 rows=1 width=4)
                                                                           Group By: a
                                                                           ->  Redistribute Motion 3:3  (slice1; segments: 3)  (cost=0.00..577.13 rows=1333503 width=4)
                                                                                 ->  Table Scan on t1  (cost=0.00..505.81 rows=1333503 width=4)
```

vraghavan78:
Got it. Thanks! So there is money on the table. In case this fix should go in immaterial to the other issue correct?

bhuvnesh2703:
> Got it. Thanks! So there is money on the table. In case this fix should go in immaterial to the other issue correct?
Yes

libgpopt/src/operators/CPhysicalSerialUnionAll.cpp
```diff
@@ -137,6 +137,13 @@ CPhysicalSerialUnionAll::PdsRequired
 		return GPOS_NEW(mp) CDistributionSpecReplicated();
 	}
 
+	if (CDistributionSpec::EdtExternal == pdsOuter->Edt())
+	{
+		// if the outer child delivers external spec, require inner child
+		// to provide any spec
+		return GPOS_NEW(mp) CDistributionSpecAny(this->Eopid());
```
d: -1

If the _problem_ we're fixing is that we have generated an unnecessary motion on top of an `External Scan`, then the _solution_ really shouldn't be in the `Append` operator: you could end up with extraneous motions on top of `External Scan`s elsewhere, should we put this logic in every operator that can possibly be on top of an external scan?

We ought to ask ourselves, what makes the following line *not work*?

```cxx
	return GPOS_NEW(mp) CDistributionSpecNonSingleton(false /*fAllowReplicated*/);
```

The operator (before the patch) is actually doing the right thing. What's actually lacking is the logic of `CDistributionSpecExternal::FSatisfies`. Compare that code to `CDistributionSpecRandom::FSatisfies`.

PR #474, opened 2019-05-02
Fix the project list of SplitUpdate

sambitesh:
Take the case below:

```
CREATE TABLE foo (
    a_dist int,
    b int,
    c_part int
)
WITH (appendonly=false) DISTRIBUTED BY (a_dist) PARTITION BY RANGE(c_part)
          (
          PARTITION p20190305 START (1) END (2) WITH (tablename='foo_1_prt_p20190305', appendonly=false    )
          );
;

insert into foo(a_dist, c_part)  values (12,1);

update foo set b=c_part;
```
For the UPDATE statement above, the CPhysicalSplitUpdate will receive delete and insert columns as below:
InsertColIdx: 0 2 2
DeleteColIdx: 0 1 2

Currently we use InsertColIdx as our project list. Doing so gave us an error while running the
update statement.

```
ERROR: no partition for partitioning key (seg1 127.0.0.1:25433 pid=41342)
```

In GPDB, the DML node (specifically, the ExecDelete codepath) will use the project list of its child (SplitUpdate) to
determine which partition it should delete from. We saw that in ExecDelete, the
tuple values were incorrect, and the wrong value caused the logic to think that
the value to be deleted did not belong in the partition.

Instead, the SplitUpdate node should project the columns of the table being
updated, so DeleteColIdx should be projected. The projected columns only seem
to be used for partition selection; they're not needed/used for the insertion
part of SplitUpdate operations.


bhuvnesh2703:
did you add the tests for the case which were not caught in ctest/icg when you had your first fix?

hardikar:
For my reference, the plan for the repro case in the PR comments is:
```
 Update  (cost=0.00..431.07 rows=1 width=1)
   ->  Partition Selector for foo  (cost=0.00..431.00 rows=1 width=26)
         ->  Split  (cost=0.00..431.00 rows=1 width=22)
               ->  Sequence  (cost=0.00..431.00 rows=1 width=22)
                     ->  Partition Selector for foo (dynamic scan id: 1)  (cost=10.00..100.00 rows=34 width=4)
                           Partitions selected: 1 (out of 1)
                     ->  Dynamic Seq Scan on foo (dynamic scan id: 1)  (cost=0.00..431.00 rows=1 width=22)
 Optimizer: Pivotal Optimizer (GPORCA) version 3.38.1
```
(The "Split" is actually a "SplitUpdate" in the code)

Anyway, I had a couple of questions:
1. If the DML node (Update in this case) takes care of picking the partitions, what's the point of that Partition Selector in the plan?
2. The PartitionSelector's project list is the same as the SplitUpdate underneath it, am I correct?
2. On the surface, it does make more sense to use delete cols. Did you manage to dig through history to figure what insert cols were chosen?

sambitesh:
@hardikar 


> Anyway, I had a couple of questions:
> 
> 1. If the DML node (Update in this case) takes care of picking the partitions, what's the point of that top Partition Selector in the plan?

This is a special Partition Selector called PartitionSelectorDML. We removed it and we still got correct results for our update queries. It seems like it is serving no purpose. But we didn't want to remove it as part of this story. We can create a story to look into PartitionSelectorDML.

> 2. The PartitionSelector's project list is the same as the SplitUpdate underneath it, am I correct?

Yes, that's correct.

> 3. On the surface, it does make more sense to use delete cols. Did you manage to dig through history to figure what insert cols were chosen?

We spent quite a bit of time in archeology. We couldn't find a reason for why did they chose insert cols in the first place. The original commit to add `CPhysicalSplit` and the design doc for DML do not mention anything about project list of `CPhysicalSplit`

hardikar:
@sambitesh Ah. It's too bad that archeology didn't reveal much. And yes, let's create that story. Meanwhile, the current fix LGTM.

Did any ICG tests fail? Anyway, waiting for that PR.

hardikar:
@sambitesh responded elsewhere saying : "Result node was simply reordering the columns. We don't need the reordering anymore after we have started using delete columns."

PR #477, opened 2019-05-07
Remove conan

dsharp-pivotal:


chrishajas:
The 5X_STABLE ubuntu jobs still use Conan correct? If so, I don't think we can remove uploading the conan artifacts from orca repo, as we use the same ORCA branch for 5/6/master currently.

khuddlefish:
@chrishajas Oh ok. Is that just the `upload_orca_source_bintray` task?

chrishajas:
> @chrishajas Oh ok. Is that just the `upload_orca_source_bintray` task?

upload_orca_source_bintray also copies conanfile.py. I think may only be able to remove the build_orca and build_gpdb jobs.

dsharp-pivotal:
We took out the second commit which was removing `conanfile.py` and the `upload_orca_source_bintray` task. The remaining commit just removes the `build_orca` and `build_gpdb` jobs.

PR #478, opened 2019-05-08
Fix PdsRequired for local CPhysicalAgg with DQAs

zellerh:
When we had a local aggregate with a DQA (e.g. select count(distinct b)
... group by a), we generated a required distribution on the DQA column
(b in this example). We now also consider an alternative distribution
spec with both the groupby and the DQA columns (a, b in the example).

Note: We have some MDP tests for multiple DQAs which are not supported
in ORCA - when run through PSQL the same query will fall back. We
don't generate good plans for those and we should probably delete
those MDPs (MDQA*.mdp).


libgpopt/src/operators/CPhysicalAgg.cpp
```diff
@@ -86,9 +86,13 @@ CPhysicalAgg::CPhysicalAgg
...
 		{
 			// If the local aggregate has distinct columns we generate
-			// one optimization requests for its children:
-			// (1) hash distribution on the distinct columns
-			ulDistrReqs = 1;
+			// two optimization requests for its children:
+			// (1) hash distribution on the distinct columns only
+			// (2) hash distribution on the grouping and distinct
+			//     columns (only if the grouping columns are not empty)
+			if (0 == m_pdrgpcr->Size()) {
```
bhuvnesh2703: is this array always initialized? do we need to check it for `!= NULL` too?

libgpopt/src/operators/CPhysicalAgg.cpp
```diff
@@ -235,8 +239,19 @@ CPhysicalAgg::PdsRequiredAgg
 
 	if (COperator::EgbaggtypeLocal == m_egbaggtype && m_pdrgpcrArgDQA != NULL && 0 != m_pdrgpcrArgDQA->Size())
 	{
-		GPOS_ASSERT(0 == ulOptReq);
-		return PdsMaximalHashed(mp, m_pdrgpcrArgDQA);
+		if (ulOptReq == 0) {
```
bhuvnesh2703: very minor, `{` on next line..

libgpopt/src/operators/CPhysicalAgg.cpp
```diff
@@ -235,8 +239,19 @@ CPhysicalAgg::PdsRequiredAgg
...
+			return PdsMaximalHashed(mp, m_pdrgpcrArgDQA);
+		}
+		else
+		{
+			GPOS_ASSERT(1 == ulOptReq);
+			CColRefArray *combinedCols = GPOS_NEW(mp) CColRefArray(mp);
```
bhuvnesh2703: var name may be .. grp_and_distinct_cols

libgpopt/src/operators/CPhysicalAgg.cpp
```diff
@@ -235,8 +239,19 @@ CPhysicalAgg::PdsRequiredAgg
...
+			combinedCols->AppendArray(m_pdrgpcr);
```
bhuvnesh2703: should we assert m_pdrgpcr != NULL

libgpopt/src/operators/CPhysicalAgg.cpp
zellerh: Done.

libgpopt/src/operators/CPhysicalAgg.cpp
zellerh: I checked and all the paths that call the CPhysicalAgg constructor seem to call AddRef() on the ColRefArray that is passed in for m_pdrgpcr, so I think it's safe to assume that it is non-null. There also is already an assert in the constructor that ensures the pointer is not NULL.

libgpopt/src/operators/CPhysicalAgg.cpp
zellerh: Done.

libgpopt/src/operators/CPhysicalAgg.cpp
zellerh: That assert already exists, in the CPhysicalAgg constructor.

PR #486, opened 2019-05-30
Fix flaky ctests, caused by an incorrectly resolved merge conflict

zellerh:
In my previous commit, I fixed some test failures incorrectly, such that
they caused flaky behavior. This restores the intended behavior of the
tests, by not checking the plan space size.

The flaky tests caused GPDB 6 and Master pipelines to go red, because
those pipelines now compile ORCA and also run ctest.


zellerh:
An explanation what happened: These tests are run in a special test unit where we don't check the plan size. That requires the set optimizer_enumerate_plans guc to be off (traceflag 101013). When resolving merge conflicts, I accidentally turned this trace flag back on, and that turned the flaky plan size check back on.

PR #487, opened 2019-05-31
Do not expand array if array size > array_expansion_threshold

sambitesh:
Earlier, the translator would not convert the IN list to an ArrayExpr if the
size of IN list is greater than the optimizer_array_expansion_threshold.

But now  an IN list of any size gets converted to an ArrayExpr. If
such an ArrayCmpExpr gets expanded to its Disjuncts, then we can run out of memory.


hardikar:
Suggested rewrite of the commit message:
```
Earlier, the translator would not convert the IN list to an ArrayExpr if the
size of IN list is greater than the optimizer_array_expansion_threshold.
However, we still should create an ArrayExpr because it can be useful in 
deriving stats on filters even if we are not deriving constraints on it.
Yet to prevent out of memory errors, we expand the ArrayCmpExpr into 
conjucts/disjuncts in `PexprExpand ` only when within that threshold.
```

Otherwise LGTM.

PR #494, opened 2019-06-12
Fix stack-use-after-scope for `CCacheHashtableAccessor` instantiation

d:
Fix stack-use-after-scope for `CCacheHashtableAccessor` instantiation

`CCacheHashtableAccessor` is constructed from a const reference to type `K`.  When we construct a `CCacheHashtableAccessor` with a temporary of type `K` -- such as the return value of `CCacheEntry::Key()` -- the lifetime of the referred-to temporary object of type `K` is only as long as the constructor call. This leads to a stack-use-after-scope when `CCacheHashtableAccessor` later tries to dereference the memoized key.

To fix this, extend the lifetime of the temporary returned by `CCacheEntry::Key()` with a local const reference, so that it outlives the `CCacheHashtableAccessor` object. Strictly speaking we only need to fix `gpos::CCache::InsertEntry`, because `gpos::CCache::ReleaseEntry` technically won't touch `m_key`. But we've just escaped by the skin of our teeth there: it's just a bad idea to initialize the accessor with an immediately invalid reference. So I chose to fix both.

I also scanned the code base for occurrences of the following pattern (static lifetime analysis, yay!):

* A class `C` (such as `gpos::CSyncHashtableAccessByKey<>`) that saves a reference through constructor parameter `M`

* An instantiation of an object of type `C`, but the argument passed to `M` (the reference parameter) is a temporary object (such as the return value of a function) that is cleaned up "immediately" after the constructor call

Turns out all occurrences are constructing `CCacheHashtableAccessor` with a short-lived temporary, and all of them instantiate the class template with very small data types as `K`. Fix those with copying the temporary onto the stack (instead of the generic const-ref-lifetime-extension).

Bug discovered with LLVM AddressSanitizer.

## Notes to reviewers

I am seeking feedback on
1. The commit message mostly, because I'm not sure I have articulated clearly the problem (and the solution)


d:
Hans had a good point in https://github.com/greenplum-db/gporca/pull/494#discussion_r293081878 .

For posterity, I wrote a quick static analysis in `clang-query`:

```
set bind-root false
set output diag

let hasConstRefParam forEachArgumentWithParam(expr().bind("materializing a temporary here, but it will be destroyed immediately after constructing the accessor"),parmVarDecl(hasType(references(isConstQualified()))))

let ctorForAccessorByKey hasDeclaration(cxxConstructorDecl(ofClass(hasName("gpos::CSyncHashtableAccessByKey"))))

match varDecl(has(exprWithCleanups(has(cxxConstructExpr(ctorForAccessorByKey, hasConstRefParam)))))
```

Run this over all of ORCA, and you'll find 6 violations:

```
alm ~/w/g/build.release.asan ((9ca48274…))> git ls-files ':/*.cpp' | parallel -m --tty -j +0 /usr/local/opt/llvm/bin/clang-query -f ../short-lived-temporaries-in-ctor.clang-query

Match #1:

../libgpos/include/gpos/memory/CCache.h:150:47: note: "materializing a temporary here, but it will be destroyed immediately after
      constructing the accessor" binds here
                                CCacheHashtableAccessor acc(m_hash_table, entry->Key());
                                                                          ^~~~~~~~~~~~

Match #2:

../libgpos/include/gpos/memory/CCache.h:209:48: note: "materializing a temporary here, but it will be destroyed immediately after
      constructing the accessor" binds here
                                        CCacheHashtableAccessor acc(m_hash_table, entry->Key());
                                                                                  ^~~~~~~~~~~~

Match #3:

../libgpos/include/gpos/memory/CCache.h:150:47: note: "materializing a temporary here, but it will be destroyed immediately after
      constructing the accessor" binds here
                                CCacheHashtableAccessor acc(m_hash_table, entry->Key());
                                                                          ^~~~~~~~~~~~

Match #4:

../libgpos/include/gpos/memory/CCache.h:209:48: note: "materializing a temporary here, but it will be destroyed immediately after
      constructing the accessor" binds here
                                        CCacheHashtableAccessor acc(m_hash_table, entry->Key());
                                                                                  ^~~~~~~~~~~~
0 matches.

Match #5:

/Users/pivotal/workspace/gporca/libgpos/src/memory/CMemoryPoolManager.cpp:173:43: note: "materializing a temporary here, but it will be
      destroyed immediately after constructing the accessor" binds here
                MemoryPoolKeyAccessor acc(m_hash_table, mp->GetHashKey());
                                                        ^~~~~~~~~~~~~~~~

Match #6:

/Users/pivotal/workspace/gporca/libgpos/src/memory/CMemoryPoolManager.cpp:352:43: note: "materializing a temporary here, but it will be
      destroyed immediately after constructing the accessor" binds here
                MemoryPoolKeyAccessor acc(m_hash_table, mp->GetHashKey());
                                                        ^~~~~~~~~~~~~~~~
6 matches.
0 matches.

Match #1:

../libgpos/include/gpos/memory/CCache.h:150:47: note: "materializing a temporary here, but it will be destroyed immediately after
      constructing the accessor" binds here
                                CCacheHashtableAccessor acc(m_hash_table, entry->Key());
                                                                          ^~~~~~~~~~~~

Match #2:

../libgpos/include/gpos/memory/CCache.h:209:48: note: "materializing a temporary here, but it will be destroyed immediately after
      constructing the accessor" binds here
                                        CCacheHashtableAccessor acc(m_hash_table, entry->Key());
                                                                                  ^~~~~~~~~~~~

Match #3:

/Users/pivotal/workspace/gporca/libgpopt/src/mdcache/CMDAccessor.cpp:660:46: note: "materializing a temporary here, but it will be
      destroyed immediately after constructing the accessor" binds here
                        MDHTAccessor mdhtacc(m_shtCacheAccessors, a_pmdaccelem->MDId());
                                                                  ^~~~~~~~~~~~~~~~~~~~
3 matches.
0 matches.
0 matches.
0 matches.

Match #1:

/Users/pivotal/workspace/gporca/libgpopt/src/search/CGroupExpression.cpp:755:21: note: "materializing a temporary here, but it will be
      destroyed immediately after constructing the accessor" binds here
        ShtAcc shta(Sht(), pcc->Poc());
                           ^~~~~~~~~~
1 match.

```

hardikar:
@d Thanks for pointing all the other cases. Were you planning on pushing a fix for them?

d:
> Were you planning on pushing a fix for them?

Yep!

chrishajas:
@d Just a reminder to write the comment you mentioned offline.

PR #495, opened 2019-06-13
Remove Broadcast from Index NLJ

ashuka24:
During optimization, we push down scalar predicates to our children when we can. 
For example, if bar had an index on c, the optimization process would push down the
predicate as below:
```
ScalarBoolOpAnd 
|--  foo.a = bar.b
+-- foo.c = bar.d
```
becomes
```
Filter 
  |-- Index Scan 
  |      +-- foo.c = bar.d
  +-- foo.a = bar.b
```
However in some cases, the original predicate is needed for the
optimization process. This is now saved as m_pexprScalar in
CLogicalApply and CPhysicalNLJoin.

By saving the original scalar predicate, we can use it to generate proper
equivalent distribution specs which removes unnecessary Broadcasts from
the plan for index nested loop joins.

This allows ORCA to generate plans using index nested loop joins on 
queries whose index are not on the distribution key.

### DIFF
--- a/libgpopt/src/operators/CPhysicalInnerIndexNLJoin.cpp
+++ b/libgpopt/src/operators/CPhysicalInnerIndexNLJoin.cpp
@@ -130,14 +130,18 @@ CPhysicalInnerIndexNLJoin::PdsRequired
 	{
 		// check if we could create an equivalent hashed distribution request to the inner child
 		CDistributionSpecHashed *pdshashed = CDistributionSpecHashed::PdsConvert(pdsInner);
-		CDistributionSpecHashed *pdshashedEquiv = pdshashed->PdshashedEquiv();
-		if (NULL != pdshashedEquiv)
+		CExpression *pexprScalar = CPhysicalNLJoin::PopConvert(exprhdl.Pop())->ScalarExpr();
+		CDistributionSpecHashed *pdshashedMatching = CUtils::CreateMatchingHashedDistribution(mp, pexprScalar, pdshashed);
+		if (NULL != pdshashedMatching)
 		{
-			// request hashed distribution from outer
-			pdshashedEquiv->Pdrgpexpr()->AddRef();
-			CDistributionSpecHashed *pdsHashedRequired = GPOS_NEW(mp) CDistributionSpecHashed(pdshashedEquiv->Pdrgpexpr(), pdshashedEquiv->FNullsColocated());
-			pdsHashedRequired->ComputeEquivHashExprs(mp, exprhdl);
-			return pdsHashedRequired;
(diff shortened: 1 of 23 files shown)


hardikar:
A couple of questions:
1. Is there a precedence to saving the scalar expression in an operator? I would love that to be the standard, but as of now, I don't think that's done yet.
2. Can you explain why this something CPhysicalFilter::PdsDerive() can't handle? It sits on top of the IndexScan.. can it not add a matching distribution spec?

ashuka24:
> Is there a precedence to saving the scalar expression in an operator? I would love that to be the standard, but as of now, I don't think that's done yet.
Not exactly, but we do something similar for Hash Joins where we store the clauses in the operator. 

> Can you explain why this something CPhysicalFilter::PdsDerive() can't handle? It sits on top of the IndexScan.. can it not add a matching distribution spec?
We actually tried this first. The filter's two children are the IndexGet and a Scalar predicate. In this context, the filter cannot obtain information about the IndexGet's child.  While this works with tables with only one distribution key, the filter does not have all the information needed to created a matching distribution spec for tables with multiple distribution keys.

libgpopt/src/operators/CPhysicalInnerIndexNLJoin.cpp
```diff
@@ -130,14 +134,18 @@ CPhysicalInnerIndexNLJoin::PdsRequired
 	{
 		// check if we could create an equivalent hashed distribution request to the inner child
 		CDistributionSpecHashed *pdshashed = CDistributionSpecHashed::PdsConvert(pdsInner);
-		CDistributionSpecHashed *pdshashedEquiv = pdshashed->PdshashedEquiv();
-		if (NULL != pdshashedEquiv)
+		CExpression *pexprScalar = CPhysicalInnerIndexNLJoin::PopConvert(exprhdl.Pop())->ScalarExpr();
+		CDistributionSpecHashed *pdshashedMatching = CUtils::CreateMatchingHashedDistribution(mp, pexprScalar, pdshashed);
```
zellerh: As far as I understand, we already have three other ways to produce a matched distribution spec:

- The equivalent distribution specs in ```pdshashed->PdshashedEquiv()```
- Another way to create equivalent distribution specs, using ```pdshashed->HashSpecEquivExprs()```
- The equivalence classes in ```exprhdl.GetRelationalProperties()->Ppc()->PdrgpcrsEquivClasses()``` could also be used to build equivalent distribution specs

If that understanding is correct, then do we really want to add a fourth way to achieve this, by storing the original join predicate in the apply? A comment not related to this PR: I would like to get rid of the second option in the above list as well.

ashuka24:
Closing this PR in preference of: https://github.com/greenplum-db/gporca/pull/508

PR #499, opened 2019-06-22
Skip group expression containing circular references

bhuvnesh2703:
Consider the group expression CLogicalSelect [ 0 3 ]
it has the 0th child coming from Group 0, where Group 0 has Duplicate Group 4
While deriving Stats, this will cause a circular loop as follows
1. CLogicalSelect child 0 -> Will ask the stats to be derived on Group 0
2. Group 0 will ask Group 4 to give the stats (as its duplicate),
which will then again ask CLogicalSelect [0 4] to derive stats resulting in a loop.
Such Group Expression can be ignored for deriving stats and implementation.
```
Group 4 (#GExprs: 5):
0: CLogicalSelect [ 0 3 ]
1: CLogicalNAryJoin [ 6 7 8 ] Origin: (xform: CXformInlineCTEConsumerUnderSelect, Grp: 4, GrpExpr: 0)
2: CLogicalCTEConsumer (0), Columns: ["a" (18), "b" (19), "a" (20), "b" (21)] [ ]
3: CLogicalNAryJoin [ 6 7 3 ] Origin: (xform: CXformInlineCTEConsumer, Grp: 4, GrpExpr: 2)
4: CLogicalInnerJoin [ 6 7 3 ] Origin: (xform: CXformExpandNAryJoinGreedy, Grp: 4, GrpExpr: 3)
```

```
Group 0 (#GExprs: 0, Duplicate Group: 4):
```

Note: It only considers skipping group expression whose immediate child
has the current group as its duplicate. Multi-level dependencies are not
considered, and not sure if we have that issue.


libgpopt/src/search/CGroupExpression.cpp
```diff
@@ -1178,6 +1179,63 @@ CGroupExpression::OsPrintCostContexts
...
+	if (m_ecirculardependency == CGroupExpression::ecdCircularDependency)
+	{
+		return true;
+	}
+
+	GPOS_ASSERT(m_ecirculardependency == CGroupExpression::ecdDefault);
+
+	// if exploration is completed, then the group expression does not have
+	// any circular dependency
+	if (Pgroup()->FStartedImplementing())
```
hardikar: Hm.. why not FExplored here?

libgpopt/src/search/CJobGroupImplementation.cpp
```diff
@@ -172,8 +172,11 @@ CJobGroupImplementation::FScheduleGroupExpressions
 	{
 		if (!pgexpr->FTransitioned(CGroupExpression::estImplemented))
 		{
-			CJobGroupExpressionImplementation::ScheduleJob(psc, pgexpr, this);
-			pgexprLast = pgexpr;
+			if (!pgexpr->ContainsCircularDependencies())
```
hardikar: Merge this IF case with the previous one?

libgpopt/src/search/CGroupExpression.cpp
```diff
@@ -1178,6 +1179,63 @@ CGroupExpression::OsPrintCostContexts
...
+	CGroupArray *child_groups = Pdrgpgroup();
+	for (ULONG ul = 0; ul < child_groups->Size(); ul++)
+	{
+		CGroup *child_group = (*child_groups)[ul];
+		if (child_group->FScalar())
+			continue;
+		CGroup *duplicate_group = child_group->PgroupDuplicate();
+		if (duplicate_group != NULL)
+		{
+			ULONG duplicate_group_id = duplicate_group->Id();
```
hardikar: s/duplicate_group_id/child_group_id/

bhuvnesh2703: will make this child_duplicate_group_id.. as this is not the child_group_id.. but (childs duplicate group id)

libgpopt/src/search/CJobGroupImplementation.cpp
bhuvnesh2703: sure

libgpopt/src/search/CGroupExpression.cpp
bhuvnesh2703: we can have FExplored.. It should be fine too. 
I just thought, it was more safer to avoid any corner cases (not sure if there are any)

bhuvnesh2703:
@hardikar applied feedback..

PR #501, opened 2019-06-25
Fix distinct on outer references in subqueries in IN clause

asubramanya:
Consider the query as shown below:
select * from foo where foo.a in (select distinct foo.a from bar);
Preprocessing step converts this to an EXISTS query by pulling out the column
from the subquery which is an outer reference.
```
Algebrized query:
+--CLogicalSelect
   |--CLogicalGet "foo" ("foo"), Columns: ["a" (0), "ctid" (1), "xmin" (2), "cmin" (3), "xmax" (4), "cmax" (5), "tableoid" (6), "gp_segment_id" (7)] Key sets: {[1,7]}
   +--CScalarSubqueryAny(=)["a" (0)]
      |--CLogicalGbAgg( Global ) Grp Cols: ["a" (0)][Global], Minimal Grp Cols: [], Generates Duplicates :[ 0 ]
      |  |--CLogicalGet "bar" ("bar"), Columns: ["b" (8), "ctid" (9), "xmin" (10), "cmin" (11), "xmax" (12), "cmax" (13), "tableoid" (14), "gp_segment_id" (15)] Key sets: {[1,7]}
      |  +--CScalarProjectList
      +--CScalarIdent "a" (0)

+--CLogicalSelect
   |--CLogicalGet "foo" ("foo"), Columns: ["a" (0), "ctid" (1), "xmin" (2), "cmin" (3), "xmax" (4), "cmax" (5), "tableoid" (6), "gp_segment_id" (7)] Key sets: {[1,7]}
   +--CScalarBoolOp (EboolopAnd)
      |--CScalarCmp (=)
      |  |--CScalarIdent "a" (0)
      |  +--CScalarIdent "a" (0)
      +--CScalarSubqueryExists
         +--CLogicalGbAgg( Global ) Grp Cols: ["a" (0)][Global], Minimal Grp Cols: [], Generates Duplicates :[ 0 ]
            |--CLogicalGet "bar" ("bar"), Columns: ["b" (8), "ctid" (9), "xmin" (10), "cmin" (11), "xmax" (12), "cmax" (13), "tableoid" (14), "gp_segment_id" (15)] Key sets: {[1,7]}
            +--CScalarProjectList
```
While performing the PexprLogicalApply transform, we end up with empty columns
from inner side of the EXISTS query (pcrsInner). The code asserted that
pcrsInner should never be empty. But this should be OK since we don't always
need to select a column from the inner side as can be seen in the above query.
So remove this restriction on pcrsInner.


asubramanya:
Closing since we found issues with the fix.

PR #506, opened 2019-07-01
Fix stats of equi join when join condtion is of type f(a) = b

sambitesh:
This PR consists of two commits. 
- It reverts commit 9769f86237229fb5f173c867c07b644bde884ddc. This revert leads to treating f(a) = b as a = b.

- The second commit, treats f(a) = b as b = constant. So the NDV of join condition is NDV(b).


libnaucrates/include/naucrates/statistics/CStatsPredJoin.h
```diff
@@ -61,15 +65,42 @@ namespace gpnaucrates
 				m_colidOuter(colid1),
 				m_stats_cmp_type(stats_cmp_type),
 				m_colidInner(colid2)
-			{}
+			{
+				m_no_stats_on_outer = false;
```
hardikar: These should go on the initializer list.

libnaucrates/src/statistics/CJoinStatsProcessor.cpp
```diff
@@ -294,33 +294,79 @@ CJoinStatsProcessor::SetResultingJoinStats
...
 		CHistogram *inner_histogram_after = NULL;
-		JoinHistograms
+
+		// When we have any form of equi join with join condition of type f(a)=b,
+		// we calculate the NDV of such a join as NDV(b).
+		if (pred_info->No_stats_on_outer() || pred_info->No_stats_on_inner())
+		{
+			if(pred_info->No_stats_on_inner())
+			{
+				JoinHistograms
```
hardikar: Refactor this to condense it to just one call, by setting some local variables.

libnaucrates/src/statistics/CStatsPredUtils.cpp
```diff
@@ -386,6 +388,9 @@ CStatsPredUtils::IsPredCmpColsOrIgnoreCast
 	(*col_ref_left) = CCastUtils::PcrExtractFromScIdOrCastScId(expr_left);
 	(*col_ref_right) = CCastUtils::PcrExtractFromScIdOrCastScId(expr_right);
 
+	if (NULL == *col_ref_left && NULL == *col_ref_right)
```
hardikar: Was this added for a failure? Seems reasonable, just want to know what prompted you to add this.

libnaucrates/src/statistics/CStatsPredUtils.cpp
```diff
@@ -394,6 +399,18 @@ CStatsPredUtils::IsPredCmpColsOrIgnoreCast
 		// in such cases, check if there is still a possibility to extract scalar ident,
 		// if there is more than one column reference on either side, this is unsupported
 		// If supported, mark the comparison as NDV-based
+
+
+		if (NULL == *col_ref_left)
```
hardikar: This should probably go above the comment, as it stands it's obscuring it's purpose.

libnaucrates/src/statistics/CStatsPredUtils.cpp
sambitesh: I am just trying to stay honest to Sellinger formula.

consider the below case:

```
create table foo (a text);
create table bar (b text);
explain select * from foo inner join bar on lower(a) = lower(b);
```

In this case of f(a) = f(b), we should fall back to default stats according to Sellinger.

libnaucrates/src/statistics/CStatsPredUtils.cpp
```diff
@@ -1102,9 +1119,11 @@ CStatsPredUtils::ExtractJoinStatsFromJoinPred
 
 	const CColRef *col_ref_left = NULL;
 	const CColRef *col_ref_right = NULL;
+	BOOL left_is_null = false;
+	BOOL right_is_null = false;
```
zellerh: Maybe rename left_is_null? From what I can tell, these don't mean that col_ref_left is NULL, it means that col_ref_left didn't come directly from a column, but that instead it was taken from an expression like f(a)? If so, maybe call it something like left_is_from_expr? Same of course goes for right_is_null.

libnaucrates/src/statistics/CStatsPredUtils.cpp
zellerh: It would be good to add that explanation as a comment here.

libnaucrates/src/statistics/CJoinStatsProcessor.cpp
```diff
@@ -294,33 +295,75 @@ CJoinStatsProcessor::SetResultingJoinStats
...
-						&local_scale_factor,
-						is_input_empty,
-						join_type,
-						DoIgnoreLASJHistComputation
-				);
 
-		output_is_empty = JoinStatsAreEmpty(outer_stats->IsEmpty(), output_is_empty, outer_histogram, inner_histogram, outer_histogram_after, join_type);
+		// When we have any form of equi join with join condition of type f(a)=b,
+		// we calculate the NDV of such a join as NDV(b) ( from Selinger et al.)
+		if (CStatsPred::EstatscmptEqNDVOuter == stats_cmp_type)
```
hardikar: I meant doing something like this:
```suggestion
		if (CStatsPred::EstatscmptEqNDVOuter == stats_cmp_type)
		{
		    inner_histogram = outer_histogram;
		} 
		else if (CStatsPred::EstatscmptEqNDVInner == stats_cmp_type)
		{
		    outer_histogram = inner_histogram;
		}
		
		JoinHistograms	
				(	
						mp,	
						outer_histogram,	
						inner_histogram,	
						pred_info,	
						outer_stats->Rows(),	
						inner_side_stats->Rows(),	
						&outer_histogram_after,	
						&inner_histogram_after,	
						&local_scale_factor,	
						is_input_empty,	
						join_type,	
						DoIgnoreLASJHistComputation	
				);
```

libnaucrates/src/statistics/CStatsPredUtils.cpp
```diff
@@ -386,8 +388,42 @@ CStatsPredUtils::IsPredCmpColsOrIgnoreCast
 	(*col_ref_left) = CCastUtils::PcrExtractFromScIdOrCastScId(expr_left);
 	(*col_ref_right) = CCastUtils::PcrExtractFromScIdOrCastScId(expr_right);
 
+	// if the equi join is of type f(a) = f(b) then it is unsupported stats comparison
+	// (from Selinger et al.)
```
hardikar:
```suggestion
	// So, we fall back to default stats. (from Selinger et al.)
```

libnaucrates/src/statistics/CStatsPredUtils.cpp
```diff
@@ -386,8 +388,42 @@ CStatsPredUtils::IsPredCmpColsOrIgnoreCast
...
+		// if the scalar cmp is of equality type, we may not have been able to extract
+		// the column referenes of scalar ident if they had any other expression than cast
+		// on top of them.
+		// in such cases, check if there is still a possibility to extract scalar ident,
+		// if there is more than one column reference on either side, this is unsupported
+		// If supported, mark the comparison as NDV-based
+
+		if (*stats_pred_cmp_type == CStatsPred::EstatscmptEq)
+		{
+			(*col_ref_left) = CUtils::PcrExtractFromScExpression(expr_left);
```
hardikar: Minor: Hmm personally, I'd prefer just implement the logic here than use the function, mainly since the name of the function could be misleading. (I'm talking about PcrExtractFromScExpression)

libnaucrates/include/naucrates/statistics/CStatsPred.h
```diff
@@ -54,7 +54,9 @@ namespace gpnaucrates
 				EstatscmptINDF,	// is not distinct from
 				EstatscmptLike,	// LIKE predicate comparison
 				EstatscmptNotLike,	// NOT LIKE predicate comparison
-
+				// NDV comparision for equality predicate on columns with functions, ex f(a) = b or f(a) = f(b)
```
hardikar:
```suggestion
				// NDV comparision for equality predicate on columns with functions, ex f(a) = b or a = f(b)
```

libnaucrates/include/naucrates/statistics/CStatsPred.h
```diff
@@ -54,7 +54,9 @@ namespace gpnaucrates
...
+				EstatscmptEqNDVOuter, // case when we have to consider NDV of only Outer
```
hardikar:
```suggestion
				EstatscmptEqNDVOuter, // use Outer NDV on inner side also
				EstatscmptEqNDVInner, // use Inner NDV on outer side also
```

libnaucrates/include/naucrates/statistics/CStatsPredJoin.h
```diff
@@ -61,7 +61,8 @@ namespace gpnaucrates
 				m_colidOuter(colid1),
 				m_stats_cmp_type(stats_cmp_type),
 				m_colidInner(colid2)
-			{}
+			{
```
ashuka24: unnecessary change

libnaucrates/include/naucrates/statistics/CStatsPred.h
sambitesh: Done

libnaucrates/include/naucrates/statistics/CStatsPred.h
sambitesh: Done

libnaucrates/src/statistics/CStatsPredUtils.cpp
sambitesh: Done

libnaucrates/src/statistics/CJoinStatsProcessor.cpp
sambitesh: Implemented thanks

libnaucrates/src/statistics/CJoinStatsProcessor.cpp
```diff
@@ -294,33 +294,79 @@ CJoinStatsProcessor::SetResultingJoinStats
...
 		GPOS_ASSERT(NULL != inner_histogram);
 		BOOL is_input_empty = CStatistics::IsEmptyJoin(outer_stats, inner_side_stats, IsLASJ);
-
 		CDouble local_scale_factor(1.0);
 		CHistogram *outer_histogram_after = NULL;
 		CHistogram *inner_histogram_after = NULL;
-		JoinHistograms
+
+		// When we have any form of equi join with join condition of type f(a)=b,
+		// we calculate the NDV of such a join as NDV(b).
```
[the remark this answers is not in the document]
sambitesh: Done

libnaucrates/src/statistics/CStatsPredUtils.cpp
```diff
@@ -338,7 +338,9 @@ CStatsPredUtils::IsPredCmpColsOrIgnoreCast
 	CExpression *expr,
 	const CColRef **col_ref_left,
 	CStatsPred::EStatsCmpType *stats_pred_cmp_type,
-	const CColRef **col_ref_right
+	const CColRef **col_ref_right,
+	BOOL &left_is_null,
```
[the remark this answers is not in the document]
sambitesh: This is a bool hence the reference object is going to be very small.

libnaucrates/src/statistics/CStatsPredUtils.cpp
sambitesh: Done, thanks for suggestion

PR #507, opened 2019-07-02
Fix constraint simplification for inferred predicates

asubramanya:
When we encounter a CExpression that cannot be converted back to a
CConstraintInterval, just return the original constraint without
simplifying it - instead of asserting.


libgpopt/src/base/CConstraint.cpp
```diff
@@ -780,6 +780,14 @@ CConstraint::PdrgpcnstrDeduplicate
...
 		CConstraint *pcnstrNew = CConstraintInterval::PciIntervalFromScalarExpr(mp, pexpr, colref);
+		if (NULL == pcnstrNew)
+		{
+			// We ran into a type conflict that prevents us from using this method to simplify the constraint.
+			// Give up and return the un-flattened constraint.
+			pcnstrChild->AddRef();
+			pcnstrNew = pcnstrChild;
+		}
+
 		GPOS_ASSERT(NULL != pcnstrNew);
```
vraghavan78: do we need this assert then?

asubramanya: Yeah. We can remove this.

asubramanya:
@hardikar We tried the query you mentioned, but did not hit our code change at all. Its true for both the string and the integer versions of the query. We will add some more comment in the code.

asubramanya:
@chrishajas We will investigate the reason for not being able to convert the expression into `CConstraintInterval` separately. For now this fix should be fine since there is no reason to crash in this case.

PR #508, opened 2019-07-03
Remove broadcast to allow for Left Outer Index NLJ

ashuka24:
This also fixes Orca issue #309.

Say that we have the following tables:
```
foo distributed by (a,b)
bar distributed by (c,d) with an index on d
```
For the query
```
select * from foo join bar on a = c and b = d
```
It will generate a PhysicalFilter on top of an IndexGet on bar. In this situation,
neither the PhysicalFilter nor the IndexGet can create a matching
distribution spec (a,b) to the outer side in order to generate a plan
using a index NLJ.

Thus, this commit allows the IndexGet to create an intermediary
distribution spec, where it matches only the columns that it can. Then
at the PhysicalFilter, the filter will complete the incomplete spec
with its knowledge.

The result would look something like this: 
```
Filter --- equivalent spec would be (a,b)
  |-- Index Scan ---- equivalent spec would be (c, b) 
  |      +-- foo.c = bar.d --- original spec here is (c,d)
  +-- foo.a = bar.c
```


libgpopt/src/operators/CPhysicalFilter.cpp
```diff
@@ -330,12 +330,64 @@ CPhysicalFilter::PosDerive
...
+		CDistributionSpecHashed *pdshashedIncomplete = pdshashedOriginal->PdshashedEquiv();
+		// complete the spec if the incomplete equivalent exists
+		if (NULL != pdshashedIncomplete)
+		{
+			CExpressionArray *pdrgpexprHashed = pdshashedIncomplete->Pdrgpexpr();
+			const ULONG size = pdrgpexprHashed->Size();
+
+			CExpressionArray *pdrgpexprResult = GPOS_NEW(mp) CExpressionArray(mp);
+
+			BOOL fSuccess = true;
```
zellerh: One way to simplify this would be to get rid of the fSuccess variable.

libgpopt/src/operators/CPhysicalFilter.cpp
```diff
@@ -330,12 +330,64 @@ CPhysicalFilter::PosDerive
...
+			for (ULONG ul = 0; ul < size; ul++)
+			{
+
+				CExpression *pexpr = (*pdrgpexprHashed)[ul];
+				CExpression *pexprMatching = CUtils::PexprMatchEqualityOrINDF(pexpr, pdrgpexpr);
+				fSuccess = (NULL != pexprMatching);
+				if (fSuccess)
```
zellerh:
```suggestion
				if (NULL != pexprMatching)
```

libgpopt/src/operators/CPhysicalScan.cpp
```diff
@@ -174,42 +175,41 @@ CPhysicalScan::PdshashedDeriveWithOuterRefs
 	CExpression *pexprIndexPred = exprhdl.PexprScalarChild(0 /*child_index*/);
 	CExpressionArray *pdrgpexpr = CPredicateUtils::PdrgpexprConjuncts(mp, pexprIndexPred);
 
-	CExpressionArray *pdrgpexprMatching = GPOS_NEW(mp) CExpressionArray(mp);
+	CExpressionArray *pdrgpexprResult = GPOS_NEW(mp) CExpressionArray(mp);
 	CDistributionSpecHashed *pdshashed = CDistributionSpecHashed::PdsConvert(m_pds);
 	CExpressionArray *pdrgpexprHashed = pdshashed->Pdrgpexpr();
 	const ULONG size = pdrgpexprHashed->Size();
 
 	BOOL fSuccess = true;
```
zellerh: Same as above, we could get rid of the fSuccess variable.

libgpopt/src/operators/CPhysicalScan.cpp
```diff
@@ -174,42 +175,41 @@ CPhysicalScan::PdshashedDeriveWithOuterRefs
...
 	pdrgpexpr->Release();
 
-	if (fSuccess)
-	{
-		GPOS_ASSERT(pdrgpexprMatching->Size() == pdrgpexprHashed->Size());
 
-		// create a matching hashed distribution request
-		BOOL fNullsColocated = pdshashed->FNullsColocated();
-		CDistributionSpecHashed *pdshashedEquiv = GPOS_NEW(mp) CDistributionSpecHashed(pdrgpexprMatching, fNullsColocated);
+	GPOS_ASSERT(pdrgpexprResult->Size() == pdrgpexprHashed->Size());
```
zellerh: It might be nice to recognize the case where pdrgpexprResult is the same as pdrgpexprHashed. In that case we don't need to add an equivalent distribution spec that would just be a copy of the original one. However, if we do that, we might also need to rethink the logic in CPhysicalFilter::PdsDerive() that only kicks in if there is an equivalent PDS (see comment in that method).

libgpopt/src/operators/CPhysicalInnerIndexNLJoin.cpp
```diff
@@ -133,11 +133,26 @@ CPhysicalInnerIndexNLJoin::PdsRequired
 		CDistributionSpecHashed *pdshashedEquiv = pdshashed->PdshashedEquiv();
 		if (NULL != pdshashedEquiv)
```
zellerh: What if there are no predicates on the scan and all the predicates are on the filter? In that case, maybe we want to work with the original distribution spec, if the scan didn't produce an equivalent one (see comment on CPhysicalScan::PdshashedDeriveWithOuterRefs()).

libgpopt/src/operators/CPhysicalInnerIndexNLJoin.cpp
```diff
@@ -133,11 +133,26 @@ CPhysicalInnerIndexNLJoin::PdsRequired
...
-			pdsHashedRequired->ComputeEquivHashExprs(mp, exprhdl);
-			return pdsHashedRequired;
+			CColRefSet *pcrshashed = pdshashed->PcrsUsed(mp);
+			CColRefSet *pcrshashedEquiv = pdshashedEquiv->PcrsUsed(mp);
+
+			// if the distribution spec is a complete spec, then proceed
+			// it is possible to get incomplete specs (see CPhysicalScan::PdshashedDeriveWithOuterRefs)
+			// For a spec to be considered complete, the equivalent spec should not contain any cols
+			// from the original distribution
+			if (!pcrshashedEquiv->FIntersects(pcrshashed))
```
hardikar: This logic is used in multiple places. Can we add a new method in HashedSpec that encapsulates it?

libgpopt/src/operators/CPhysicalScan.cpp
```diff
@@ -152,14 +152,15 @@ CPhysicalScan::EpetOrder
...
-//---------------------------------------------------------------------------
+// Derive hashed distribution when predicates have outer references
+// This function allows for incomplete distribution specs. This means
+// that it is possible for a spec to have a column from both the inner
+// and the outer table.
+// For example, if we have foo (a, b) distributed by (a,b)
+//                         bar (c, d) distributed by (c, d)
+// with an index on idx_bar_d, if we have the query
+// select * from foo join bar on a = c and b = d,
+// it is possible to get a spec of [a, d].
```
hardikar: Add another line about when the "incomplete" spec is completed

libgpopt/src/operators/CPhysicalInnerIndexNLJoin.cpp
```diff
@@ -133,11 +133,19 @@ CPhysicalInnerIndexNLJoin::PdsRequired
...
-			// request hashed distribution from outer
-			pdshashedEquiv->Pdrgpexpr()->AddRef();
-			CDistributionSpecHashed *pdsHashedRequired = GPOS_NEW(mp) CDistributionSpecHashed(pdshashedEquiv->Pdrgpexpr(), pdshashedEquiv->FNullsColocated());
-			pdsHashedRequired->ComputeEquivHashExprs(mp, exprhdl);
-			return pdsHashedRequired;
+			// if the distribution spec is a complete spec, then proceed
+			// it is possible to get incomplete specs (see CPhysicalScan::PdshashedDeriveWithOuterRefs)
+			// For a spec to be considered complete, the equivalent spec should not contain any cols
+			// from the original distribution
+			if (!pdshashed->FColsIntersects(mp, pdshashedEquiv))
```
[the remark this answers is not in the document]
bhuvnesh2703: identify at what stage to discard an incomplete spec is gonna be tricky.. so this is fine...

libgpopt/src/operators/CPhysicalFilter.cpp
```diff
@@ -330,12 +330,64 @@ CPhysicalFilter::PosDerive
...
 	(
-	CMemoryPool *, // mp
+	CMemoryPool *mp,
 	CExpressionHandle &exprhdl
 	)
 	const
 {
-	return PdsDerivePassThruOuter(exprhdl);
+	CDistributionSpec *pdsChild = PdsDerivePassThruOuter(exprhdl);
+	if (CDistributionSpec::EdtHashed == pdsChild->Edt() && exprhdl.HasOuterRefs())
```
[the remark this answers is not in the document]
hardikar: So it works like an extra gate? The `exprhdl.HasOuterRefs()` check is not necessary, but prevents us from doing work when it's not true, because we're sure it's not relevant?

libgpopt/src/operators/CPhysicalLeftOuterIndexNLJoin.cpp
```diff
@@ -97,7 +97,8 @@ CPhysicalLeftOuterIndexNLJoin::PdsRequired
 			// it is possible to get incomplete specs (see CPhysicalScan::PdshashedDeriveWithOuterRefs)
 			// For a spec to be considered complete, the equivalent spec should not contain any cols
 			// from the original distribution
-			if (!pdshashed->FColsIntersects(mp, pdshashedEquiv))
+
+			if (!pdshashed->FColsIntersects(mp, pdshashedEquiv) || pdshashedEquiv->Matches(pdshashed))
```
bhuvnesh2703: why Matches?

libgpopt/src/operators/CPhysicalFilter.cpp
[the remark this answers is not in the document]
bhuvnesh2703: you can say that..

libgpopt/src/operators/CPhysicalFilter.cpp
ashuka24: Done

libgpopt/src/operators/CPhysicalScan.cpp
ashuka24: Done

libgpopt/src/operators/CPhysicalScan.cpp
ashuka24: Done

libgpopt/src/operators/CPhysicalScan.cpp
```diff
@@ -174,42 +176,39 @@ CPhysicalScan::PdshashedDeriveWithOuterRefs
...
+	// create a matching hashed distribution request
+	BOOL fNullsColocated = pdshashed->FNullsColocated();
+	CDistributionSpecHashed *pdshashedEquiv = GPOS_NEW(mp) CDistributionSpecHashed(pdrgpexprResult, fNullsColocated);
 
-	pdrgpexprMatching->Release();
+	pdrgpexprHashed->AddRef();
+	CDistributionSpecHashed *pdshashedResult = GPOS_NEW(mp) CDistributionSpecHashed(pdrgpexprHashed, fNullsColocated, pdshashedEquiv);
 
-	return NULL;
+	return pdshashedResult;
```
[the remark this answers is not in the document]
ashuka24: An incomplete spec is only relevant when we have an index join on a	multi-key distributed table. In such cases, the entire distribution key of the table must be present in the join clause to produce the index scan alternative in the first place (see `CXformJoin2IndexApplyBase`). Therefore, when an incomplete spec is created here for such a table, there will always be a CPhysicalFilter (that has the remaining predicates) on top to complete the spec.

libgpopt/src/operators/CPhysicalLeftOuterIndexNLJoin.cpp
ashuka24: Got rid of this

PR #510, opened 2019-07-11
Remove multi-threading code

zellerh:
ORCA was written to do query optimization using multiple threads, but
this feature is not used, not well tested and likely not working right
now. We would like to remove this code, as this promises to lead to
performance improvements. Our test showed a 12% decrease in the time
for explaining the TPC-DS queries.

There are multiple classes we are removing:

- `CMutex`,
`CSpinlockBase`,
`CSpinlockOS`,
`CSpinlockCache`,
`CSpinlockDummy`:  Mutexes and spinlocks, used to obtain locks on objects.

- `CEvent`:          Event to wait on and to wake up threads
	  
- `CAutoMutex`,
`CAutoSpinlock`:   Objects allocated on the stack, used to lock a mutex or spinlock and automatically unlock it at the end of the scope.

- `CWorkerId`:       Relies on thread id, no longer needed as we only have a single worker
- `CThreadManager`: Low-level code to create and stop threads

Tests related to these classes are removed as well.

We are also removing the atomic operations defined in gpos/sync/atomic.h,
as well the `volatile` declaration of many data member used with these
atomic operations:

- `ExchangeAddUlongPtrWithInt()` and related
- `CompareSwap()`

The following classes no longer are thread-safe, but we kept the "Sync" in their names, since renaming
them would be disruptive:

- `CSyncHashTable`
- `CSyncList`
- `CSyncPool`

These classes now have reduced functionality:

- `CAutoTaskProxy`
- `CWorkerPoolManager`
- `CWorker`
- `CTask`
- `CScheduler`

We might be able to eliminate some of these, but so far we have not yet attempted that.

Finally, a debug-only feature to track timeslices in workers is also removed.

You can review the three commits independently, if you prefer.


ivannovick:
LOVE IT!

hardikar:
Super excited to see this! Using the multi-threaded code path did lead to crashes.
Quick question before I review, were there no GPDB side changes?

zellerh:
@hardikar, this needs a one line change on the GPDB side, removing the #include for `CSpinlock.h`  from file `src/include/gpopt/utils/COptServer.h`.

libgpos/server/src/unittest/gpos/common/CSyncListTest.cpp
```diff
@@ -120,280 +119,4 @@ CSyncListTest::EresUnittest_Basics()
...
-//---------------------------------------------------------------------------
-//	@function:
-//		CSyncListTest::RunPush
-//
-//	@doc:
-//		Iteratively push elements to list
-//
-//---------------------------------------------------------------------------
-void *
-CSyncListTest::RunPush
```
chrishajas: Is this (and below) test still relevant? It seems like it's just pushing elements to the list and not dealing with concurrency.

libgpos/server/src/unittest/gpos/memory/CCacheTest.cpp
```diff
@@ -1015,194 +1014,4 @@ CCacheTest::EresUnittest_IterativeDeletion()
...
-//---------------------------------------------------------------------------
-//	@function:
-//		CCacheTest::PvLookupTask
-//
-//	@doc:
-//		A task that iterates over entries in the cache
-//
-//---------------------------------------------------------------------------
-void *
-CCacheTest::PvLookupTask
```
chrishajas: Same here, are the non-concurrent tests relevant?

libgpos/src/error/CErrorContext.cpp
```diff
@@ -14,6 +14,7 @@
 #include "gpos/error/CMessageRepository.h"
 #include "gpos/error/CMiniDumper.h"
 #include "gpos/error/CSerializable.h"
+#include "gpos/io/ioutils.h"
```
hardikar: is this needed?

libgpos/src/error/CLogger.cpp
```diff
@@ -165,7 +160,7 @@ CLogger::Format
 		m_entry_wrapper.AppendFormat
 			(
 			GPOS_WSZ_LIT(",THD%03d,%s,\"%ls\",\n"),
-			thread_id,
+			1, //thread id
```
hardikar: or 0? I though I saw 0 used somewhere else.

libgpos/src/task/CAutoTaskProxy.cpp
```diff
@@ -103,13 +102,11 @@ CAutoTaskProxy::Destroy
...
 	)
 {
-	GPOS_ASSERT(OwnerOf(task) && "Task not owned by this ATP object");
-
 	// cancel scheduled task
 	if (task->IsScheduled() && !task->IsReported())
 	{
 		Cancel(task);
-		Wait(task);
+		CheckError(task);
```
hardikar: Earlier, Wait() also called SetReported(). That function is still around (I think)

hardikar: Or maybe we don't need reported status anymore..

libgpos/src/task/CWorkerPoolManager.cpp
```diff
@@ -258,58 +168,10 @@ CWorkerPoolManager::RegisterWorker
...
-	CWorkerId wid
-	)
+void
+CWorkerPoolManager::RemoveWorker()
 {
-	// get access
-	CSyncHashtableAccessByKey<CWorker, CWorkerId, CSpinlockOS> shta(m_shtWLS, wid);
-
-	return shta.Find();
+	m_single_worker = NULL;
```
hardikar: There's no need to release/delete the worker?

libgpopt/include/gpopt/search/CGroup.h
```diff
@@ -85,8 +82,7 @@ namespace gpopt
 			typedef
 				CSyncHashtable<
 					COptimizationContext, // entry
-					COptimizationContext, // search key
-					CSpinlockOC> ShtOC;
+					COptimizationContext> ShtOC;
```
ashuka24: do we want to keep the search key comment here?

libgpopt/include/gpopt/search/CGroupExpression.h
```diff
@@ -76,8 +75,7 @@ namespace gpopt
 			typedef
 				CSyncHashtable<
 					CCostContext, // entry
-					OPTCTXT_PTR, // search key
-					CSpinlockCC> ShtCC;
+					OPTCTXT_PTR> ShtCC;
```
ashuka24: same here, do we want to keep the search key comment here?

libgpos/src/error/CErrorHandlerStandard.cpp
```diff
@@ -13,6 +13,7 @@
 #include "gpos/error/CErrorContext.h"
 #include "gpos/error/CErrorHandlerStandard.h"
 #include "gpos/error/CLogger.h"
+#include "gpos/io/ioutils.h"
```
ashuka24: same as shreedhar's aboce comment, is this needed here?

libgpos/src/error/CLogger.cpp
```diff
@@ -15,6 +15,7 @@
 #include "gpos/error/CLoggerStream.h"
 #include "gpos/error/CLoggerSyslog.h"
 #include "gpos/error/CMessageRepository.h"
+#include "gpos/io/iotypes.h"
```
ashuka24: necessary include?

libgpopt/src/engine/CEngine.cpp
```diff
@@ -1774,122 +1735,33 @@ CEngine::MainThreadOptimize()
...
-		PssCurrent()->SetBestExpr(pexprPlan);
 
-		FinalizeSearchStage();
+	if (optimizer_config->GetEnumeratorCfg()->FSample())
+	{
+		SamplePlans();
 	}
 }
 
+
```
ashuka24: remove extra line

libgpopt/src/search/CScheduler.cpp
```diff
@@ -352,24 +289,7 @@ CScheduler::Schedule
 	(void) ExchangeAddUlongPtrWithInt(&m_ulpQueued, 1);
```
ashuka24: change to `m_ulpQueued++`

libgpos/include/gpos/error/CAutoExceptionStack.h
```diff
@@ -19,6 +19,7 @@
 #define GPOS_CAutoExceptionStack_H
 
 #include "gpos/base.h"
+#include "gpos/common/CStackObject.h"
```
ashuka24: why is this needed?

libgpos/src/memory/CMemoryPoolStack.cpp
```diff
@@ -110,7 +110,7 @@ CMemoryPoolStack::Allocate
 	// find block to allocate memory in it
 	SBlockDescriptor *desc = FindMemoryBlock(as, alloc);
 
-	GPOS_ASSERT_IMP(IsThreadSafe(), m_lock.IsOwned());
+	//GPOS_ASSERT_IMP(IsThreadSafe(), m_lock.IsOwned());
```
ashuka24: delete?

libgpos/src/memory/CMemoryPoolStack.cpp
```diff
@@ -216,7 +216,7 @@ CMemoryPoolStack::New
 void
 CMemoryPoolStack::TearDown()
 {
-	GPOS_ASSERT(!m_lock.IsOwned());
+	//GPOS_ASSERT(!m_lock.IsOwned());
```
ashuka24: delete?

libnaucrates/include/naucrates/traceflags/traceflags.h
```diff
@@ -86,7 +86,7 @@ namespace gpos
 		///////////////////////////////////////////////////////
 
 		// use threads in optimization engine
-		EopttraceParallel = 103000,
+		EopttraceParallel = 103000, // no longer used
```
ashuka24: is there a reason we are keeping this traceflag if it is no longer used?

ashuka24: this traceflag also exists in CEngineTest.cpp. Does it need to be removed there at all?

zellerh:
@hardikar, you asked whether we could replace CSyncHashTable with CHashMap, since they now provide very similar functions. Unfortunately, that could get quite complicated, since the interfaces are different in several ways. Here are really simplified class declarations:

```
template <class T, class K> class CSyncHashtable
{

    // access to the hash table is done via iterators
    friend class CSyncHashtableAccessByKey<T, K>;
    friend class CSyncHashtableAccessByIter<T, K>;
    friend class CSyncHashtableIter<T, K>;

    void Init
        (
        CMemoryPool *mp,
        ULONG size,
        // hash table contents have embedded links
        ULONG link_offset,
        ULONG key_offset,
        const K *invalid_key,
        // hash and comparison functions deal with references
        ULONG (*func_hash)(const K&),
        BOOL (*func_equal)(const K&, const K&)
        )

    void DestroyEntries(DestroyEntryFuncPtr pfunc_destroy);
    // the key is embedded in the value
    void Insert(T *value);
};


template <class K, class T, 
          ULONG (*HashFn)(const K*), 
          BOOL (*EqFn)(const K*, const K*),
          void (*DestroyKFn)(K*),
          void (*DestroyTFn)(T*)> class CHashMap : public CRefCount
{
    BOOL Insert(K *key, T *value);
    T *Find(const K *key) const;
    BOOL Replace(const K *key, T *ptNew);
};
```

If we want to change these to a common interface, maybe another option might be to switch to `std::unordered_map` instead.

9DemonFox:
no wander the autotrace has only THREAD 0000

PR #511, opened 2019-07-12
On demand property derivation (relational)

hardikar:
The overall purpose of this PR to reduce memory consumption & improve optimization time.
On some queries it gives a ~30% boost to both memory & time. We intend to follow this pattern
(outlined in the PR for relational properties), for scalar properties as well, since we see that as 
the next bottleneck. We are opening the PR now to get early feedback - hence the WIP.

Note that it is not possible to make the property derivation be on-demand 100% of the time. There
are instances in ORCA logic, where the derived properties must access/saved independant of the expression used to derive them. In such cases, it is necessary to complete derive the properties.

Please look at the PR commit-by-commit. It'll make no sense otherwise.
All CTESTs pass on every commit (AFAIK). That's what we've been using to test correctness. ICG, Workload & DS pipelines are currently running as well.


libgpopt/src/base/CCostContext.cpp
```diff
@@ -379,6 +379,7 @@ CCostContext::BreakCostTiesForJoinPlans
 
 	// both plans have equal estimated rows for both children, break tie based on join depth
 	*pfTiesResolved = true;
+	// XXX: do something here?
```
zellerh: Remove this?

libgpopt/src/operators/CExpressionHandle.cpp
```diff
@@ -1281,13 +1118,13 @@ CExpressionHandle::GetRelationalProperties() const
...
 
-	if (NULL != m_pcc)
+	if (NULL != m_pcc || NULL != m_pgexpr)
 	{
 		// get relational props from group
 		return CDrvdPropRelational::GetRelationalProperties(Pgexpr()->Pgroup()->Pdp());
 	}
 
-	return CDrvdPropRelational::GetRelationalProperties(m_pdp);
+	GPOS_RAISE(gpopt::ExmaGPOPT, gpopt::ExmiUnsatisfiedRequiredProperties);
```
hardikar: Wrong exception type.

libgpopt/src/operators/CExpressionHandle.cpp
```diff
@@ -1328,11 +1165,22 @@ CExpressionHandle::Pdpplan
...
+		return CDrvdPropPlan::Pdpplan((*m_pexpr)[child_index]->Pdp(DrvdPropArray::EptPlan));
+	}
 
-	DrvdPropArray *pdp = (*m_pdrgpdp)[child_index];
+	if (NULL != m_pcc)
+	{
+		COptimizationContext *pocChild = (*m_pcc->Pdrgpoc())[child_index];
+		CDrvdPropPlan *pdpplan = pocChild->PccBest()->Pdpplan();
+
+		// CDrvdPropCtxt::AddDerivedProps(pdpplan, pdpctxtplan);
```
hardikar: Remove this.

libgpopt/include/gpopt/base/CDrvdPropRelational.h
```diff
@@ -120,47 +144,63 @@ namespace gpopt
...
 			}
 
+			virtual
+			BOOL IsComplete() const { return m_is_complete; }
+
 			// derivation function
 			void Derive(CMemoryPool *mp, CExpressionHandle &exprhdl, CDrvdPropCtxt *pdpctxt);
 
 			// output columns
 			CColRefSet *PcrsOutput() const;
```
hardikar: GPOS_ASSERT that this is called on complete pdps only.

libgpopt/include/gpopt/base/CDrvdPropRelational.h
```diff
@@ -120,47 +144,63 @@ namespace gpopt
...
+			CColRefSet *PcrsOutput(CExpressionHandle &);
```
hardikar: Make these protected and then befriend Expresssion/ExpresionHandle?

libgpopt/include/gpopt/base/CDrvdPropRelational.h
hardikar: Rename to PcrsOutputNoDerive, PcrsOutputNoCheck or something?

libgpopt/src/operators/CExpression.cpp
```diff
@@ -569,30 +575,35 @@ CExpression::PdpDerive
...
 			default:
 				break;
 		}
 
 		Pdp(ept)->Derive(m_mp, exprhdl, pdpctxt);
 	}
 
+	else if (!Pdp(ept)->IsComplete())
+	{
+		Pdp(ept)->Derive(m_mp, exprhdl, pdpctxt);
```
hardikar: Add more comments here.

libgpopt/src/operators/CExpression.cpp
```diff
@@ -1572,4 +1583,136 @@ CExpression::FValidPartEnforcers
 	return true;
 }
 
+CColRefSet *
+CExpression::PcrsOuter()
+{
+	if (m_pdprel == NULL)
```
hardikar: Can we move this to the constructor?

libgpopt/src/operators/CExpressionHandle.cpp
```diff
@@ -1095,7 +1096,6 @@ CExpressionHandle::GetRelationalProperties
 	GPOS_RAISE(gpopt::ExmaGPOPT, gpopt::ExmiUnsatisfiedRequiredProperties);
 }
 
-
```
hardikar: Assert in CExpressionHandle ::GetRelationalProperties() that the returned property is complete. Probably litter these asserts everywhere.

libgpopt/src/base/CCostContext.cpp
chrishajas: done!

libgpopt/include/gpopt/operators/CExpressionHandle.h
```diff
@@ -301,6 +281,8 @@ namespace gpopt
 			// as columns used by all its scalar children
 			CColRefSet *PcrsUsedColumns(CMemoryPool *mp);
 
+		friend class CExpression;
```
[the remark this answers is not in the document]
chrishajas: done

libgpopt/src/base/CDrvdPropRelational.cpp
```diff
@@ -373,7 +355,7 @@ CDrvdPropRelational::OsPrint
 	)
 	const
 {
-	os	<<	"Output Cols: [" << *m_pcrsOutput << "]"
+	os	<<	"Output Cols: [" << *PcrsOutput() << "]"
 		<<	", Outer Refs: [" << *m_pcrsOuter << "]"
```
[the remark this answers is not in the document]
chrishajas: done

libgpopt/src/base/CDrvdPropRelational.cpp
```diff
@@ -414,5 +396,271 @@ CDrvdPropRelational::OsPrint
...
+{
+	return m_pdrgpfd;
+}
+
+CFunctionalDependencyArray *
+CDrvdPropRelational::Pdrgpfd(CExpressionHandle &exprhdl)
+{
+	if (!m_is_prop_derived->ExchangeSet(EdptPdrgpfd))
+	{
+		// XXX: Am I really needed?
```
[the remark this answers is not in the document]
chrishajas: done

libgpopt/src/operators/CExpressionHandle.cpp
```diff
@@ -1712,7 +1559,7 @@ CExpressionHandle::PfpChild
 //
 //---------------------------------------------------------------------------
 BOOL
-CExpressionHandle::FChildrenHaveVolatileFuncScan() const
+CExpressionHandle::FChildrenHaveVolatileFuncScan()
```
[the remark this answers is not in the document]
chrishajas: The header file still includes the const, so removing this was unintentional. I'm adding it back.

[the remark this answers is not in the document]
chrishajas: `FChildrenHaveVolatileFuncScan` calls `PfpChild`, which now can cache the results of derive. Therefore, since `PfpChild` can't be const, `FChildrenHaveVolatileFuncScan` also can't be const.

libgpopt/include/gpopt/operators/CExpressionHandle.h
```diff
@@ -301,6 +281,44 @@ namespace gpopt
 			// as columns used by all its scalar children
 			CColRefSet *PcrsUsedColumns(CMemoryPool *mp);
 
+			CColRefSet *PcrsOuter();
+			CColRefSet *PcrsOuter(ULONG i);
```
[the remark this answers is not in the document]
chrishajas: Done, thanks!

libgpopt/src/base/CDrvdPropRelational.cpp
```diff
@@ -91,77 +98,52 @@ CDrvdPropRelational::~CDrvdPropRelational()
...
-	{
-		m_fHasPartialIndexes =
-				CLogicalDynamicGet::PopConvert(popLogical)->Ptabdesc()->FHasPartialIndexes();
-	}
-	else if (COperator::EopLogicalSelect == op_id)
-	{
-		m_fHasPartialIndexes =
-				exprhdl.GetRelationalProperties(0 /*child_index*/)->FHasPartialIndexes();
-	}
+	m_is_complete = true;
```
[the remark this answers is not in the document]
chrishajas: If we end up deriving everything separately, we won't mark `m_is_complete` as `true`. However, we only check `m_is_complete` in `PdpDerive()`. Even though we would attempt to derive everything as a result, since each individual member has been cached, it won't add any measurable time.

libgpopt/src/operators/CExpressionHandle.cpp
```diff
@@ -861,8 +701,9 @@ CExpressionHandle::DerivePlanProps
 	pdpctxtplan->SetExpectedPartitionSelectors(pop, m_pcc);
 
 	// create/derive local properties
-	m_pdp = Pop()->PdpCreate(m_mp);
-	m_pdp->Derive(m_mp, *this, pdpctxtplan);
+	// MARKED
```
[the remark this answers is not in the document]
chrishajas: I'm not sure, I'll remove it.

libgpopt/src/operators/CExpressionHandle.cpp
```diff
@@ -1929,4 +1776,297 @@ CExpressionHandle::PcrsUsedColumns
...
+
+IStatistics *
+CExpressionHandle::Pstats()
+{
+	return m_pstats;
+}
+
+
+CColRefSet *
+CExpressionHandle::PcrsOuter(ULONG i)
```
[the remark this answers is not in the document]
chrishajas: Done, thanks!

libgpopt/src/operators/CExpressionHandle.cpp
```diff
@@ -1245,20 +1086,15 @@ CExpressionHandle::GetRelationalProperties
...
 		// handle is used for deriving plan properties, get relational props from child group
 		return CDrvdPropRelational::GetRelationalProperties((*Pgexpr())[child_index]->Pdp());
 	}
 
-	GPOS_ASSERT(child_index < m_pdrgpdp->Size());
-
-	DrvdPropArray *pdp = (*m_pdrgpdp)[child_index];
-
-	return CDrvdPropRelational::GetRelationalProperties(pdp);
+	GPOS_RAISE(gpopt::ExmaGPOPT, gpopt::ExmiUnsatisfiedRequiredProperties);
```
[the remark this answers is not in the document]
chrishajas: We're doing the same thing with this code block:
```
	if (NULL != m_pgexpr)
	{
		return CDrvdPropScalar::GetDrvdScalarProps((*m_pgexpr)[child_index]->Pdp());
	}
```

Previously we didn't have the if check, but we're just trying to confirm assumptions and ensure we don't crash (eg: that since the handle had no expression and no cost context, the only other possibility is a group expression).

libgpopt/include/gpopt/base/CDrvdPropRelational.h
chrishajas: @hardikar did we decide on retail asserts vs regular asserts in the const functions? A regular assert caught a place we missed, but I haven't done any perf testing.

Also, any preference on naming? I think I prefer`PcrsOutputNoDerive` (or even `PcrsOutputGet`, but NoDerive is probably more descriptive.

libgpopt/src/operators/CExpressionHandle.cpp
chrishajas: Added asserts in a few places, let me know if there's others you think we should add them.

libgpopt/src/operators/CExpressionHandle.cpp
chrishajas: Added a new exception type `ExmiIncompletePropSpec` with message `No valid context to derive properties from`, open to suggestions on naming/messages.

libgpopt/src/operators/CExpression.cpp
chrishajas: Moved these to the constructor. Had to make a couple of other changes when we reset/copy properties, let me know if it still looks ok (it's in Address PR feedback 4)

chrishajas:
@hardikar @asubramanya @ashuka24 I think I've addressed all the comments. The last few commits just do mass renamings. Please review the additional commits that addressed review feedback.

libgpopt/src/exception.cpp
```diff
@@ -94,6 +94,12 @@ gpopt::EresExceptionInit
 					 GPOS_WSZ_WSZLEN("CTE Producer-Consumer execution locality mismatch for CTE id %lld"),
 					 1,
 					 GPOS_WSZ_WSZLEN("CTE Producer-Consumer execution locality mismatch")),
+
+			CMessage(CException(gpopt::ExmaGPOPT, gpopt::ExmiIncompleteDerivedProperty),
+					 CException::ExsevError,
+					 GPOS_WSZ_WSZLEN("No valid context to derive properties from"),
```
ashuka24: these look like unfinished sentences.

libgpopt/src/operators/CLogicalBitmapTableGet.cpp
```diff
@@ -130,14 +130,14 @@ CLogicalBitmapTableGet::Matches
 
 //---------------------------------------------------------------------------
 //	@function:
-//		CLogicalBitmapTableGet::PcrsDeriveOutput
+//		CLogicalBitmapTableGet::DeriveOutputColumns
```
hardikar: Not sure if this change is intentional.

libgpopt/src/operators/CExpression.cpp
hardikar: LGTM

libgpopt/src/base/CDrvdPropRelational.cpp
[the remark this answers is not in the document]
hardikar: Added this myself.

libgpopt/src/operators/CExpressionHandle.cpp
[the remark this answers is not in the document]
hardikar: Is this the right exception to raise here?

libgpopt/include/gpopt/base/CDrvdPropRelational.h
```diff
@@ -47,11 +47,34 @@ namespace gpopt
...
+		friend class CExpression;
+
+		enum EDrvdPropType
+		{
+			EdptPcrsOutput = 0,
+			EdptPcrsOuter,
+			EdptPcrsNotNull,
+			EdptPcrsCorrelatedApply,
+			EdptPkc,
+			EdptPdrgpfd,
```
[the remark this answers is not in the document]
hardikar: Hmm. We could. Honestly, I wonder if we should rename these enums.
For some reference, their corresponding `m_pkc` etc member variables have comments on them.

libgpopt/include/gpopt/operators/CExpressionHandle.h
```diff
@@ -243,25 +226,24 @@ namespace gpopt
...
-				return (0 < GetRelationalProperties()->PcrsOuter()->Size());
+				return (0 < DeriveOuterReferences()->Size());
 			}
 
 			// check if attached expression must execute on a single host
-			BOOL NeedsSingletonExecution() const
+			BOOL NeedsSingletonExecution()
 			{
-				return GetRelationalProperties()->Pfp()->NeedsSingletonExecution();
+				return this->DeriveFunctionProperties()->NeedsSingletonExecution();
```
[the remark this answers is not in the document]
hardikar: Removed it

libgpopt/src/exception.cpp
hardikar: it is correct English :P This is an example of https://en.wikipedia.org/wiki/Preposition_stranding

libgpopt/src/operators/CLogicalBitmapTableGet.cpp
chrishajas: I think so. In some cases, the derive function is very simple, but this is being called in:
```
// output columns
CColRefSet *
CDrvdPropRelational::DeriveOutputColumns(CExpressionHandle &exprhdl)
{
	if (!m_is_prop_derived->ExchangeSet(EdptPcrsOutput))
	{
		CLogical *popLogical = CLogical::PopConvert(exprhdl.Pop());
		m_pcrsOutput = popLogical->DeriveOutputColumns(m_mp, exprhdl);
	}

	return m_pcrsOutput;
```

For some operators (CLogicalGbAgg for example), it's actually deriving the output columns.

libgpopt/src/operators/CExpression.cpp
```diff
@@ -398,40 +404,28 @@ CExpression::Pdp
...
-			break;
-		default:
-			GPOS_ASSERT(!"Invalid property type");
-			break;
-	}
+CDrvdPropScalar *
+CExpression::GetDrvdPropScalar()
+	const
+{
+	GPOS_ASSERT(m_pdpscalar->IsComplete());
```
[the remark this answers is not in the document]
chrishajas: I don't see any performance difference when making these retail asserts, so I support making these retail.

libgpopt/include/gpopt/base/CDrvdPropRelational.h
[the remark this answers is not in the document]
hardikar: Ok adding a comment up to tell people where to look.

libgpopt/src/operators/CExpressionHandle.cpp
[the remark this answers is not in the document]
hardikar: Got rid of the exception altogether.

libgpopt/src/base/CDrvdPropRelational.cpp
```diff
@@ -369,7 +369,7 @@ CDrvdPropRelational::OsPrint
 	)
 	const
 {
-	os	<<	"Output Cols: [" << *m_pcrsOutput << "]"
+	os	<<	"Output Cols: [" << *PcrsOutput() << "]"
 		<<	", Outer Refs: [" << *m_pcrsOuter << "]"
```
vraghavan78: why not PcrsOuter() here too?

chrishajas: It's fixed in a later commit.

libgpopt/include/gpopt/operators/CExpressionHandle.h
```diff
@@ -65,9 +65,7 @@ namespace gpopt
 			// attached cost context
 			CCostContext *m_pcc;
 
-			// derived properties of attached expr/gexpr;
-			// set during derived property computation
-			DrvdPropArray *m_pdp;
+			DrvdPropArray *m_pdpplan;
```
vraghavan78: why *m_pdpplan sounds like plan property?

vraghavan78: very confusing. when is it used. comment would help

libgpopt/src/base/CDrvdPropRelational.cpp
```diff
@@ -87,77 +94,52 @@ CDrvdPropRelational::~CDrvdPropRelational()
 void
 CDrvdPropRelational::Derive
 	(
-	CMemoryPool *mp,
+	CMemoryPool *, //mp,
```
vraghavan78: do we need this at all?

libgpopt/src/base/CDrvdPropRelational.cpp
```diff
@@ -87,77 +94,52 @@ CDrvdPropRelational::~CDrvdPropRelational()
...
 	CDrvdPropCtxt * // pdpctxt
 	)
 {
 	GPOS_CHECK_ABORT;
 
-	CLogical *popLogical = CLogical::PopConvert(exprhdl.Pop());
-
 	// call output derivation function on the operator
-	m_pcrsOutput = popLogical->PcrsDeriveOutput(mp, exprhdl);
+	PcrsOutput(exprhdl);
```
vraghavan78: Every one calls PopConvert. Should the api be CLogical pop?

vraghavan78: this is mute.

hardikar:
```
-                       // ctor for expression with derived properties
-                       CExpression
-                               (
-                               CMemoryPool *mp,
-                               DrvdPropArray *pdprop
-                               );
```
Looks like this function is not used, let's get rid of it.

libgpopt/include/gpopt/exception.h
```diff
@@ -37,7 +37,7 @@ namespace gpopt
 		ExmiUnsatisfiedRequiredProperties,
 		ExmiEvalUnsupportedScalarExpr,
 		ExmiCTEProducerConsumerMisAligned,
-		ExmiIncompletePropSpec,
+		ExmiIncompleteDerivedProperties,
```
vraghavan78: incomplete relational?

chrishajas: We got rid of that exception type in a subsequent commit and replaced it with an assert as it wasn't being used in a valid codepath.

libgpopt/src/base/CDrvdPropRelational.cpp
chrishajas: We inherit from `CDrvdProp`, and sibling classes `CDrvdPropPlan` and `CDrvdPropScalar` use `mp` in `Derive()`.

PR #512, opened 2019-07-15
Only create PropConstraint hashmap if necessary

chrishajas:
In many cases, we create the propConstraint hashmap `m_phmcrcrs` with all 127
buckets and then add nothing to it. Instead, only create this hash map if
we'll actually insert something into it.

This reduces memory usage by ~30-40% for some queries.

Another approach is to still create the hash map in all cases, but simply give it a size of 1 when the size of `m_phmcrcrs` is 0 so it takes up less space. This way we don't have to do null checks, but we'll lose out on around ~10% of the memory improvement.


libgpopt/src/base/CPropConstraint.cpp
```diff
@@ -71,8 +71,11 @@ CPropConstraint::InitHashMap
 	)
 {
 	GPOS_ASSERT(NULL == m_phmcrcrs);
-	m_phmcrcrs = GPOS_NEW(mp) ColRefToColRefSetMap(mp);
 	const ULONG ulEquiv = m_pdrgpcrs->Size();
+	if (0 != ulEquiv)
+	{
+		m_phmcrcrs = GPOS_NEW(mp) ColRefToColRefSetMap(mp);
```
hardikar: Add a comment here saying that m_phmcrcrs is only needed when storing equivalent columns.

PR #515, opened 2019-07-23
Support GIN Indexes with ORCA.

bhuvnesh2703:
This commit adds the ORCA side changes required to support GIN Indexes

GIN indexes are not supported with index expression or predicate constraints.
ORCA does not support it currently for other types of indexes too.

Tests are added to the corresponding GPDB commit.

GIN Indexes are only supported using Bitmap Index Scan, so Btree Index Scan
plan are not generated.

Corresponding GPBD PR: https://github.com/greenplum-db/gpdb/pull/8208


libgpopt/src/operators/CPredicateUtils.cpp
```diff
@@ -2031,8 +2031,8 @@ CPredicateUtils::PexprIndexLookup
 
 	if (cmptype == IMDType::EcmptNEq ||
 		cmptype == IMDType::EcmptIDF ||
-		(cmptype == IMDType::EcmptOther && pmdindex->IndexType() != IMDIndex::EmdindGist) || // only GiST indexes with a comparison type other are ok
-		(pmdindex->IndexType() == IMDIndex::EmdindGist && pexprScalar->Arity() < 2)) // we do not support index expressions for GiST indexes
+		(cmptype == IMDType::EcmptOther && !(pmdindex->IndexType() == IMDIndex::EmdindGist || pmdindex->IndexType() == IMDIndex::EmdindGin)) || // only GiST indexes with a comparison type other are ok
```
hardikar:
```suggestion
BOOL gist_or_gin_index = (pmdindex->IndexType() == IMDIndex::EmdindGist || pmdindex->IndexType() == IMDIndex::EmdindGin);
if (...)
		(cmptype == IMDType::EcmptOther && !gist_or_gin_index) || // only GiST & GIN indexes with a comparison type other are ok
		(gist_or_gin_index && pexprScalar->Arity() < 2)) // we do not support index expressions for GiST or GIN indexes
```

hardikar:
Actually, we should have 1 or 2 c-tests here anyway to catch simple bugs, since we run them much more often than GPDB tests when developing.

PR #516, opened 2019-07-24
Add shorthand script to reinstall after switching versions

berlin-ab:
Each time I go through these steps:

git fetch
git co v3.X.X
cmake
ninja

... I can't remember the exact steps recommended by the README.

So, I've added the steps from the readme into a script called
reinstall.sh in the root directory. Now I can do:

git fetch
git co v3.x.x
./reinstall.sh

### DIFF
--- a/reinstall.sh
+++ b/reinstall.sh
@@ -1,0 +1,4 @@
+#!/usr/bin/env bash
+
+cmake -GNinja -H. -Bbuild
+ninja install -C build


hardikar:
Did you consider writing a bash function for this and adding it to you bashrc (or corresponding shell's rc file)?

We actually have scripts for this already. I don't use them, but others in the team do. You should take a look those also.

berlin-ab:
Are the scripts you mention in this repository? I could just use them instead. 

My intention is to make it easy for non-gporca developers to quickly be able to update to a new version, and this script is not intended to be used by people that work on orca often. 

I have considered adding a bash function in my personal tools library, but I thought I would try to help the community of developers that have the same pain as myself (my teammates?) rather than just help myself.

hardikar:
> Are the scripts you mention in this repository? I could just use them instead.

No. And, I'm not sure why. As I said, I don't use them, since I personally believe that they belong to the user, since they will vary a lot based on the who's using them. However the rest of the team does. So maybe we work on making them more "official".

> My intention is to make it easy for non-gporca developers to quickly be able to update to a new version, and this script is not intended to be used by people that work on orca often.

I agree wholeheartedly with your intention, but not with the method. This PR doesn't handle the version change, multi-version management, debug/release builds etc. As it stands it runs two commands, makes assumptions about where the code, install location is, which may or may not be generally true. There is no "default" in those scenarios unfortunately. 

> I have considered adding a bash function in my personal tools library, but I thought I would try to help the community of developers that have the same pain as myself (my teammates?) rather than just help myself.

Since, as you say, it's not intended to be used by people that work on orca often, perhaps it should belong somewhere else where that use is more well-defined.

danielgustafsson:
Couldn't this be added as a custom command in the CMake scripts already used by Orca?

PR #518, opened 2019-07-30
Add 4 xforms for LOJ index apply on dynamic table scan

zellerh:
While we had xforms for inner index joins on partitioned tables and
also xforms for left outer index joins on non-partitioned tables, we
were missing four more xforms that handle left outer index joins on
partitioned tables:

- btree index
- bitmap index
- btree index with an additional select on the inner
- bitmap index with an additional select on the inner


hardikar:
ORCA side changes LGTM. I think we need a GPDB side change for this with correctness tests in ICG. Unless of course, such tests already exist in ICG.

libgpopt/include/gpopt/xforms/CXformLeftOuterJoin2DynamicBitmapIndexGetApply.h
```diff
@@ -0,0 +1,59 @@
+//---------------------------------------------------------------------------
+//	Greenplum Database
+//	Copyright (C) 2017 Pivotal, Inc.
```
bhuvnesh2703: does the year change to 2019?

chrishajas: Yeah, it's been changed.

PR #521, opened 2019-08-06
Remove unused columns from subqueries

ashuka24:
This commit reduces the output columns for existential subqueries.
For existential subqueries, we add a groupby operation on a key of the
outer table, to simulate a semi-join. The grouping columns must be a key
of the outer table, plus any columns that are required by the parent
node.

In an earlier commit, we pruned away unused columns in a query, but
PcrsOutput() could still sometimes contain such unused columns. Our fix
makes sure we only add used ColRefs to the grouping columns, to avoid
problems when we generate and execute a query plan.

Co-Authored-By: Hans Zeller <hzeller@pivotal.io>


libgpopt/src/base/CUtils.cpp
```diff
@@ -2596,9 +2596,21 @@ CUtils::PdrgpcrGroupingKey
...
+		CColRef *pcr = it.Pcr();
+
+		if (CColRef::EUsed == pcr->GetUsage())
+		{
+			pcrsUsedOuter->Include(pcr);
+		}
+	}
 	// filter out system columns since they may introduce columns with undefined sort/hash operators
-	CColRefArray *colref_array = PdrgpcrNonSystemCols(mp, pcrsOutput);
+	CColRefArray *colref_array = PdrgpcrNonSystemCols(mp, pcrsUsedOuter);
```
chrishajas: If you want you can combine the functionality of `PdrgpcrNonSystemCols` into the loop above since it's just filtering out system columns and it's only used here. Up to you though.

libgpopt/src/translate/CTranslatorExprToDXL.cpp
```diff
@@ -7406,6 +7407,10 @@ CTranslatorExprToDXL::MakeDXLTableDescr
 			colref = (*pdrgpcrOutput)[ul];
 			if (colref->GetUsage() != CColRef::EUsed)
 			{
+				if (NULL != requiredProperties && NULL != requiredProperties->PcrsRequired())
+				{
+					GPOS_ASSERT(!requiredProperties->PcrsRequired()->FMember(colref));
```
chrishajas: Can we add a comment on the purpose (and importance) of this assert and what someone should do if this assert is hit.

bhuvnesh2703:
does it mean that PcrsOutput wherever used should always look only the Used Columns? So should be add the exclusion of ununsed columns in the derivation of PcrsOutput?

libgpopt/src/translate/CTranslatorExprToDXL.cpp
```diff
@@ -7377,7 +7377,8 @@ CDXLTableDescr *
 CTranslatorExprToDXL::MakeDXLTableDescr
 	(
 	const CTableDescriptor *ptabdesc,
-	const CColRefArray *pdrgpcrOutput
+	const CColRefArray *pdrgpcrOutput,
+	const CReqdPropPlan *requiredProperties
```
hardikar: Is this a debug-build only requirement then?

hardikar: Also perhaps we should pass only pcrsRequired, rather than all the reqd props here.

hardikar: Also, I know this is minor, but can we please rename requiredProperties -> reqd_prop_plan or keep plan in the name please.

ashuka24:
> does it mean that PcrsOutput wherever used should always look only the Used Columns? So should we add the exclusion of ununsed columns in the derivation of PcrsOutput instead of doing it later..

Hi @bhuvnesh2703, that is a good idea, but it's not so easy to implement. That's because sometimes we call the derivation of PcrsOutput in the preprocessor, before we have complete usage information. In those few cases where we do, PcrsOutput will have additional, unused ColRefs in it. Otherwise (if we derive PcrsOutput after the preprocessor), only used columns will be included.

PR #522, opened 2019-08-08
Consider predicate of type <cast(ident)> array cmp <const array>

bhuvnesh2703:
While processing constraint interval, also consider predicate of type
<cast(ident)> array cmp <const array> else we lose the opportunity
to generate implied quals

```
CREATE TABLE varchar_sc_array_cmp(a varchar);
INSERT INTO varchar_sc_array_cmp VALUES ('a'), ('b'), ('c'), ('d');
EXPLAIN SELECT * FROM varchar_sc_array_cmp t1, varchar_sc_array_cmp t2 where t1.a = t2.a and t1.a in ('b', 'c');
                                                QUERY PLAN
----------------------------------------------------------------------------------------------------------
 Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..862.00 rows=4 width=4)
   ->  Hash Join  (cost=0.00..862.00 rows=2 width=4)
         Hash Cond: orca.varchar_sc_array_cmp.a::text = orca.varchar_sc_array_cmp.a::text
         ->  Table Scan on varchar_sc_array_cmp  (cost=0.00..431.00 rows=1 width=2)
               Filter: (a::text = ANY ('{b,c}'::text[])) AND (a::text = 'b'::text OR a::text = 'c'::text)
         ->  Hash  (cost=431.00..431.00 rows=1 width=2)
               ->  Table Scan on varchar_sc_array_cmp  (cost=0.00..431.00 rows=1 width=2)
                     Filter: a::text = 'b'::text OR a::text = 'c'::text
 Settings:  optimizer_cte_inlining_bound=1000; optimizer_metadata_caching=on
 Optimizer status: PQO version 3.63.1
(10 rows)

SELECT * FROM varchar_sc_array_cmp t1, varchar_sc_array_cmp t2 where t1.a = t2.a and t1.a in ('b', 'c');
 a | a
---+---
 c | c
 b | b
(2 rows)

SET optimizer_array_constraints=on;
EXPLAIN SELECT * FROM varchar_sc_array_cmp t1, varchar_sc_array_cmp t2 where t1.a = t2.a and (t1.a in ('b', 'c') OR t1.a = 'a');
                                        QUERY PLAN
 Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..862.00 rows=4 width=4)
   ->  Hash Join  (cost=0.00..862.00 rows=2 width=4)
         Hash Cond: orca.varchar_sc_array_cmp.a::text = orca.varchar_sc_array_cmp.a::text
         ->  Table Scan on varchar_sc_array_cmp  (cost=0.00..431.00 rows=1 width=2)
               Filter: a = ANY ('{a,b,c}'::character varying[])
         ->  Hash  (cost=431.00..431.00 rows=1 width=2)
               ->  Table Scan on varchar_sc_array_cmp  (cost=0.00..431.00 rows=1 width=2)
                     Filter: a = ANY ('{a,b,c}'::character varying[])
 Settings:  optimizer_cte_inlining_bound=1000; optimizer_metadata_caching=on
 Optimizer status: PQO version 3.63.1
(10 rows)

SELECT * FROM varchar_sc_array_cmp t1, varchar_sc_array_cmp t2 where t1.a = t2.a and (t1.a in ('b', 'c') OR t1.a = 'a');
 a | a
---+---
 c | c
 b | b
 a | a
(3 rows)
```.
TBD:
Was not able to decide where to add the test cases. Because, In gpdb explain plan diffs doesn't capture filters, and in orca with `optimizer_enable_constant_expression_evaluation=off` these can't be done.
Note: But still better to add in gpdb for checking results which are correct anyways, but executes the code path.

ctests passes, icg passes, no changes.

### CODE REVIEW

**Comment by d** [comment]
I believe this patch means well, but...

From a high level, it seems that you are assuming transitivity of different operators. Specifically, in the following query:

```sql
CREATE TABLE foo (b float8);
CREATE TABLE bar (d float8);

SELECT foo.b, bar.d FROM foo JOIN bar ON b = d AND b = '-0';
```

It seems that ORCA _always_ concludes that there should be a relation of d = '-0'. But not so fast, let's look at what we _really_ see in the query:

1. `(b,d)` is in an equivalent class of `float8` values
1. `(b, '-0')` is in an equivalent class of `float8` and type `R` values

The conclusion ORCA draws, is that `(d, '-0')` is also in the 2nd equivalent class mentioned above. But we cannot conclude that, because the two operators used are necessarily transitive (unless they explicitly belong to at least one operator family).

To give a concrete (and somehow contrived) example:

```sql
CREATE SCHEMA evil_op;
SET search_path TO evil_op;

CREATE TYPE floatjz;

CREATE FUNCTION floatjzin(cstring) RETURNS floatjz
LANGUAGE internal IMMUTABLE STRICT AS 'float8in';

CREATE FUNCTION floatjzout(floatjz) RETURNS cstring
LANGUAGE internal IMMUTABLE STRICT AS 'float8out';

CREATE TYPE floatjz (
	INPUT = floatjzin,
	OUTPUT = floatjzout,
	CATEGORY = 'N',
	LIKE = pg_catalog.float8
);

CREATE FUNCTION floatjzeq(floatjz, floatjz) RETURNS bool
LANGUAGE internal IMMUTABLE STRICT AS 'float8eq';
CREATE OPERATOR = (
	PROCEDURE = floatjzeq,
	LEFTARG = floatjz,
	RIGHTARG = floatjz
);
CREATE FUNCTION floatjzlt(floatjz, floatjz) RETURNS bool
LANGUAGE internal IMMUTABLE STRICT AS 'float8lt';
CREATE OPERATOR < (
	PROCEDURE = floatjzlt,
	LEFTARG = floatjz,
	RIGHTARG = floatjz
);

-- spaceship
CREATE FUNCTION btfloatjzcmp(floatjz, floatjz) RETURNS int
LANGUAGE internal IMMUTABLE STRICT AS 'btfloat8cmp';

CREATE OPERATOR FAMILY floatjz_ops USING btree;

CREATE OPERATOR CLASS floatjz_ops
DEFAULT FOR TYPE floatjz USING btree FAMILY floatjz_ops AS
	OPERATOR 1 <,
	OPERATOR 3 =,
	FUNCTION 1 btfloatjzcmp(floatjz, floatjz)
;

CREATE FUNCTION floatjzfloat8mi(floatjz, float8) RETURNS floatjz
LANGUAGE internal IMMUTABLE STRICT AS 'float8mi';

CREATE OPERATOR - (
	PROCEDURE = floatjzfloat8mi,
	LEFTARG = floatjz,
	RIGHTARG = float8
);

CREATE CAST (floatjz AS float8)
	WITHOUT FUNCTION
	;

CREATE FUNCTION floatjzfloat8eq (floatjz, float8) RETURNS bool
SET work_mem FROM CURRENT
LANGUAGE SQL IMMUTABLE STRICT AS $fn$
	SELECT $1::float8 = ($2 - 1);
$fn$;

CREATE OPERATOR = (
	PROCEDURE = floatjzfloat8eq,
	LEFTARG = floatjz,
	RIGHTARG = float8
);

ALTER OPERATOR FAMILY floatjz_ops USING btree ADD
	OPERATOR 3 = (floatjz, float8)
	;

CREATE CAST (float8 AS floatjz)
	WITHOUT FUNCTION
	AS IMPLICIT;

CREATE TABLE foo (b float8);
CREATE TABLE bar (d float8);

INSERT INTO foo(b) VALUES ('-0'),
 ('0'),
 ('3'),
 ('4'),
 ('-1');

INSERT INTO bar(d) SELECT b FROM foo;

SELECT b, d FROM foo, bar WHERE b = d AND b::floatjz = ANY (ARRAY['-0'::float8, '4'::float8]);

SELECT b, d FROM foo, bar WHERE b = d AND b::floatjz = '-0'::float8;
```

We can expect the last two queries to return:


b | d
---|---
-1|-1
3|3

And

b|d
---|---
-1|-1


(The wrong plans, on the other hand, will return zero tuples for both cases)
I also wrote the second query to highlight that this is a pre-existing flaw in ORCA, just being brought further by this patch.

Issue #523, opened 2019-08-16
UNION ALL with many columns may cause explain so slow


wuboy0307:
Greenplum: 5.19.0
PQO version: 3.42.0

Here is a simple example which has 20 SELECT UNION ALL and each table has 500 columns
https://gist.github.com/wuboy0307/3e67f1ebb9f28128e519679d158f9591

When I explain this query. It costs 30 seconds or more to finish.
However, if I turn off optimizer. It costs less than 1 second.

If I explain only one select with 500 columns. It's fast.
https://gist.github.com/wuboy0307/ab6f5b47768ea5bf5b8cf5b26524cb42

Also, if I explain 20 select with less(20) columns, it's also fast.
https://gist.github.com/wuboy0307/7843e4a3c952b7d49b9ce69b1afbfdf6

That issue only happen when there are many UNION ALL and many columns.

The example is quite simple that only have NULL as columns.
In my production databases, the explain process take forever to finish...

chrishajas:
Thanks for the report and clear repro! Looked into this and the bulk of the slowdown here is due to hash collisions in our column hashmap. I've opened a PR (https://github.com/greenplum-db/gporca/pull/524)  that reduces the optimization time from ~25s to ~9s. There's still other improvements to optimization time (and really improving our hashmap/set hashing functions and implementation), but this was low-hanging fruit.

PR #524, opened 2019-08-16
Reduce hash collisions when processing a large number of columns

chrishajas:
Our hashing function, which is dependent on the size of the hash table,
was hashing to the same bucket during constant expression evaluation.
Instead, modify the hash table size to be an odd number to significantly
reduce hash collisions. In the future, we should modify our hashing
function to not be influenced by the hash table size, but that will be a
larger effort.

This reduces optimization time by 50%+ for some queries. In one case,
this reduced optimization time from 25s to 9s on my laptop.

Partially addresses https://github.com/greenplum-db/gporca/issues/523


sambitesh:
LGTM, are there any more even sized Hashmaps in ORCA left?

chrishajas:
> LGTM, are there any more even sized Hashmaps in ORCA left?

I don't believe so, unless we create a custom sized hashmap based on the number of columns. I can also just add logic when initializing the hashmap to always create an odd-sized hashmap. That way we don't accidentally hit this issue again.

hsyuan:
Memo bucket size is 50000.
see https://github.com/greenplum-db/gporca/blob/master/libgpopt/src/search/CMemo.cpp#L36

wuboy0307:
After update PQO to 3.72.0. The issue still exists. Are you going to enhance this more on next version or is there any workaround to overcome this issue? 
I have a table which contains 700up columns. The performance is really bad. 

Thanks a lot...

zellerh:
Hi @wuboy0307, would you be able to provide a test case to reproduce your issue?

wuboy0307:
> Hi @wuboy0307, would you be able to provide a test case to reproduce your issue?

Hi @zellerh ,
Please take a look
https://gist.github.com/wuboy0307/1f314bd4293be5653378652f21bd5472

I create a table with 100000 rows, and the table only have 1 column. All other columns will be fill with "NULL AS" to 700 columns.

As you can see in my comment, 
1 column and 100000 rows: It takes 0.135 seconds 
700 column and 0 rows: It takes 0.084 seconds
700 column and 1000 rows: It takes 0.284 seconds
700 column and 10000 rows: It takes 1.705 seconds
700 column and 50000 rows: It takes 8.398 seconds
700 column and 100000 rows: It takes 16.16 seconds

It won't happen if it's 1 column with many rows or 0 row with many columns.
With many columns and many result rows, it's getting slower and slower (which is linear increasing trend).

zellerh:
When I tried this on GPDB 6.0, I got a bit faster results

|rows|cols|wuboy0307's time|zellerh's time|time with limit clause|explain time on MacOS|
|---|---|---|---|---|---|
|100000|1|0.135|x|y|z|
|0|700|0.084|x|y|z|
|0|700|0.084|x|y|z|
|0|700|0.084|x|y|z|
|0|700|0.084|x|y|z|
|0|700|0.084|x|y|z|
|0|700|0.084|x|y|z|
|0|700|0.084|x|y|z|
|0|700|0.084|x|y|z|

PR #528, opened 2019-08-23
Update README instructions

ashuka24:


README.md
```diff
@@ -100,7 +105,7 @@ DEBUG-mode tests tend to be more rigorous.
 ## Adding tests
 
 Most of the regression tests come in the form of a "minidump" file.
-A minidump is an XML file that contains all the input needed to plan a query,
```
chrishajas: we should keep this as "xml". mdp is a made up word.

README.md
```diff
@@ -147,19 +153,54 @@ xmllint --format ~/data-master/minidumps/Minidump_20160610_220222_4_14.mdp > dat
                 // TODO:  - Jul 14 2015; disabling it for debug build to reduce testing time
 ```
 
+Alternatively, it could also be added to the proper test suite as follows:
```
chrishajas: test suite in server/CMakeLists as follows...

README.md
```diff
@@ -147,19 +153,54 @@ xmllint --format ~/data-master/minidumps/Minidump_20160610_220222_4_14.mdp > dat
...
+ PartTbl-LeftOuterHashJoin-DPE-IsNull PartTbl-LeftOuterNLJoin-DPE-IsNull
+-PartTbl-List-DPE-Varchar-Predicates PartTbl-List-DPE-Int-Predicates;
++PartTbl-List-DPE-Varchar-Predicates PartTbl-List-DPE-Int-Predicates
++Mytest;
+```
+
+<a name="updatetest"></a>
+## Update tests
+
+In some situations, a failing test does not necessarily imply that the fix is
```
chrishajas: Add "Note: This will overwrite existing mdp files. This is best used after committing existing changes, so you can more easily see the diff. Alternatively, you can use fix_mdps --dryRun to not change mdp files".

README.md
```diff
@@ -328,12 +369,14 @@ We accept contributions via [Github Pull requests](https://help.github.com/artic
...
     * Follow the previous sections on this page to setup and build in your environment.
-1. Follow the naming and formatting style guide described [here](StyleGuide.md).
-1. Run through all the [tests](#test) in your feature branch and ensure they are successful.
+3. Follow the naming and formatting style guide described [here](StyleGuide.md).
+4. Run through all the [tests](#test) in your feature branch and ensure they are successful.
     * Follow the [Add tests](#addtest) section to add new tests.
-1. Push your local branch to your fork (e.g. `git push origin better_orca`) and [submit a pull request](https://help.github.com/articles/creating-a-pull-request)
+    * Follow the [Update tests](#updatetest) section to update existing tests.
+    * Make sure that ctest passes in both debug and retail build since there are some tests that do not overlap.
+5. Push your local branch to your fork (e.g. `git push origin better_orca`) and [submit a pull request](https://help.github.com/articles/creating-a-pull-request)
```
chrishajas: In some cases, ICG tests in gpdb may also need to be modified or additional ICG tests may need to be added to provide full coverage for the fix (for example, if the patch fixes wrong results or an execution specific issue).

Issue #505, opened 2019-06-27
ORCA does not generate a direct dispatch plan for aggregate filtered on the distribution key


khannaekta:
ORCA Version: 3.48.0
**Repro**
```
CREATE TABLE foo(a int) DISTRIBUTED BY(a);
EXPLAIN SELECT count(*) FROM foo WHERE a = 1;
```
**ORCA generated plan**
```
EXPLAIN SELECT count(*) FROM foo WHERE a = 1;
                                     QUERY PLAN
------------------------------------------------------------------------------------
 Aggregate  (cost=0.00..431.00 rows=1 width=8)
   ->  Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..431.00 rows=1 width=8)
         ->  Aggregate  (cost=0.00..431.00 rows=1 width=8)
               ->  Seq Scan on foo  (cost=0.00..431.00 rows=1 width=1)
                     Filter: (a = 1)
 Optimizer: Pivotal Optimizer (GPORCA) version 3.48.0
(6 rows)
```
**Planner generated plan**
```
EXPLAIN SELECT count(*) FROM foo WHERE a = 1;
                                       QUERY PLAN
----------------------------------------------------------------------------------------
 Aggregate  (cost=1304.05..1304.06 rows=1 width=8)
   ->  Gather Motion 1:1  (slice1; segments: 1)  (cost=1303.99..1304.04 rows=1 width=8)
         ->  Aggregate  (cost=1303.99..1304.00 rows=1 width=8)
               ->  Seq Scan on foo  (cost=0.00..1303.75 rows=33 width=0)
                     Filter: (a = 1)
 Optimizer: Postgres query optimizer
(6 rows)
```

sambitesh:
Fixed by #575

PR #531, opened 2019-09-04
Fix Direct Dispatch with Agg havings no grouping columns

bhuvnesh2703:
Currently, if there is no grouping column on the Agg, we return a NULL
constraint from CLogicalGbAgg::PpcDeriveConstraint.

However, if there is no grouping column on the agg, we can simply pass
the constraint from the child. This allows us to attach direct dispatch
information if possible using the constraint available on the child.

For example:

```create table foo (a int);
explain analyze select count(*) from foo where a = 1;;
INFO:  (slice 1) Dispatch command to SINGLE content
                                                          QUERY PLAN
------------------------------------------------------------------------------------------------------------------------------
 Aggregate  (cost=0.00..431.00 rows=1 width=8) (actual time=1.205..1.206 rows=1 loops=1)
   ->  Gather Motion 1:1  (slice1; segments: 1)  (cost=0.00..431.00 rows=1 width=1) (actual time=1.192..1.192 rows=0 loops=1)
         ->  Seq Scan on foo  (cost=0.00..431.00 rows=1 width=1) (never executed)
               Filter: (a = 1)
 Planning time: 22.794 ms
   (slice0)    Executor memory: 59K bytes.
   (slice1)    Executor memory: 42K bytes (seg1).
 Memory used:  128000kB
 Optimizer: Pivotal Optimizer (GPORCA) version 3.67.0
 Execution time: 2.506 ms
```

In `CTranslatorExprToDXLUtils::SetDirectDispatchInfo` if there is no contraint, we don't
attach the direct dispatch info.
```
	CPropConstraint *ppc = pdpRel->Ppc();
	if (NULL == ppc->Pcnstr())
	{
		return;
	}
```
For a node, the bare minimum constraint can be the constraint of the child, if
the node itself has no constraint.

Before the fix, we will dispatch the query to all the nodes.
Corresponding GPDB PR: https://github.com/greenplum-db/gpdb/pull/8555/files

### DIFF
--- a/libgpopt/src/operators/CLogicalGbAgg.cpp
+++ b/libgpopt/src/operators/CLogicalGbAgg.cpp
@@ -377,6 +377,17 @@ CLogicalGbAgg::PpcDeriveConstraint
 	)
 	const
 {
+	// if there is no grouping column, pass the constraint of the child
+	if (m_pdrgpcr->Size() == 0)
+	{
+		CPropConstraint *ppc = exprhdl.GetRelationalProperties(0 /*first child*/)->Ppc();
+		if (NULL != ppc)
+		{
+			ppc->AddRef();
+		}
+		return ppc;
+	}
+
 	CColRefSet *pcrsGrouping = GPOS_NEW(mp) CColRefSet(mp);
 	pcrsGrouping->Include(m_pdrgpcr);
 
--- a/data/dxl/minidump/DirectDispatchOnAggWithoutAnyGroupingCol.mdp
+++ b/data/dxl/minidump/DirectDispatchOnAggWithoutAnyGroupingCol.mdp
@@ -1,0 +1,269 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<dxl:DXLMessage xmlns:dxl="http://greenplum.com/dxl/2010/12/">
+  <dxl:Comment><![CDATA[
+    Objective: Since the table foo is distributed by column a and the filter is on column a for count(*), it should
+    be dispatched to SINGLE segment.
+    CREATE TABLE foo (a int);
+    EXPLAIN SELECT count(*) FROM foo WHERE a = 1;
+  ]]>
+  </dxl:Comment>
+  <dxl:Thread Id="0">
+    <dxl:OptimizerConfig>
+      <dxl:EnumeratorConfig Id="0" PlanSamples="0" CostThreshold="0"/>
+      <dxl:StatisticsConfig DampingFactorFilter="0.750000" DampingFactorJoin="0.010000" DampingFactorGroupBy="0.750000"/>
+      <dxl:CTEConfig CTEInliningCutoff="0"/>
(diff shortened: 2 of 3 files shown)


libgpopt/src/operators/CLogicalGbAgg.cpp
```diff
@@ -377,6 +377,17 @@ CLogicalGbAgg::PpcDeriveConstraint
...
 {
+	// if there is no grouping column, pass the constraint of the child
+	if (m_pdrgpcr->Size() == 0)
+	{
+		CPropConstraint *ppc = exprhdl.GetRelationalProperties(0 /*first child*/)->Ppc();
+		if (NULL != ppc)
+		{
+			ppc->AddRef();
+		}
+		return ppc;
```
vraghavan78: hold on, how can you have constraint on the column that is not available above it. This does not smell right.

bhuvnesh2703: i may be way off :) but here is my thought..
constraint are part of the relational property, do they plan any role in the physical world (no i believe)? the way direct dispatch is handled is 
1. find out what are the base tables (start traversing from top most node to child node)
2. what is the hash distribution of these tables
3. if the hash distribution matches the constraint columns then we attach the direct dispatch info in the dxl.. (direct dispatch is not decided in optimization stage)
we derive constraints as part of the relational properties, and in the physical stage we use the the additionally generated predicates etc and convert them to physical but don't use the constraint, nor in the optimization we use constraints for any properties..

for the top node the relational property will list all the columns, but the derived plan property will only have the required ones...
may be i am missing something?

bhuvnesh2703:
the github https://github.com/greenplum-db/gporca/issues/505 issue is a simple use case, this pr does not fix the below case as thats a separate issue.. but i thought about it earlier if i really need the column a at agg level, so just sharing.
let's consider the use case
```sql
explain select b from foo where a = 1 group by b;
                                              QUERY PLAN
 Gather Motion 3:1  (slice2; segments: 3)  (cost=0.00..431.00 rows=1 width=4)
   ->  GroupAggregate  (cost=0.00..431.00 rows=1 width=4)
         Group Key: b
         ->  Sort  (cost=0.00..431.00 rows=1 width=4)
               Sort Key: b
                     Hash Key: b
                     ->  Seq Scan on foo  (cost=0.00..431.00 rows=1 width=4)
                           Filter: (a = 1)
 Optimizer: Pivotal Optimizer (GPORCA) version 3.67.0
```
Here -> by the time we are in the expr to dxl stage, we have already added non gather motions (i.e redistribute).. if we have to make this plan direct dispatch able (which it can be), we will have to not add redistribute motion in the optimization phase based on the constraint and surety that the node below is direct dispatchable, and if that is all good  request AnyDistribution from global agg child based on the 100% gaurantee that its direct dispatch able, but even in that case we will not need the column a at agg level in physical world
i am open for suggestion or changes...

bhuvnesh2703:
Closing this for now, trying to see a better way for it...

PR #532, opened 2019-09-05
Fix stats calculation for predicates with UUID columns

asubramanya:
For predicates that involved columns of UUID types, ORCA would consider the
predicate as unsupported and would estimate the selectivity to be 40%. This is
because ORCA did not have statistics for UUID columns. This would cause a bad
plan to be selected in some cases.  This patch fixes the issue by adding
support for the UUID type in CMDIdGPDB class and make it a LINT value for
comparison purposes.


libnaucrates/src/statistics/CStatsPredUtils.cpp
```diff
@@ -304,7 +312,7 @@ CStatsPredUtils::GetPredStats
 	IDatum *datum = scalar_const_op->GetDatum();
 
 	BOOL is_text_related_type = IsTextRelatedType(datum->MDId()) && IsTextRelatedType(col_ref->RetrieveType()->MDId());
-	if (is_text_related_type && !CHistogram::SupportsTextFilter(stats_cmp_type))
+	if ((is_text_related_type || IsUUIDType(datum->MDId())) && !CHistogram::SupportsTextFilter(stats_cmp_type))
```
vraghavan78: `CHistogram::SupportsTextFilter` is a horrible function name. What we wanted it to be called is `SupportedStatsOpOnTextColumns` or something there off.

libnaucrates/src/statistics/CStatsPredUtils.cpp
```diff
@@ -304,7 +312,7 @@ CStatsPredUtils::GetPredStats
 	IDatum *datum = scalar_const_op->GetDatum();
 
 	BOOL is_text_related_type = IsTextRelatedType(datum->MDId()) && IsTextRelatedType(col_ref->RetrieveType()->MDId());
```
vraghavan78: why not have uuid be part of `IsTestRelatedType`?

libnaucrates/src/statistics/CStatsPredUtils.cpp
```diff
@@ -242,6 +242,14 @@ CStatsPredUtils::IsTextRelatedType(const IMDId *mdid)
 	|| mdid->Equals(&CMDIdGPDB::m_mdid_text);
```
chrishajas: The `IsTextRelatedType` is also used in `IsHistogramForTextRelatedTypes`. I think we also would want to include UUID types there?

Ideally, if it's correct, it would be nice to just have to add the UUID comparison here so we don't miss it in future refactors.

libnaucrates/src/base/IDatumStatisticsMappable.cpp
```diff
@@ -11,6 +11,7 @@
 
 #include "naucrates/base/IDatumStatisticsMappable.h"
 #include "naucrates/md/CMDTypeGenericGPDB.h"
+#include "naucrates/statistics/CStatsPredUtils.h"
```
ashuka24: why was this added?

libnaucrates/src/md/CMDTypeGenericGPDB.cpp
```diff
@@ -24,6 +24,8 @@
 #include "naucrates/dxl/xml/CXMLSerializer.h"
 #include "naucrates/dxl/CDXLUtils.h"
 
+#include "naucrates/dxl/gpdb_types.h"
```
ashuka24: it seems like this wasn't needed before, why is this needed now?

libnaucrates/src/base/IDatumStatisticsMappable.cpp
asubramanya: Removed.

libnaucrates/src/md/CMDTypeGenericGPDB.cpp
asubramanya: Removed.

libnaucrates/src/statistics/CStatsPredUtils.cpp
asubramanya: it seems like we store UUID as strings of hex digits separated by hyphens. So it should be OK to consider them as text related types.

server/src/unittest/CTestUtils.cpp
```diff
@@ -4284,7 +4284,7 @@ CTestUtils::CreateGenericDatum
 	{
 		dxl_datum = GPOS_NEW(mp) CDXLDatumStatsDoubleMappable(mp, mdid_type, default_type_modifier, pmdtype->IsPassedByValue() /*is_const_by_val*/, false /*is_const_null*/, data, ulbaSize, CDouble(value));
 	}
-	else if (mdid_type->Equals(&CMDIdGPDB::m_mdid_bpchar) || mdid_type->Equals(&CMDIdGPDB::m_mdid_text))
+	else if (mdid_type->Equals(&CMDIdGPDB::m_mdid_bpchar) || mdid_type->Equals(&CMDIdGPDB::m_mdid_text) || mdid_type->Equals(&CMDIdGPDB::m_mdid_uuid))
```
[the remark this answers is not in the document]
asubramanya: Done.

vraghavan78:
`CHistogram::SupportsTextFilter` can we refactor that.

PR #533, opened 2019-09-13
Only print Unexpected Errors

sambitesh:
Earlier ORCA would log all erorrs. Even when they are due to unsupported
features. During gpinit when we are running catalog queries which ORCA doesn't
support by default, we ended up rpinting unecessary error messages to the log.

Now we want to log only unexpected errors. This will cut lot of clutter from log files.


libgpos/src/_api.cpp
```diff
@@ -22,10 +22,88 @@
 #include "gpos/task/CAutoTaskProxy.h"
 #include "gpos/task/CWorkerPoolManager.h"
 
+
```
ashuka24: remove extra line

libgpos/src/_api.cpp
```diff
@@ -22,10 +22,88 @@
...
+#include "gpopt/exception.h"
+
 using namespace gpos;
 
+const ULONG expected_opt_fallback[] =
+{
+	gpopt::ExmiInvalidPlanAlternative,		// chosen plan id is outside range of possible plans
+	gpopt::ExmiUnsupportedOp,				// unsupported operator
+	gpopt::ExmiUnsupportedPred,				// unsupported predicate
+	gpopt::ExmiUnsupportedCompositePartKey	// composite partitioning keys
```
chrishajas: Should `ExmiUnsupportedNonDeterministicUpdate` also be added here?

libgpos/CMakeLists.txt
```diff
@@ -17,6 +17,9 @@ endif()
 
 include_directories(include)
 
+include_directories(${PROJECT_SOURCE_DIR}/libgpopt/include/)
+include_directories(${PROJECT_SOURCE_DIR}/libnaucrates/include/)
```
chrishajas: Why is this needed now?

libgpos/src/_api.cpp
```diff
@@ -22,10 +22,88 @@
...
+
+// array of DXL minor exception types that error out and NOT fallback to the planner
+const ULONG expected_dxl_errors[] =
+{
+	gpdxl::ExmiDXL2PlStmtExternalScanError,	// external table error
+	gpdxl::ExmiQuery2DXLNotNullViolation,	// not null violation
+};
+
+BOOL
+ShouldErrorOut
```
zellerh: Is this method used anywhere?

sambitesh: It's used in GPDB side.

https://github.com/greenplum-db/gpdb/blob/beac8ab91e44903e67f9ab1207f20ed2716902f7/src/backend/gpopt/utils/COptTasks.cpp#L794

libgpos/CMakeLists.txt
sambitesh: These two lines are needed to compile correctly since we have included `gpopt/exception.h` and `naucrates/exception.h` in the `_api.cpp` which is inside `gpos` directory.

libgpos/src/_api.cpp
sambitesh: Done.

PR #535, opened 2019-09-17
Introduce infrastructure to support PallocMemoryPool and PallocMemoryPoolManager

chrishajas:
This PR includes the ORCA-side changes to support the PallocMemoryPool. The majority of the commits are refactors/removals of unused or unnecessary code. The primary functionality of this PR is in the commit: `Allow overriding the singleton MemoryPoolManager`. 

See the corresponding GPDB PR for details on the new CPallocMemoryPool and CPallocMemoryPoolManager classes.

Note that when running ORCA tests, we will still use the old CMemoryPoolTracker, and thus there is no performance change for tests or when gpdb allocators are not used.

Thanks to @hlinnaka for his proposal and suggestions for this method. See https://github.com/greenplum-db/gporca/pull/529 for the previous implementation/discussion.

Co-authored-by: Shreedhar Hardikar <shardikar@pivotal.io>


libgpos/src/task/CWorkerPoolManager.cpp
```diff
@@ -78,8 +78,8 @@ CWorkerPoolManager::Init()
 		CMemoryPoolManager::GetMemoryPoolMgr()->Create
 			(
 			CMemoryPoolManager::EatTracker,
-			true /*fThreadSafe*/,
-			GPOS_WORKERPOOL_MEM_POOL_SIZE
+			true /*fThreadSafe*/
+			// FIXME: GPOS_WORKERPOOL_MEM_POOL_SIZE
```
vraghavan78: why is the there?

libgpos/server/src/unittest/gpos/memory/CMemoryPoolBasicTest.cpp
```diff
@@ -307,8 +295,7 @@ CMemoryPoolBasicTest::EresLeak
 	{
 		CAutoMemoryPool amp
 			(
-			CAutoMemoryPool::ElcStrict,
```
vraghavan78: what does Strict do?

libgpos/include/gpos/memory/CMemoryPoolInjectFault.h
```diff
@@ -65,7 +67,7 @@ namespace gpos
 			virtual
 			ULLONG TotalAllocatedSize() const
 			{
-				return GetUnderlyingMemoryPool()->TotalAllocatedSize();
+				return m_total_allocated_size;
```
ashuka24: as this is not set anywhere in GPORCA, but is overwritten by the GPDB side function if `gpdb_allocators` is on, does this mean that it would return gibberish when `gpdb_allocators` is off?

libgpos/src/memory/CMemoryPoolTracker.cpp
```diff
@@ -228,13 +151,7 @@ CMemoryPoolTracker::Free
...
 
 	// pass request to underlying memory pool;
-	GetUnderlyingMemoryPool()->Free(header);
-
-	// update committed memory value
-	if (m_capacity != gpos::ullong_max)
-	{
-		m_reserved -= total_size;
-	}
+	clib::Free(header);
```
zellerh: This comment really applies to the existing code in CMemoryPool::NewImpl().

CMemoryPool::NewImpl adds an AllocHeader in front of the allocated memory and a 1 byte allocation type at the end. That adds three fields to each allocation:
- Pointer to the memory pool. As far as I can tell, we don't need that pointer. This method calls clib::Free and CMemoryPoolPaloc calls pfree. So, we could just call a global method that either calls static CMemoryPoolTracker::Free or static CMemoryPoolPalloc::Free, based on the guc setting.
- Allocation size: That size might be needed to compute the number of elements if we do an array allocation, but we don't need it when we allocate a single element.
- Allocation type (single or array). As far as I can tell, we don't need this in a retail build, all we do is assert that the allocation type matches when we deallocate?

I would hope this could reduce the amount of memory allocated by ORCA. In an earlier prototype https://github.com/greenplum-db/gporca/pull/529 we saw very significant memory savings by removing this and other similar overheads.

zellerh:
There are some places where we generate a CAutoMemoryPool just because we need a memory pool for a function call. It would be better to provide a static method to get a default memory pool instead of creating and destroying a new one for a single call. Examples:

- CDefaultComparator::Equals()
- 4 more CDefaultComparator methods
- CUtils::FUsesChildColsOnly()
- CUtils::GenerateFileName()
- CUtils::FHasDuplicates()

Maybe we could do some of the optimizations suggested in my comments as a follow-up PR?

libgpos/src/task/CWorkerPoolManager.cpp
chrishajas: This was removed in a subsequent commit.

libgpos/server/src/unittest/gpos/memory/CMemoryPoolBasicTest.cpp
chrishajas: It forces the memory pool to always check for leaks.

libgpos/include/gpos/memory/CMemoryPool.h
```diff
@@ -42,6 +42,7 @@ GPOS_CPL_ASSERT(GPOS_MEM_ALIGNED_STRUCT_SIZE(gpos::ULONG) == GPOS_MEM_ARCH);
 
 // static pattern to init memory
 #define GPOS_MEM_INIT_PATTERN_CHAR	(0xCC)
+#define GPOS_MEM_FREED_PATTERN_CHAR	(0xCD)
```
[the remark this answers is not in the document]
chrishajas: Done.

libgpos/src/memory/CMemoryPoolTracker.cpp
chrishajas: Thanks for this analysis! We definitely agree, this will reduce the memory overhead and optimization time quite a bit. Since this will require further refactors, we're going to do it as a separate PR. A POC showed that we'll get similar savings to the other PR once we remove the overhead.

chrishajas:
> There are some places where we generate a CAutoMemoryPool just because we need a memory pool for a function call. It would be better to provide a static method to get a default memory pool instead of creating and destroying a new one for a single call. Examples:
> 
> * CDefaultComparator::Equals()
> * 4 more CDefaultComparator methods
> * CUtils::FUsesChildColsOnly()
> * CUtils::GenerateFileName()
> * CUtils::FHasDuplicates()
> 
> Maybe we could do some of the optimizations suggested in my comments as a follow-up PR?

Yes, I've seen that for some cases we create many, many short-living memory pools (which will therefore create/destroy many, many memory contexts). I've created a story to do this (should be fairly quick).

libgpos/include/gpos/memory/CMemoryPoolInjectFault.h
chrishajas: This is for the CMemoryPoolInjectFault. This class also isn't being used anywhere, so I'm going to remove it. It previously was used only for testing in debug builds. With removing the concept of stacked memory pools, this isn't being used however.

libgpos/src/_api.cpp
```diff
@@ -114,17 +114,9 @@ gpos::BOOL IsUnexpectedFailure
...
-
-	if (NULL == func_ptr_alloc || NULL == func_ptr_free) {
-	  func_ptr_alloc = clib::Malloc;
-	  func_ptr_free = clib::Free;
-	}
-
 	CWorker::abort_requested_by_system = params->abort_requested;
 
-	if (GPOS_OK != gpos::CMemoryPoolManager::Init(func_ptr_alloc, func_ptr_free))
+	if (GPOS_OK != gpos::CMemoryPoolManager::Init((CMemoryPoolManager *) params->memorypool_manager))
```
d: Avoid a C-style cast here. Prefer to use named casts. Namely, this seems to be a `reinterpret_cast<>`. Or better, why don't we define `memorypool_manager` with proper type instead of `void*`?

d: I misspoke earlier: I really meant to suggest replacing the C-style cast with `static_cast<>`. How do you like the idea of declaring the struct member in `gpos_init_params` with an actual type, instead of `void*`?

chrishajas: I tried that and got into a bit of header circular dependencies. I'll play around with it a bit more though.

chrishajas: Ok, I was able to figure out the header situation. Got rid of the `void*`.

libgpos/src/memory/CMemoryPoolTracker.cpp
```diff
@@ -228,13 +151,7 @@ CMemoryPoolTracker::Free
 	m_allocations_list.Remove(header);
 
 	// pass request to underlying memory pool;
```
ashuka24: There no longer is an underlying memory pool. Can we update this comment.

libgpos/src/memory/CMemoryPoolTracker.cpp
```diff
@@ -96,24 +88,14 @@ CMemoryPoolTracker::Allocate
 	GPOS_ASSERT(GPOS_MEM_ALLOC_MAX >= bytes);
 
 	ULONG alloc = GPOS_MEM_BYTES_TOTAL(bytes);
-	const BOOL mem_available = Reserve(alloc);
 
 	// allocate from underlying
```
ashuka24: We no longer have an underlying memory pool. Can we update this comment?

zellerh:
Took another look at the recent commits and really like how this came out!

PR #537, opened 2019-09-30
Improve index selection for bitmap indexes

ashuka24:
This PR contains the following components
- Prevent btree indexes on AO tables from following the btree path
- Allow col op col predicates as a supported index predicate
- If the predicate is not supported, break it down and combine the
supported conjuncts instead of keeping them separate
- Use predicate selectivity to determine the best index

Btree indexes on AO tables are not yet supported. Now that such indexes
are sent over to ORCA as type btree instead of type bitmap, we need to
prevent this path from being followed. Instead, btree indexes on AO
tables can be handled by creating a Bitmap index with an underlying
btree index.

In the bitmap index path, alternatives that should not have been
generated (such as picking an index when the first column of the index
was not in the predicate) were being added to the memo because col op
col predicates were being handled separately from col op const
predicates.

In reality, these two scenarios should be following the same code path
when generating viable indexes.

Previously, if the predicate was not all conjuncts, it would be
immediatly broken down into separate predicates. For example, say you
had the predicate a = 4 AND b = 3 AND c = 3 OR (d = 2 AND e = 1), this
would get broken down and ORCA would find indexes for a  = 4, b = 3, c =
3 as individual components.Instead, we want to combine these 3 conjuncts
and find the best matching index.

Additionally, use more than just the number of residuals as an indicator
for a good index. Use the selectivity of the index predicate as an
indiator. If we do not have the NDV of the col, the selectivity
calculation of sel(col = outer ref) = default selectivity of 0.4.

Pick best index based on 1) selectivity, 2) number of residual
predicates and 3) number of columns in the index.

For example, if the predicate was a = 4 and c = 3 and the two indexes
were idx_abc and idx_ac. Say the selectivity was the same, as the number
of residuals is both 0, we want to select the index with the least
number of columns, idx_ac.

Co-Authored-By: Ashuka Xue <axue@pivotal.io>
Co-Authored-By: Hans Zeller <hzeller@pivotal.io>


libgpopt/src/xforms/CXformUtils.cpp
```diff
@@ -3672,9 +3607,6 @@ CXformUtils::CreateBitmapIndexProbes
 	CColRefArray *pdrgpcrOutput,
 	CColRefSet *pcrsOuterRefs,
 	CColRefSet *pcrsReqd,
-	BOOL fConjunction,
-	CExpressionArray *pdrgpexprBitmap,
-	CExpressionArray *pdrgpexprRecheck,
 	CExpressionArray *pdrgpexprBitmapResult,
```
zellerh: Given that the caller always merges/joins these arrays into a single expression for every call, it might be easier to allocate these arrays inside CreateBitmapIndexProbesWithOrWithoutPredBreakdown and to return two CExpression pointers instead of two arrays.

libgpopt/src/xforms/CXformUtils.cpp
```diff
@@ -3746,6 +3679,16 @@ CXformUtils::CreateBitmapIndexProbes
 							&pexprRecheck,
 							&pexprResidual
 							);
+			// if no index path was constructed for this predicate, return it as residual.
+			// Example, with schema: t(a, b, c, d), index i1(a,b)
+			// and predicate: (a = 3) OR (c = 4 AND d =5)
+			// no index path will be found for (c = 4 AND d =5), in which case the entire
+			// disjunct will become a residual.
+			if (NULL == pexprBitmap)
```
zellerh: Could add a comment that in case of OR predicates, the residuals won't contain meaningful data, since ANDs and ORs will be jumbled together (we don't have an fConjunction parameter for this method). There is logic in the caller, method ComputeBitmapTableScanResidualPredicate() called from PexprScalarBitmapBoolOp() that will ignore such jumbled predicates and replace them with the original predicate.

server/CMakeLists.txt
```diff
@@ -91,10 +91,14 @@ CTypeModifierTest:
 TypeModifierColumn TypeModifierCast TypeModifierConst TypeModifierDoubleMappableConst TypeModifierArrayRef;
 
 CIndexScanTest:
-IndexedNLJBitmap BTreeIndex-Against-InList BTreeIndex-Against-ScalarSubquery BitmapIndex-Against-InList
+IndexedNLJBitmap BTreeIndex-Against-InList BTreeIndex-Against-ScalarSubquery
```
vraghavan78: There is still one Bitmap test

libgpopt/src/operators/CPredicateUtils.cpp
```diff
@@ -165,6 +165,59 @@ CPredicateUtils::FComparison
...
+		return false;
+	}
+
+	CExpression *pexprLeft = (*pexpr)[0];
+	CExpression *pexprRight = (*pexpr)[1];
+
+	BOOL leftIsACol = (CUtils::FScalarIdent(pexprLeft) ||
+					   CScalarIdent::FCastedScId(pexprLeft));
+	BOOL rightIsACol = (CUtils::FScalarIdent(pexprRight) ||
+						CScalarIdent::FCastedScId(pexprRight));
```
sambitesh: Minor comment, it will be good to have IsACol function which calls `FCastedScId` and `FScalarIdent` internally.

libgpopt/src/xforms/CXformUtils.cpp
zellerh: Done.

libgpopt/src/xforms/CXformIndexGet2IndexScan.cpp
```diff
@@ -69,42 +69,45 @@ CXformIndexGet2IndexScan::Transform
...
 	CIndexDescriptor *pindexdesc = pop->Pindexdesc();
-	pindexdesc->AddRef();
 
 	CTableDescriptor *ptabdesc = pop->Ptabdesc();
-	ptabdesc->AddRef();
 
-	CColRefArray *pdrgpcrOutput = pop->PdrgpcrOutput();
-	GPOS_ASSERT(NULL != pdrgpcrOutput);
-	pdrgpcrOutput->AddRef();
+	if (!(pindexdesc->IndexType() == IMDIndex::EmdindBtree && ptabdesc->IsAOTable())) {
```
chrishajas: Does this check belong here or in the promise/pattern match?

libnaucrates/src/statistics/CFilterStatsProcessor.cpp
```diff
@@ -69,6 +69,120 @@ CFilterStatsProcessor::MakeStatsFilterForScalarExpr
 	return result_stats;
 }
 
+// compute the selectivity of a predicate, applied to a base table,
+// making some simple default assumptions about outer refs
+CDouble
+CFilterStatsProcessor::SelectivityOfPredicate
```
chrishajas: Should there be a lower bound if there's a bunch of unsupported conjuncts for example (or a bunch of non-equality filters)

libgpopt/src/xforms/CXformUtils.cpp
```diff
@@ -3432,6 +3476,21 @@ CXformUtils::PexprBitmap
...
 
+			pdrgpexprIndex->AddRef();
+			CExpression *pexprIndex = CPredicateUtils::PexprConjunction(mp, pdrgpexprIndex);
+
+			CDouble selectivity = CFilterStatsProcessor::SelectivityOfPredicate(mp, pexprIndex, ptabdesc, pcrsOuterRefs);
+
+			pexprIndex->Release();
+
+			// Btree indexes on AO tables are only great when the NDV is high. Do this check here
+			if (selectivity > 0.05 && pmdrel->IsAOTable() && pmdindex->IndexType() == IMDIndex::EmdindBtree)
```
hardikar: Can we #define this number "0.05"? AO_TABLE_BTREE_INDEX_SELECTIVITY_THRESHOLD e.g

libgpopt/src/operators/CPredicateUtils.cpp
```diff
@@ -165,6 +165,59 @@ CPredicateUtils::FComparison
...
+    CColRefSet *pcrsUsedLeft = CDrvdPropScalar::GetDrvdScalarProps(pexprLeft->PdpDerive())->PcrsUsed();
```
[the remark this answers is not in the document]
chrishajas: 🔜 The scalar PR hasn't been merged yet

libnaucrates/src/statistics/CFilterStatsProcessor.cpp
zellerh: Yes, good point. We should apply the same damping factor of 0.75 that we use in calculating the regular stats for the filter.

hardikar:
Looks much better! Thanks. I hope we can do something similar for btrees as well.

PR #539, opened 2019-10-02
Simplify CMemoryPool to reduce unnecessary headers and logic

chrishajas:
Move headers and logic that are only needed for CMemoryPoolTracker out
of CMemoryPool and into CMemoryPoolTracker, and collapse the headers
into into one there. This reduces memory utilization for external memory
pools (e.g.: CMemoryPoolPalloc), as they don't need the additional
headers.

Specifically, the AllocHeader and SAllocHeader have been merged
into CMemoryPoolTracker, and the recording methods have been moved into
Tracker. UserSizeOfAlloc, DeleteImpl, and NewImpl are overridden in
CMemoryPoolPalloc and CMemoryPoolTracker.

Also, remove unused code/tests.

TODO: Create test case for creating/destroy array

Co-authored-by: Shreedhar Hardikar <shardikar@pivotal.io>


libgpos/src/memory/CMemoryPoolTracker.cpp
```diff
@@ -114,72 +117,65 @@ CMemoryPoolTracker::Allocate
...
+	void *ptr,
+	EAllocationType eat
 	)
 {
 	SAllocHeader *header = static_cast<SAllocHeader*>(ptr) - 1;
-	ULONG user_size = header->m_size;
+
+	ULONG user_size = header->m_user_size;
+	BYTE *alloc_type = static_cast<BYTE*>(ptr) + user_size;
+
```
zellerh: Could add a comment that this assert will catch writing past the allocated memory

libgpos/include/gpos/memory/CMemoryPool.h
```diff
@@ -222,10 +156,13 @@ namespace gpos
 				return 0;
 			}
 
-			// determine the size (in bytes) of an allocation that
-			// was made from a CMemoryPool
+			// requested size of allocation
+			static
+			ULONG UserSizeOfAlloc(const void *ptr);
+
+			// free allocation
```
zellerh: It might be helpful to have a comment somewhere in this file that explains a few rules. As far as I understand, here are some of the rules we have:

- While we specify a memory pool when allocating memory, we don't have a memory pool pointer when deleting memory, that pointer has to be determined somehow from the address to be deleted. How this is done depends on the memory pool used, and the memory pool manager ensures that this process can be done for all allocations.
- When deleting an array and retrieving the original user-specified allocated length, we can guarantee that the allocation has been done with type EatArray. Therefore we don't need to store the user-requested length for non-array allocations.

PR #540, opened 2019-10-03
Debugging code to count events, quantities, CPU time

zellerh:
This code makes it easy to count and measure various things in ORCA,
assuming you are using a debug build and are willing to modify the code.

The commit adds a new class CDebugCounter, which is enabled only in debug
builds (although that can be changed temporarily to measure retail builds,
by defining GPOS_DEBUG_COUNTERS).

You instrument the code by using the following macros to count events,
add up quantities (integers and floating point values) and to count CPU
time:

```
  GPOS_DEBUG_COUNTER_BUMP(counter_name)
  GPOS_DEBUG_COUNTER_ADD(counter_name, delta)
  GPOS_DEBUG_COUNTER_ADD_DOUBLE(counter_name, delta)
  GPOS_DEBUG_COUNTER_START_CPU_TIME(counter_name)
  GPOS_DEBUG_COUNTER_STOP_CPU_TIME(counter_name)
```

This commit adds code to ORCA to collect this information and write it
to the event log at the end of each statement compilation (again, this
is only enabled for debug builds unless changed).

You can have different counters, each one with a name chosen by you.
However, a given counter name can be used only with one of the measurement
types (count, sum of integers, sum of doubles, CPU time).

When you run multiple queries, ORCA will report the counter values separately
for each query. Any totals would have to be done by you with an SQL query
or a spreadsheet.

For convenience, you can name queries in a file, like so:

```
   select 'query name: q1';
   explain select * from foo;
```

In other words, a simple constant get with a string that starts with
"query name: " will be recognized by ORCA to indicate the name of the
next query. The counters will not be reported fir this first select,
only for the second query. Query names are useful when you have a test
with very many queries in it, e.g. the TPC-DS queries.

Finally, the commit contains a Python program to retrieve the recorded
counters from the log. The program can either print a series of comma-
separated values (CSV) to stdout or it can insert the result into an
SQL table name debug_counters in a database you can choose.

Run this for more info:

```
.../gporca/scripts/get_debug_event_counters.py --help
```


scripts/get_debug_event_counters.py
```diff
@@ -0,0 +1,236 @@
...
+	grep_command = 'grep CDebugCounterEvent '
+	gather_command = ['sh', '-c']
+
+	if logfile is None or len(logfile) == 0:
+		if 'MASTER_DATA_DIRECTORY' in os.environ:
+			master_data_dir = os.environ['MASTER_DATA_DIRECTORY']
+		else:
+			print("$MASTER_DATA_DIRECTORY environment variable is not defined, exiting")
+			exit()
+		grep_command = grep_command + master_data_dir + '/pg_log/*.csv'
```
chrishajas: This is looking through all log files, is that intentional?

libgpos/src/common/CDebugCounter.cpp
```diff
@@ -0,0 +1,310 @@
...
+		BOOL found = m_instance->FindByName(counter_name,
+											&key,
+											&val,
+											ECounterTypeCpuTime);
+
+		// note that we tolerate starting a timer but never stopping
+		// it (it will be ignored), but we assert that if a timer is
+		// being stopped and therefore logged, it has been started
+		GPOS_RTL_ASSERT(NULL != val->m_cpu_timer && val->m_timer_is_running);
+		val->m_counter_val_long += val->m_cpu_timer->ElapsedUS();
```
chrishajas: What's your guesstimate on the timing accuracy here for profiling many short-running operations (e.g.: as an extreme example, if I put a start_cpu_time and stop_cpu_time around a memory allocation).

Also, any particular reason for retail asserts?

zellerh: Good question, I have not yet tested CPU time measurements. It would be nice to have a nanosecond-based timer, microseconds are probably a bit crude for smaller sections of code. Hope to try it out sometimes and maybe I can look into changing it to a nanosecond CPU timer.

The reason for the retail assert is that we might measure a retail build and it is probably a good idea to ensure that we have the correct start/stop sequences.

PR #541, opened 2019-10-08
Fallback when citext op non-citext join predicate is present

hardikar:
When a query contains a join where one side is distributed on a citext
column, and the other isn't, ORCA may generate a plan that redistributes
the citext column. However, the redistribution is done using the hash
function in citext type's default opclass which may be different than
the function used to distribute the other side of the join. Thus, after
redistribution the data from the citext-distributed table may end up on
the wrong segment, ultimately returning incorrect results.

To fix this, ORCA must fully support opclasses in the distribution spec,
so that it can pick the correct hash function to redistribute. This
commit contains a temporary fix that falls back to planner whenever it
encounters a `citext op non-citext` or `non-citext op citext` join
predicate, ignore binary coercible casts.

REPRO:
```
create table tc (a int, c citext) ;
create table tt (b int, v varchar) distributed by (v);

insert into tc values (1, 'a'), (1, 'A');
insert into tt values (1, 'a'), (1, 'A');

insert into tc values (1, 'b'), (1, 'B');
insert into tt values (1, 'b'), (1, 'B');


-- orca
shardikar=#  select * from tc, tt where c = v;
 a | c | b | v
---+---+---+---
 1 | b | 1 | b
 1 | a | 1 | a
 1 | A | 1 | A
(3 rows)

-- planner
shardikar=# select * from tc, tt where c = v;
 a | c | b | v
---+---+---+---
 1 | b | 1 | b
 1 | a | 1 | a
 1 | A | 1 | A
 1 | B | 1 | B
(4 rows)
```


zellerh:
Just one more comment: You might want to fix the example in the text above, it doesn't match the DDL.

hardikar:
@chrishajas Done: https://github.com/greenplum-db/gpdb/pull/8794

PR #542, opened 2019-10-10
Better handling of null-rejecting predicates for ANDs and ORs

zellerh:
The methods that recognize null-rejecting predicates didn't handle
the case where an operand could be false or null, but not true.

We introduced a new Enum EberNotTrue to handle this case.

We also renamed the EberUnknown enum to EberAny, to avoid confusion
with the concept of "unknown" in the SQL standard, which means
something very similar to "null".


libgpopt/src/operators/CScalarBooleanTest.cpp
```diff
@@ -37,40 +37,46 @@ const WCHAR CScalarBooleanTest::m_rgwszBoolTest[EbtSentinel][30] =
...
-	{EbtIsNotUnknown, EberTrue, EberTrue},		// IS_NOT_UNKNOWN(True) = True
-	{EbtIsNotUnknown, EberFalse, EberTrue},		// IS_NOT_UNKNOWN(False) = True
-	{EbtIsNotUnknown, EberNull, EberFalse},		// IS_NOT_UNKNOWN(Null) = False ---> Note that NOT_UNKNOWN in BooleanTest means NOT-NULL
-	{EbtIsNotUnknown, EberUnknown, EberUnknown},	// IS_NOT_UNKNOWN(Unknown) = Unknown
+	{EbtIsUnknown, EberNotTrue, EberAny},	// IS_UNKNOWN(NotTrue) = Any
+	{EbtIsUnknown, EberAny, EberAny},		// IS_UNKNOWN(Any) = Any
+
+	{EbtIsNotUnknown, EberTrue, EberTrue},	// IS_NOT_UNKNOWN(True) = True
+	{EbtIsNotUnknown, EberFalse, EberTrue},	// IS_NOT_UNKNOWN(False) = True
+	{EbtIsNotUnknown, EberNull, EberFalse},	// IS_NOT_UNKNOWN(Null) = False ---> Note that NOT_UNKNOWN in BooleanTest means NOT-NULL
```
hardikar: Should we rename IS_NOT_UNKNOWN?

zellerh: This name comes from the PostgreSQL syntax: <expr> IS NOT UNKNOWN.

PR #544, opened 2019-10-14
Prefer gather with limit under it and bump down value of Epsilon

sambitesh:
This PR has two independent commits

Given our optimization framework produces two alternatives like below :
Limit (global)
Gather
Limit (local)
Limit (global)
Gather
We would like to pick the case 1 always. It is possible that our cardinality estimation severely
underestimated the cardinality and we can lose the benefit of having a local limit under gather.

0.001 was too high a value for Epsilon. In billion+ rows table with highly selective predicate,
we ended up severely under estimating the cardinality of predicate.
Co-authored-by: Sambitesh Dash sdash@pivotal.io
Co-authored-by: Chris Hajas chajas@pivotal.io

### DIFF
--- a/libgpdbcost/include/gpdbcost/CCostModelParamsGPDB.h
+++ b/libgpdbcost/include/gpdbcost/CCostModelParamsGPDB.h
@@ -103,7 +103,7 @@ namespace gpopt
 				EcpBitmapNDVThreshold, // bitmap NDV threshold
 				EcpBitmapScanRebindCost, // cost of rebind operation in a bitmap scan
 				EcpPenalizeHJSkewUpperLimit, // upper limit for penalizing a skewed hashjoin operator
-
+				EcpLocalLimitReward, // cost we save from having a limit under a gather
 				EcpSentinel
 			};
 
@@ -323,6 +323,10 @@ namespace gpopt
 			// upper limit for penalizing a skewed hash operator
 			static
 			const CDouble DPenalizeHJSkewUpperLimit;
+
+			// default value of rewarding a local limit under a gather motion
+			static
+			const CDouble DLocalLimitReward;
 
 			// private copy ctor
 			CCostModelParamsGPDB(CCostModelParamsGPDB &);
--- a/libgpdbcost/include/gpdbcost/CCostModelParamsGPDB.h
+++ b/libgpdbcost/include/gpdbcost/CCostModelParamsGPDB.h
@@ -103,7 +103,7 @@ namespace gpopt
 				EcpBitmapNDVThreshold, // bitmap NDV threshold
 				EcpBitmapScanRebindCost, // cost of rebind operation in a bitmap scan
 				EcpPenalizeHJSkewUpperLimit, // upper limit for penalizing a skewed hashjoin operator
-				EcpLocalLimitReward, // cost we save from having a limit under a gather
+				EcpLimitReward, // cost we save from having a limit
 				EcpSentinel
 			};
 
@@ -326,7 +326,7 @@ namespace gpopt
 
 			// default value of rewarding a local limit under a gather motion
 			static
-			const CDouble DLocalLimitReward;
+			const CDouble DLimitReward;
 
 			// private copy ctor
 			CCostModelParamsGPDB(CCostModelParamsGPDB &);
--- a/libgpdbcost/src/CCostModelGPDB.cpp
+++ b/libgpdbcost/src/CCostModelGPDB.cpp
@@ -19,6 +19,7 @@
 #include "gpopt/operators/CPhysicalIndexScan.h"
 #include "gpopt/operators/CPhysicalDynamicIndexScan.h"
 #include "gpopt/operators/CPhysicalHashAgg.h"
+#include "gpopt/operators/CPhysicalLimit.h"
 #include "gpopt/operators/CPhysicalUnionAll.h"
 #include "gpopt/operators/CPhysicalMotion.h"
 #include "gpopt/operators/CPhysicalPartitionSelector.h"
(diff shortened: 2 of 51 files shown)


ivannovick:
INTERESTING!

chrishajas:
For the epsilon change, can we add an example to the commit message and describe specifically what is happening. This value is used in translation, and since the percent remaining was lower than this epsilon value it wasn't sent to ORCA. 

For the gather change, we should add a new minidump test.

libgpdbcost/src/CCostModelParamsGPDB.cpp
```diff
@@ -171,6 +171,9 @@ const CDouble CCostModelParamsGPDB::DBitmapScanRebindCost(0.06);
 // see CCostModelGPDB::CostHashJoin() for why this is needed
 const CDouble CCostModelParamsGPDB::DPenalizeHJSkewUpperLimit(10.0);
 
+// see CCostModelGPDB::CostMotion() for why is it needed
+const CDouble CCostModelParamsGPDB::DLocalLimitReward(0.99);
```
ashuka24: How did you come up with a reward value of 0.99?

libgpdbcost/src/CCostModelGPDB.cpp
```diff
@@ -1367,6 +1380,27 @@ CCostModelGPDB::CostMotion
...
+		//	 Gather
+		//		<Expression tree>
+
+		COperator *popChild = exprhdl.Pop(0);
+		if (NULL != popChild && COperator::EopPhysicalLimit == popChild->Eopid())
+		{
+			costLocal = CCost(costLocal.Get() * localLimitReward);
+		}
+		
+	}
```
hardikar: I'd move this into `CostChildren()` if possible. It's where all the rest of these kinds of hacks live. ;)

libgpdbcost/src/CCostModelGPDB.cpp
```diff
@@ -242,10 +243,21 @@ CCostModelGPDB::CostUnary
...
 	CCost costLocal = CCost(num_rebinds * CostTupleProcessing(rows, width, pcp).Get());
 	CCost costChild = CostChildren(mp, exprhdl, pci, pcp);
 
+	// if it is a local limit, then we cost it to 0. We always want to have a local
+	// limit if it is available.
+	// e.g.
+	// Limit (global)
+	//   Gather
+	//		Limit (local) <-- this is costed 0
+	if(plimit && !plimit->FGlobal())
```
hardikar: Why is this needed, if we're discounting the Gather anyway?

libgpdbcost/src/CCostModelParamsGPDB.cpp
sambitesh: We didn't want to substantially change the cost of Limit->Gather->Limit. Just enough so that we pick it over Limit->Gather-><Expression Tree>. 0.99 is mall enough to achieve that.

libgpdbcost/src/CCostModelGPDB.cpp
sambitesh: This cannot be moved to CostChildren because we want to change the local cost of the Node here.

libgpdbcost/src/CCostModelGPDB.cpp
sambitesh: We want to pick 

Limit
 |_Gather
      |_Limit

over 

Limit
  |_Gather


The local `Limit` node has a cost associated with it. But we want it to be picked whenever it is available. So we are costing it to 0.

zellerh: Looks ok, but I'm wondering whether another solution might work as well and might be simpler: What if we reduce costChild by some factor here, since the limit operator might cancel the child:

```
  if (plimit)
    costChild *= pcmgpdb->GetCostModelParams()->PcpLookup(CCostModelParamsGPDB::EcpLocalLimitReward)->Get();
```

With this change, there should not be a need to change the CostMotion() method.

libgpdbcost/include/gpdbcost/CCostModelParamsGPDB.h
```diff
@@ -103,7 +103,7 @@ namespace gpopt
 				EcpBitmapNDVThreshold, // bitmap NDV threshold
 				EcpBitmapScanRebindCost, // cost of rebind operation in a bitmap scan
 				EcpPenalizeHJSkewUpperLimit, // upper limit for penalizing a skewed hashjoin operator
-
+				EcpLimitReward, // cost we save from having a limit under a gather
```
chrishajas: Not necessarily specific to a gather anymore.

libgpdbcost/src/CCostModelGPDB.cpp
```diff
@@ -242,10 +243,18 @@ CCostModelGPDB::CostUnary
 	DOUBLE rows = pci->Rows();
 	DOUBLE width = pci->Width();
 	DOUBLE num_rebinds = pci->NumRebinds();
-
+	CPhysicalLimit *plimit = dynamic_cast<CPhysicalLimit*>(exprhdl.Pop());
```
zellerh: If (and only if) you want to save a few microseconds, you could replace this with

```
BOOL is_limit = (COperator::EopPhysicalLimit == exprhdl.Pop()->Eopid());
```

libgpdbcost/include/gpdbcost/CCostModelParamsGPDB.h
sambitesh: Done thanks

libgpdbcost/src/CCostModelGPDB.cpp
sambitesh: Done thanks, every microsecond counts

libgpdbcost/src/CCostModelParamsGPDB.cpp
```diff
@@ -303,6 +306,7 @@ CCostModelParamsGPDB::CCostModelParamsGPDB
 	m_rgpcp[EcpBitmapNDVThreshold] = GPOS_NEW(mp) SCostParam(EcpBitmapNDVThreshold, DBitmapNDVThreshold, DBitmapNDVThreshold - 1.0, DBitmapNDVThreshold + 1.0);
 	m_rgpcp[EcpBitmapScanRebindCost] = GPOS_NEW(mp) SCostParam(EcpBitmapScanRebindCost, DBitmapScanRebindCost, DBitmapScanRebindCost - 1.0, DBitmapScanRebindCost + 1.0);
 	m_rgpcp[EcpPenalizeHJSkewUpperLimit] = GPOS_NEW(mp) SCostParam(EcpPenalizeHJSkewUpperLimit, DPenalizeHJSkewUpperLimit, DPenalizeHJSkewUpperLimit - 1.0, DPenalizeHJSkewUpperLimit + 1.0);
+	m_rgpcp[EcpLimitReward] = GPOS_NEW(mp) SCostParam(EcpLimitReward, DLimitReward, DLimitReward - 0.0001, DLimitReward + 0.0001);
```
hardikar: Use CStatistics::Epsilon, instead of hard coding 0.0001?

sambitesh:
Split the PR into two. For now merging only the change to value of epsilon. Will revisit the case of local limit later.

PR #546, opened 2019-10-22
Optimize statistics comparisons 

chrishajas:
This optimizes the logic around time comparisons, and removes some unnecessary dynamic casts. See the individual commits for details. With these changes, the TPC-DS optimization time goes from 67s to 56s on my laptop (~16% improvement).


libnaucrates/src/md/CMDIdGPDB.cpp
```diff
@@ -314,8 +314,8 @@ CMDIdGPDB::Equals
 	{
 		return false;
 	}
-	
-	const CMDIdGPDB *mdidGPDB = CMDIdGPDB::CastMdid(const_cast<IMDId *>(mdid));
+
+	const CMDIdGPDB *mdidGPDB = static_cast<CMDIdGPDB *>(const_cast<IMDId *>(mdid));
```
ashuka24: why can we do a static cast here?

chrishajas: We already check that the mdid is of type CMDIdGPDB above the changed line.

libnaucrates/src/base/IDatumStatisticsMappable.cpp
```diff
@@ -216,12 +216,16 @@ IDatumStatisticsMappable::StatsAreComparable
...
-			&& CMDTypeGenericGPDB::IsTimeRelatedType(datum_cast->MDId());
+
 	// the statistics for different time related types can't be directly compared, eg: timestamp vs timestamp with time zone.
 	// to prevent inaccurate statistics, mark as non-comparable
-	if (is_time_comparison && !is_types_match)
-		return false;
+	if (!is_types_match)
+	{
+		BOOL is_time_comparison = CMDTypeGenericGPDB::IsTimeRelatedType(this->MDId())
+		&& CMDTypeGenericGPDB::IsTimeRelatedType(datum_cast->MDId());
```
hardikar: indent.

chrishajas:
> Reg commits : "Optimize logic around time comparisons" & "Remove unnecessary dynamic cast " LGTM.
> 
> Regarding the last commit:
> 
> > Additionally, since we dereference it right after anyway, we're not
> > gaining anything from the dynamic cast. I added a GPOS_ASSERT with a dynamic
> > cast if we think it's useful.
> 
> It's not useful. The IDatum class is useless, if the class hierarchy you listed is correct. Time to get rid of this `IDatumStatisticsMappable` class. Do you see any reason not to?

@hardikar I merged the two classes and it seems to work out quite nicely! Please take another look when you have a chance.

PR #547, opened 2019-10-22
Optimize histogram copying to lazily copy buckets

chrishajas:
Previously, when constructing new histograms, we would copy all buckets
from an existing histogram. This was expensive, both in optimization
time and memory, as this is used in PstatsDerive().

Now, we lazily copy the underlying bucket array, `m_histogram_buckets` only when modifying the individual buckets. We only modify the buckets on 3 occasions:

1. CapNDVs(): We set the distinct values of the bucket in some
cases.

2. NormalizeHistogram(): We set the frequency of the bucket to
another value.

3. DistributeBucketProperties(): We set both the frequency and distinct
values of the bucket.

In the vast majority of cases, we're simply copying this array and
reading from it; modifications are relatively rare, which makes this
much more efficient.

From profiling on my laptop, I saw TPC-DS Q72 memory utilization go from
700MB to 270MB and optimization time go from 5.5s to 4s.

The entire TPC-DS optimization time went from 67s to 59s (12%
improvement).

I'm open to suggestions on making this less error-prone. I manually went through places we referenced `m_histogram_buckets` and made sure we aren't modifying it and added a `const` modifier to some of the function CBucketArray parameters.


libnaucrates/src/statistics/CStatisticsUtils.cpp
```diff
@@ -556,17 +560,29 @@ CStatisticsUtils::DistributeBucketProperties
...
+			GPOS_ASSERT(GPOPT_BUCKET_DEFAULT_FREQ == bucket->GetFrequency());
+			GPOS_ASSERT(GPOPT_BUCKET_DEFAULT_DISTINCT == bucket->GetNumDistinct());
+
 			CDouble factor = bucket->Width() / bucket_width;
-			bucket->SetFrequency(total_frequency * factor);
+			newBucket->SetFrequency(total_frequency * factor);
 			// TODO: , Aug 1 2013 - another heuristic may be max(1, dDisinct * factor)
-			bucket->SetDistinct(total_distinct_values * factor);
+			newBucket->SetDistinct(total_distinct_values * factor);
+
```
ashuka24: do we need to release `m_histogram_buckets` here?

chrishajas: I release it in CStatisticsUtils.cpp:: SplitHistBucketGivenMcvBuckets (L429) and `buckets` is made const in this function. I'm open to suggestions to make this clearer.

ashuka24: You could possibly add a comment to make it more clearer. I didn't realize this was in StatsUtils and not CHistogram which also plays a part

libnaucrates/src/statistics/CHistogram.cpp
```diff
@@ -553,15 +555,19 @@ CHistogram::CapNDVs
 	}
 
 	m_NDVs_were_scaled = true;
-
 	CDouble scale_ratio = (rows / distinct).Get();
+	// since we are modifying individual buckets, we must make a deep copy and overwrite m_histogram_buckets. We
+	// cannot modify m_histogram_buckets, as this is shared among histograms
```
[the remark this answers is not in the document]
chrishajas: I changed it to be a bit more verbose, but hopefully clearer.

libnaucrates/src/statistics/CHistogram.cpp
```diff
@@ -1004,10 +1013,20 @@ CHistogram::NormalizeHistogram()
 
 	CDouble scale_factor = std::max(DOUBLE(1.0), (CDouble(1.0) / GetFrequency()).Get());
 
-	for (ULONG ul = 0; ul < m_histogram_buckets->Size(); ul++)
+	// if the scale factor is 1.0, we don't need to copy the buckets
+	if (scale_factor != DOUBLE(1.0))
 	{
-		CBucket *bucket = (*m_histogram_buckets)[ul];
-		bucket->SetFrequency(bucket->GetFrequency() * scale_factor);
+		// since we are modifying individual buckets, we must make a deep copy and overwrite m_histogram_buckets. We
```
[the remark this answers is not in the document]
chrishajas: I copied the comment from above, let me know if it's better.

chrishajas:
@hardikar Thanks for the review! I addressed 1 and 2, and created a new constructor without passing in a CBucketArray (and changed the callers where possible). Please take a look at the last 3 commits.

server/src/unittest/dxl/statistics/CHistogramTest.cpp
```diff
@@ -241,7 +241,7 @@ CHistogramTest::PhistExampleInt4Remain
 		histogram_buckets->Append(bucket);
 	}
 
-	return GPOS_NEW(mp) CHistogram(mp, histogram_buckets, true, 0.1 /*null_freq*/, 2.0 /*distinct_remaining*/, 0.4 /*freq_remaining*/);
+	return GPOS_NEW(mp) CHistogram(mp, histogram_buckets, true, 0.1 /*null_freq*/, 2.0 /*distinct_remaining*/, 0.4 /*freq_remaining*/, false /*is_col_stats_missing*/);
```
ashuka24: why was this added?

chrishajas: This was leftover from something else I was trying. I've removed it.

hardikar:
Looks much better! Thanks!
Leaving a note here for the 2 times:
1. Use DeepCopyHistogramBuckets() in DistributeBucketProperties
2. There are many functions that use mp, instead of m_mp because it's passed in. We don't need to. Now's a good time to clean that up, but if in a rush, the current state is also good.

chrishajas:
> Looks much better! Thanks!
> Leaving a note here for the 2 times:
> 
> 1. Use DeepCopyHistogramBuckets() in DistributeBucketProperties
> 2. There are many functions that use mp, instead of m_mp because it's passed in. We don't need to. Now's a good time to clean that up, but if in a rush, the current state is also good.

@hardikar Thanks! I've fixed both of those.

PR #550, opened 2019-10-28
Fix "3rd party error log" message in the log with debug builds

zellerh:
The new CDebugCounter class caused error messages when we called
gpos_terminate in a debug build, because it didn't release its
memory pool before the CMemoryPoolManager that managed it got
destroyed.

The log messages started with "3rd party error log:".

The fix makes sure that we delete the static CDebugCounter object
and its associated memory pool before deleting the CMemoryPoolManager
in the gpos_terminate call.


libgpos/src/common/CDebugCounter.cpp
```diff
@@ -47,6 +47,13 @@ CDebugCounter::CDebugCounter(CMemoryPool *mp) :
 	m_hashmap = GPOS_NEW(mp) CounterKeyToValueMap(mp);
 }
 
+CDebugCounter::~CDebugCounter()
+{
+  CRefCount::SafeRelease(m_hashmap);
+  m_hashmap = NULL;
+  CMemoryPoolManager::GetMemoryPoolMgr()->Destroy(m_mp);
```
hardikar: I don't understand why this is needed. Isn't this done elsewhere already?

zellerh: Yes, CMemoryPoolManager::Shutdown() also destroys the memory pool if we don't do it here, but that will issue a warning in debug builds, the warning is logged by CMemoryPoolManager::DestroyMemoryPoolAtShutdown().

libgpos/src/common/CDebugCounter.cpp
```diff
@@ -59,6 +66,15 @@ void CDebugCounter::Init()
 	(void) amp.Detach();
 }
 
+void CDebugCounter::Shutdown()
+{
+  if (NULL != m_instance)
+  {
+	GPOS_DELETE(m_instance);
```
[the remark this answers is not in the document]
zellerh: Thanks, will fix that.

PR #552, opened 2019-10-31
Collapse inner and left outer joins into NAry joins

zellerh:
In the preprocessor, we only collapsed inner joins into NAry joins
so far. This commit also adds LOJs. The change is guarded by a
traceflag that is off by default, so this is dead code at this stage
until we deliver the needed optimizer changes.

We also change the algorithm to be more efficient.

- Renamed PexprCollapseInnerJoins to PexprCollapseJoins.
- Added a list of integers to LogicalNAryJoin. These integers
  are present only when the NAry join contains at least one
  LOJ. The indexes help in associating join predicates to
  children.
- Moved the CollectJoinChildren method from to CPredicateUtils
  to class CExpressionPreprocessor, because it now calls
  PexprCollapseJoins recursively.
- Added a new scalar operator, CScalarNAryJoinPredList
  that contains the inner join predicates as well as the ON
  predicates for LOJs.
- Fixed preprocessor issues that came up during unit testing.
- Changed MaxCard computation for both existing and new
  NAry left outer joins to include the inner table MaxCard as
  long as it is not 0.


libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -830,66 +831,124 @@ CExpressionPreprocessor::PexprCollapseInnerJoins
...
+			CExpression(mp,
+						GPOS_NEW(mp) CLogicalNAryJoin(mp, lojChildIndexes),
+						newChildNodes);
+
 		COptimizerConfig *optimizer_config = COptCtxt::PoctxtFromTLS()->GetOptimizerConfig();
 		ULONG ulJoinArityLimit = optimizer_config->GetHint()->UlJoinArityForAssociativityCommutativity();
 
-		// The last child of an n-ary join expression is the scalar expression
-		if (pexprResult->Arity() - 1 > ulJoinArityLimit)
+		if (pexprNAryJoin->Arity() - 1 > ulJoinArityLimit)
```
ashuka24: can you keep the comment here about the last child of the n-ary join being the scalar expression?

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -830,66 +831,124 @@ CExpressionPreprocessor::PexprCollapseInnerJoins
...
+		lojPredicates->Append(PexprCollapseJoins(mp, pexprScalar));
+
+		ULONG *indexOfThisLOJInTheArray = GPOS_NEW(mp)ULONG(lojPredicates->Size());
+		lojChildIndexes->Append(indexOfThisLOJInTheArray);
+	}
+	else
+	{
+		logicalLeafNodes->Append(PexprCollapseJoins(mp, pexpr));
+
+		ULONG *zeroIndex = GPOS_NEW(mp)ULONG(0);
```
ashuka24: It seems here that we would be adding a `zeroIndex` to `lojChildIndexes` every time we hit a leaf node, is this what we want? Could you please elaborate, in a comment or otherwise, what it means to add a `zeroIndex` to this array? Is there a reason that we do not add to this array in the inner join case?

hardikar: Can we add the following asserts wherever possible? 
`GPOS_ASSERT(logicalLeafNodes->Size() == lojChildIndexes->Size())`
(assuming that it is an invariant we are trying to maintain through the recursion)

Also 0 here represents that it is a non-loj predicate. Since, after completing `CollectJoinChildren`, all inner join predicates are collected into the 0th child of CScalarNAryJoinPredList. A comment explaining this here would be great, and/or using a #define const would also help.

Also I think something to note, is that loj's left child would come here if its a non-join operator. Let's add a comment here explaining some of the cases that would lead us into this else case. The right child of an loj is "attached" with its predicate via lojChildIndexes.

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -830,66 +831,124 @@ CExpressionPreprocessor::PexprCollapseInnerJoins
...
+		{
+			// just add the ANDed preds as the scalar child
+			newChildNodes->Append(CPredicateUtils::PexprConjunction(mp, innerJoinPredicates));
+			lojPredicates->Release();
+			lojChildIndexes->Release();
+			lojChildIndexes = NULL;
 		}
 
+		CExpression *pexprNAryJoin = GPOS_NEW(mp)
+			CExpression(mp,
```
hardikar: indentation

libgpopt/src/operators/CLogicalLeftOuterJoin.cpp
```diff
@@ -55,7 +55,17 @@ CLogicalLeftOuterJoin::DeriveMaxCard
 	)
 	const
 {
-	return CLogical::Maxcard(exprhdl, 2 /*ulScalarIndex*/, exprhdl.DeriveMaxCard(0));
+	CMaxCard maxCard = exprhdl.DeriveMaxCard(0);
+	CMaxCard maxCardInner = exprhdl.DeriveMaxCard(1);
+
+	// if the inner has a max card of 0, that will not make the LOJ's
+	// max card go to 0
+	if (1 <= maxCardInner.Ull())
```
hardikar: comment doesn't match the check:

if the inner has a max card of 0, => `maxCardInner.Ull() <= 1` ?

libnaucrates/include/naucrates/traceflags/traceflags.h
```diff
@@ -184,6 +184,9 @@ namespace gpos
 
 		// ExpandFullJoin
 		EopttraceExpandFullJoin = 103032,
+
+		// Expand LOJs in N-aryjoin
+		EopttraceEnableLOJInNAryJoin = 103033,
```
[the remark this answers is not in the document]
zellerh: The next phase of this project will be to support LOJs in the CXformExpandNAryJoinDPv2 transform. We imagine that we will set this traceflag when the user sets guc optimizer_join_order to 'exhaustive2'. 

I hope that we can eventually get rid of the CXformExpandNAryJoinDP transform and make CXformExpandNAryJoinDPv2 the default and map join order 'exhaustive' to 'exhaustive2'. At that time we can get rid of the traceflag again.

There aren't any changes in minidumps with this flag, since it only has an effect from the preprocessor to the exploration phase - except for some possible join order changes, of course.

libgpopt/src/base/CConstraint.cpp
```diff
@@ -230,6 +230,10 @@ CConstraint::PcnstrFromScalarExpr
 		case COperator::EopScalarCmp:
 			return PcnstrFromScalarCmp(mp, pexpr, ppdrgpcrs, infer_nulls_as);
 
+		case COperator::EopScalarNAryJoinPredList:
+			// return the constraints of the inner join predicates
+			return PcnstrFromScalarExpr(mp, (*pexpr)[0], ppdrgpcrs, infer_nulls_as);
```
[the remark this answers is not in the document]
hardikar: Ah. I see. `CLogicalLeftOuterJoin::DerivePropertyConstraint()` ignores its predicates.

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -858,24 +864,65 @@ CExpressionPreprocessor::PexprCollapseJoins
...
+					GPOS_ASSERT(GPOPT_ZERO_INNER_JOIN_PRED_INDEX == nextNumber);
+				}
+			}
+			GPOS_ASSERT(highestNumberSeen == lojPredicates->Size());
+#endif
 		}
 		else
 		{
-			// just add the ANDed preds as the scalar child
+			// no LOJs involved, just add the ANDed preds as the scalar child
```
ashuka24: do we want to add a check here that ensures the number of `innerJoinPredicates` matches that of the original `pexpr`?

libgpopt/src/operators/CExpressionPreprocessor.cpp
```diff
@@ -830,66 +831,184 @@ CExpressionPreprocessor::PexprCollapseInnerJoins
...
+		CExpression *rightChild = (*pexpr)[1];
+		CExpression *pexprScalar = (*pexpr)[2];
+
+		CollectJoinChildrenRecursively(mp, leftChild, logicalLeafNodes, lojChildPredIndexes, innerJoinPredicates, lojPredicates);
+
+		// stop collecting join children at the right child of the LOJ,
+		// just add the child, regardless of whether it is a join or not
+		logicalLeafNodes->Append(PexprCollapseJoins(mp, rightChild));
+
+		// create an entry in lojPredicates...
```
ashuka24: can we update this comment with the new array name?

libgpopt/src/operators/CExpressionPreprocessor.cpp
sambitesh: The `innerJoinPredicates` are collected by calling `CollectJoinChildrenRecursively` on pexpr. So the check here would be redundant.

PR #553, opened 2019-11-05
Consider ArrayCmp in IndexScans only for Bitmap indices

zellerh:
We only support array comparison predicates on bitmap index scans, not on
btree index scans or index scans of other types. However, ORCA was attempting
to generate such plans, leading to a fallback to planner.

With the fix, we will only consider bitmap indexes for array comparison
predicates. This should avoid the fallback.

Here is a simple test case:

```
drop table if exists foo;
create table foo (a int, b int) distributed by (a);
insert into foo values (1,1);
create index foo_ix on foo using btree (b);

set optimizer_print_plan to on;
set client_min_messages to 'log';

-- before: btree index scan with an array comparison, fallback
-- after: file scan with array comparison, ORCA plan
explain select * from foo where b in (1, 3);

-- before: btree index scan with array comparison, fallback, planner error in 5X
-- after: file scan with array comparison, ORCA plan, succeeds in 5X
explain update foo set a = 1 where b in (1, 3);
```


zellerh:
@hardikar, thanks, good idea with the ICG test. This test does exist and it no longer shows a fallback. Here is my commit with the updated result: https://github.com/zellerh/gpdb/commit/29752f68b12c5e53d461ebeba981ec66d70c99c0
(there might be more once the ICG pipeline finishes)

hardikar:
Perfect, Thanks for checking!

chrishajas:
In 6 and master, planner is able to generate and execute both btree and bitmap scans with  `explain update foo set a = 1 where b in (1, 3);` for the test case above. Should we instead allow such a plan to be generated (and not fall back)? I don't know if this also applies to 5, since planner doesn't generate a plan updating multiple distribution key columns in 5.

libgpopt/src/operators/CPredicateUtils.cpp
```diff
@@ -2069,7 +2069,7 @@ CPredicateUtils::PexprIndexLookup
 	{
 		cmptype = CScalarCmp::PopConvert(pexprScalar->Pop())->ParseCmpType();
 	}
-	else if (CUtils::FScalarArrayCmp(pexprScalar))
+	else if (CUtils::FScalarArrayCmp(pexprScalar) && IMDIndex::EmdindBitmap == pmdindex->IndexType())
```
ashuka24: Prior to 976cba9b0, btree indexes on AO tables were sent to ORCA as type bitmap. Now, they are sent to ORCA as type btree. Would this conditional impact plans for AO tables with btree indexes?

ashuka24: Hans and I spoke on this issue and it seems like AO tables with btree indexes will not be affected by this PR. This is most likely due to the fact that such indexes do not follow the btree path, and instead, follow the bitmap path, completely avoiding the issue found here.

zellerh:
> In 6 and master, planner is able to generate and execute both btree and bitmap scans with `explain update foo set a = 1 where b in (1, 3);` for the test case above. Should we instead allow such a plan to be generated (and not fall back)? I don't know if this also applies to 5, since planner doesn't generate a plan updating multiple distribution key columns in 5.

@chrishajas,  you are right that this PR alone would cause a regression for IN predicates on a btree index: Before, such a plan would fall back and generate a bitmap scan in planner. Now, with this fix, we stay in ORCA and don't consider the index and generate a table scan instead. PR #554 will fix this issue and allow bitmap scans. So, with both PRs #553 and #554 there will only be a slight regression in that ORCA does not consider a regular index scan, while previously we could get that when ORCA fell back for this type of query. I hope that will be acceptable.

Here is one repro from file data/dxl/minidump/BTreeIndex-Against-InList.mdp:

```
CREATE TABLE test as SELECT * FROM generate_series(1,1000000) as a distributed randomly;
CREATE INDEX test_index ON test(a);
SELECT * FROM test WHERE a in (1, 47);
```

PR #554, opened 2019-11-13
Consider Bitmap scans for OR and IN predicates on btree indexes

zellerh:
When we have predicates that are too complicated for BTree index scans,
then also consider bitmap index scans.

- OR predicates: This is implemented by passing
a boolean argument, indicating that we are processing a complex predicate
and by adding an additional parameter to allow btree indexes in addition
to the primary index type, bitmap.

- IN predicates: These will show as an ArrayCmp. The fix is to allow a btree
index to be used in a bitmap scan when we see an ArrayCmp predicate.

The change also required updating some older MDP files with the return
type of the index, since this is now required for BTree indexes as well
as for bitmap indexes (that has been the case since commit 976cba9).

Finally, fixed an assert violation by ensuring we don't calculate a negative
NDV remaining value in CHistogram::ComputeJoinNDVRemainInfo().

Note that the plan in data/dxl/minidump/BTreeIndex-Against-InList.mdp
will change from a sequential scan to a bitmap scan once PR #553
has been merged and this code has been rebased. This will serve as
the test for an IN predicate on a btree index.


zellerh:
Please note that this is meant to go in after PR #553.

libgpopt/src/xforms/CXformUtils.cpp
```diff
@@ -3344,7 +3348,8 @@ CXformUtils::PexprBitmapSelectBestIndex
 	CColRefSet *pcrsReqd,
 	CColRefSet *pcrsOuterRefs,
 	CExpression **ppexprRecheck,
-	CExpression **ppexprResidual
+	CExpression **ppexprResidual,
+	BOOL alsoConsiderBTreeIndexes
```
ashuka24: it seems like this maps to `isAPartialPredicate` which can be a bit confusing without a comment. can we add one here?

libgpopt/src/xforms/CXformUtils.cpp
```diff
@@ -3936,7 +3961,8 @@ CXformUtils::PexprBitmapTableGet
 				pcrsReqd,
 				fConjunction,
 				&pexprRecheck,
-				&pexprResidual
+				&pexprResidual,
+				false
```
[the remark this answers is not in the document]
zellerh: Done.

libgpopt/src/xforms/CXformUtils.cpp
zellerh: Yes, I'll add a comment where the name changes.

PR #558, opened 2019-12-24
Do not pushdown predicate when it contains CTEAnchor

asubramanya:
If there is a predicate which contains a CTEAnchor, pushing it down can cause
duplicate CTEProducers to be generated in the plan. So for now disable pushing
down the predicate in such cases.


asubramanya:
@hardikar It seems to work fine for the query you have posted. I don't see any duplicate CTE producers in the plan.

libgpopt/src/operators/CNormalizer.cpp
```diff
@@ -1151,7 +1151,7 @@ CNormalizer::PushThru
 	{
 		CExpression *pexprConj = (*pdrgpexprConjuncts)[ul];
 		pexprConj->AddRef();
-		if (FPushable(pexprLogical, pexprConj))
+		if (FPushable(pexprLogical, pexprConj) && !CUtils::FHasCTEAnchor(pexprConj))
```
chrishajas: Might be good to add a comment saying why we can't push down CTE anchors (or maybe even put this function call inside `FPushable()`)

asubramanya: Done.

libgpopt/src/base/CUtils.cpp
```diff
@@ -897,6 +897,32 @@ CUtils::FHasSubquery
 	return pexpr->DeriveHasSubquery();
 }
 
+// check for existence of CTE anchor
+BOOL
+CUtils::FHasCTEAnchor
+(
+ CExpression *pexpr
+ )
```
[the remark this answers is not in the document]
asubramanya: Done

PR #560, opened 2020-01-07
Fix PdsRequired for CPhysicalAssert

asubramanya:
When we want to insert data into a randomly distributed table, we need a
redistribute motion to make sure that the data is indeed randomly distributed.
In order to do this, the CPhysicalDML node requests a STRICT RANDOM motion from
its child. The code assumed that the child of a CPhysicalDML would always be a
CPhysicalComputeScalar. But in case of a not null column, the child will be a
CPhysicalAssert and its child would be CPhysicalComputeScalar. The PdsRequired
method for CPhysicalAssert node did not consider distribution requests of type
STRICT RANDOM and hence the CPhysicalComputeScalar node never received the
STRICT RANDOM request. Thus the optimization would fail since it would not be
able to compute a plan with the required properties. The fix is to pass through
the distribution request to its children.


libgpopt/src/operators/CPhysicalAssert.cpp
```diff
@@ -138,7 +138,8 @@ CPhysicalAssert::PdsRequired
 	// pass through singleton and broadcast requests
 	if (CDistributionSpec::EdtSingleton == edt ||
 		CDistributionSpec::EdtStrictSingleton == edt ||
-		CDistributionSpec::EdtReplicated == edt
+		CDistributionSpec::EdtReplicated == edt ||
+		CDistributionSpec::EdtStrictRandom == edt
```
vraghavan78: need to update the comment above

asubramanya: updated.

PR #561, opened 2020-01-09
WIP: Add new nth-root method to calculate join damping factor

chrishajas:
The join damping factor controls how much impact successive predicates
have on the cardinality of a join. For example, given the join condition
"a=b and c=d and e=f", we currently sort the predicates descending by
selectivity and the 2nd predicate does not have as much influence as the
first, depending on the `optimizer_join_factor` GUC. By default, this is
set to .01, meaning the 2nd predicate's contribution to the cardinality
is multiplied by (.01)^2, and the 3rd predicate is multiplied by (.01)^3.

This effectively assumes that predicates are fairly highly correlated.
Although increasing the default value of this GUC was considered (which
would make the assumption that the columns are a bit more independent),
another approach is to scale the successive predicates down moderately
based on their selectivity.

This new method (I'll call it the nth root method) makes successive
predicates have a smaller and smaller impact to account for the higher
probability that there is some correlation between the columns.

For example, given ANDed predicates with selectivities [.5, .3, .1], the
cumulative selectivity would be:
S = S1 * sqrt(S2) * 4root(S3)
.15 = .5 * sqrt(.3) * 4root(.1)
For scale factors, this is equivalent to SF1 * sqrt(SF2) * 4root(SF3)

Using this nth-root method causes us to estimate much lower
cardinalities than before for joins with multiple predicates, causing
plans to use more indexes. This method will not be used by default and
will only be used when the `optimizer_join_factor` GUC is set to 0.

### DIFF
--- a/data/dxl/minidump/MultiplePredJoinCardinality.mdp
+++ b/data/dxl/minidump/MultiplePredJoinCardinality.mdp
@@ -1,0 +1,4828 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<dxl:DXLMessage xmlns:dxl="http://greenplum.com/dxl/2010/12/">
+    <dxl:Comment><![CDATA[
+Objective: The join cardinality estimate for multiple uncorrelated columns should be accurate.
+The query returns 1-5 rows when run, and we should estimate 50-100 rows or lower when
+optimizer_damping_factor_join is set to 0. When `optimizer_damping_factor_join` is set
+to .01 (the default), we estimate ~1500 rows returned.
+
+create table t1 (a int, b int) distributed by (a);
+create table t2 (a int, b int) distributed by (a);
+insert into t1 select floor(random()*1000),floor(random()*1000) from generate_series(1,1000)i;
+insert into t2 select floor(random()*1000),floor(random()*1000) from generate_series(1,1000)i;
+analyze t1;
+analyze t2;
+explain select * from foo, bar where foo.a=bar.a and foo.b=bar.b;
+
+                                   QUERY PLAN
+--------------------------------------------------------------------------------
+ Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..862.27 rows=62 width=16)
+   ->  Hash Join  (cost=0.00..862.27 rows=21 width=16)
+         Hash Cond: foo.a = bar.a AND foo.b = bar.b
+         ->  Table Scan on foo  (cost=0.00..431.01 rows=334 width=8)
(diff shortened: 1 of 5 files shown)


zellerh:
This is definitely an improvement, but I would suggest we go one step further. Here is a motivating example: @sambitesh and I have been looking at TPC-DS query 19, and focused in on this simplified version of Q19:

```
explain
select  *
  from date_dim, store_sales, item
 where d_date_sk = ss_sold_date_sk
   and ss_item_sk = i_item_sk
   and i_manager_id=7
   and d_moy=11
   and d_year=1999
;
```

We looked at two ways to compute the cardinality of this join:

```
date_dim cross join item inner join store_sales
```

and

```
date_dim inner join store_sales inner join item
```

In the first case, we multiply the cardinalities of date_dim and item to get a cardinality for the cross product. Then, we multiply that by the cardinality of store_sales and the selectivities of the two join predicates. We also use the damping factor, as we have two join predicates:

```
card = |d| * |i| * |ss| * sel(p1) * 0.0001 * sel(p2)
```

- |d| = cardinality of date_dim
- |i| = cardinality of item
- |ss| = cardinality of store_sales
- sel(p1) = selectivity of the join predicate between date_dim and store_sales
- sel(p2) = selectivity of the join predicate between item and store_sales
- 0.0001 = damping factor for second predicate, 0.0001 until now, without this PR

In the second case, we have one join predicate for each join, so we don't apply a damping factor:

```
card = |d| * |ss| * sel(p1) * |i| * sel(p2)
```

In other words, depending on the join order, the cardinality will be 10,000 times higher if we pick the second join order. IMHO this is not acceptable. When we do join enumeration, the cardinality is key to getting the right join order, and when the cardinality estimates are radically different for different join orders, that means we don't have an apples-to-apples comparison.

Here is our suggestion:

- Use the damping factor as you implemented it when we have multiple predicates linking the _same_ two tables, like `t1.a=t2.a and t1.b=t2.b`.
- Don't use a damping factor when we are dealing with join predicates that link _different_ tables, like `t1.a=t2.a and t1.b=t3.b`. This will make cardinality estimates much less dependent of the join order.

If you prefer, we could combine this suggested change with our work on the DPv2 transform (we hope to submit a PR for that in the near future).

By the way, we tried TPC-DS with your change, in the hope your fix would give us a better plan, but even the smaller damping factor used with your n-th root formula still caused issues in our case.

ashuka24:
Closing in favor of https://github.com/greenplum-db/gporca/pull/567

PR #566, opened 2020-01-16
Better cardinality estimation for citext

hardikar:
Currently we only derive cardinality for built-in types, by using a hard coded list of mdids. This means, other text-like types such as citext have to fall back to using default cardinality of 40%. This PR extends ORCA to support citext and other text-like types in a similar manner to text & varchar.

GPDB side PR: https://github.com/greenplum-db/gpdb/pull/9440


hlinnaka:
I'm a bit confused by the commit message. What does it mean for "citext have to fall back to using default cardinality of 40%" ? Surely the cardinality estimate depends on the operator being used? What does ORCA do for text, that it doesn't currently do for citext?

hardikar:
Yes the cardinality doesn’t depend on specific operators. Currently ORCA does histogram analysis to determine the selectivity of predicates etc only for certain built in types such as text. Otherwise (eg for citext) it estimates it using the default selectivity of 0.4. 
For text types (bpchar, text, varchar) ORCA does this analysis only for = and != operators and no other operators.

Hope that explains that statement better.

libnaucrates/src/md/CMDIdGPDB.cpp
```diff
@@ -74,6 +77,9 @@ CMDIdGPDB CMDIdGPDB::m_mdid_varchar(GPDB_VARCHAR);
 // text mdid
 CMDIdGPDB CMDIdGPDB::m_mdid_text(GPDB_TEXT);
 
+// text mdid
+CMDIdGPDB CMDIdGPDB::m_mdid_name(GPDB_NAME);
```
vraghavan78: does this change with the version of GPDB?

libnaucrates/src/base/CDatumGenericGPDB.cpp
```diff
@@ -319,7 +322,9 @@ CDatumGenericGPDB::IsDatumMappableToDouble() const
 BOOL
 CDatumGenericGPDB::IsDatumMappableToLINT() const
 {
-	return CMDTypeGenericGPDB::HasByte2IntMapping(this->MDId());
+	CMDAccessor *md_accessor = COptCtxt::PoctxtFromTLS()->Pmda();
+	const IMDType *type = md_accessor->RetrieveType(MDId());
```
chrishajas: My only concern would be if `RetrieveType` is somewhat expensive. This is used when comparing buckets, and I think it might be a fairly hot codepath.

libnaucrates/include/naucrates/md/IMDType.h
```diff
@@ -142,6 +142,12 @@ namespace gpmd
 			virtual
 			BOOL IsComposite() const = 0;
 
+			virtual
+			BOOL IsTextRelated() const
```
ashuka24: is there a reason why this is not a pure virtual function like the others?

libnaucrates/src/base/CDatumGenericGPDB.cpp
chrishajas: I ran tpcds explain tests and there weren't any measurable differences in optimization time, so we should be ok.

libnaucrates/src/md/CMDIdGPDB.cpp
asubramanya: No.

libnaucrates/src/base/CDatumGenericGPDB.cpp
asubramanya: I believe the MDAccessor caches the results when it fetches the metadata for the first time, so subsequent calls shouldn't be expensive.

libnaucrates/include/naucrates/dxl/parser/CParseHandlerMDType.h
```diff
@@ -108,6 +108,9 @@ namespace gpdxl
 			// is type composite
 			BOOL m_is_composite;
 
+			// is type composite
+			BOOL m_is_text_related_type;
```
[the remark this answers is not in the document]
asubramanya: Done

libnaucrates/include/naucrates/dxl/parser/CParseHandlerMDType.h
```diff
@@ -108,6 +108,9 @@ namespace gpdxl
 			// is type composite
 			BOOL m_is_composite;
 
+			// is type composite
```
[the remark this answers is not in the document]
asubramanya: Done

libnaucrates/include/naucrates/md/IMDType.h
asubramanya: I guess the reason is to avoid implementing the function in all the derived classes since we know that except `CMDTypeGenericGPDB` the other classes cannot be text related.

PR #570, opened 2020-02-07
Change bitmap index costing to choose bitmap NLJs more often 

ashuka24:
The experiments and assertions made below were found using the
cal_bitmap_test.py calibration script. We used regression analysis and
isolated a single variable to determine the coefficients.

This commit makes substantial changes to costing bitmap indexes. Our
goal was to choose bitmap index NL joins instead of hash joins, as the
execution time of the bitmap NL joins was 10X+ less than hash joins
in many cases.

Previously, we were multiplying the rebinds by the bitmap page cost
which caused the cost to be much more expensive than a hash join in
many cases.

Now, we no longer multiply the page cost by the number of rebinds, and
instead multiply the rebinds by a much smaller rebind cost.

Additionally, we took this opportunity to simplify the cost model a bit
by removing the separate code path for small vs large NDVs. We did not
see the large NDV path being used in joins, and in non-join cases it had
very minimal impact on the cost.

This functionality is guarded by a traceflag, EopttraceCalibratedBitmapIndexCostModel.
In GPDB, it will be enabled by setting
`optimizer_cost_model=experimental`. The intent is to enable this by
default in the near future.


libgpdbcost/src/CCostModelGPDB.cpp
```diff
@@ -1561,15 +1561,33 @@ CCostModelGPDB::CostBitmapTableScan
...
+			}
+			else
+			{
+				result = CostBitmapSmallNDV(pcmgpdb, pci, dNDV);
+			}
 		}
 		else
 		{
-			result = CostBitmapSmallNDV(pcmgpdb, pci, dNDV);
+			// The numbers below were experimentally determined using regression analysis in the cal_test.py script
```
hardikar: cal_test.py was renamed, right?

libgpdbcost/src/CCostModelGPDB.cpp
```diff
@@ -1561,15 +1561,33 @@ CCostModelGPDB::CostBitmapTableScan
...
+			CDouble dSize = rows * (1 + std::max(width * 0.005, 1.0)) * 0.05;
```
hardikar: Can you add a comment about the derivation of this formula: `rows * (1 + std::max(width * 0.005, 1.0)) * 0.05;`?

sambitesh:
I have a concern. The new cost model for BitmapScan climbs too fast compared to our TableScan cost which is practically a line parallel to x-axis. The implication of this formula is that for higher cardinality, we will never pick a BitmapScan. 


I was just considering the case of scan on a predicate and not joins.

Consider a table with 1 billion rows. Even if we have a predicate with selectivity 1%, we still won't be picking BitmapScan with this new formula.

PR #575, opened 2020-03-01
Check if there is direct dispatchable filter

sambitesh:
Consider two queries :

a) Explain Select RANK FROM ( select RANK() OVER ( order BY dist_key asc ) as
RANK, *  from foo )A WHERE '2014-06-29' = dist_key ;

b) Explain  Select RANK FROM ( select RANK() OVER ( order BY dist_key asc ) as
RANK, *  from foo WHERE '2014-06-29' = dist_key )A;

Query 'a' gives a plan like below with a filter over Window. So in spite of
having a filter on distribution key, we shouldn't direct dispatch.

	Filter: dist_key= '2014-06-29'::date
   	->  Window
         	Order By: dist_key

Query 'b' on other hand has a filter directly over a table scan, so we should
direct dispatch this plan.

We had a very simple heuristic to decide if a plan is direct dispatchable or
not. Now we add a check that if anywhere in the plan we run into a
'CPhyscialFilter' which is not above a scan, we bail out from direct
dispatching. We are erring on the side of caution here.


libgpopt/include/gpopt/base/COptCtxt.h
```diff
@@ -160,6 +163,11 @@ namespace gpopt
 				m_has_replicated_tables = true;
 			}
 
+			void SetFalseHasDirectDispatchableFilter()
```
hardikar: Rename to ResetHasDirectDispatchableFilter.

libgpopt/src/translate/CTranslatorExprToDXLUtils.cpp
```diff
@@ -2620,4 +2627,28 @@ CTranslatorExprToDXLUtils::FMotionHazardSafeOp
...
+
+BOOL
+CTranslatorExprToDXLUtils::FDirectDispatchFilter
+	(
+	 CExpression *pexprFilter
+	)
+{
+	GPOS_ASSERT(NULL != pexprFilter);
+
+	CExpression *pexprRelational = (*pexprFilter)[0];
```
hardikar: So this is `child_expr`
and `child_op` then?

libgpopt/src/translate/CTranslatorExprToDXLUtils.cpp
```diff
@@ -2029,6 +2029,13 @@ CTranslatorExprToDXLUtils::SetDirectDispatchInfo
 		// direct dispatch not supported for join queries
 		return;
 	}
+
+	// if we have filter not on top of Physical Scan, then we should not
```
hardikar: This comment is maybe out of sync of the IF-checks in `FDirectDispatchFilter`

libgpopt/include/gpopt/translate/CTranslatorExprToDXLUtils.h
```diff
@@ -585,6 +585,10 @@ namespace gpopt
 			// extract the column ids of the ident from project list
 			static
 			void ExtractIdentColIds(CDXLNode *dxlnode, CBitSet *pbs);
+
+			// is this Filter node direct dispatchable?
+			static
+			BOOL FDirectDispatchFilter (CExpression *pexprFilter);
```
ashuka24:
```suggestion
			BOOL FDirectDispatchableFilter (CExpression *pexprFilter);

```

Can we be consistent about using `DirectDispatchable` and `DirectDispatch`? It seems like everywhere else we use "dispatchable"

libgpopt/src/translate/CTranslatorExprToDXLUtils.cpp
sambitesh: Yes, I can rename the variables for clarity

libgpopt/src/translate/CTranslatorExprToDXLUtils.cpp
sambitesh: Will update it thanks

sambitesh:
@hardikar today ORCA applies a very naive heuristic to apply direct dispatch which can be found here : `CTranslatorExprToDXLUtils::SetDirectDispatchInfo`

Basically, if it is not a join query and it has any filter on distribution key, then it makes it direct dispatchable. Which is fine in all cases except when we have filter over window functions.

libgpopt/src/base/COptCtxt.cpp
```diff
@@ -55,7 +55,8 @@ COptCtxt::COptCtxt
 	m_fDMLQuery(false),
 	m_has_master_only_tables(false),
 	m_has_volatile_or_SQL_func(false),
-	m_has_replicated_tables(false)
+	m_has_replicated_tables(false),
+	m_has_direct_dispatchable_filter(true)
```
chrishajas: The naming here is a bit confusing. If there are multiple filters, and one is not direct "dispatchable", then we set this to false. Maybe `m_all_filters_support_direct_dispatch` or even `m_consider_direct_dispatch`, just something to indicate that any filter that doesn't support direct dispatch makes the boolean false.

sambitesh:
@chrishajas At present ORCA is very cautious in deciding direct dispatch while Planner has a walker `AssignContentIdsToPlanData_Walker`. We are being consistent with ORCA approach here. I ran the ICG and none of the plans which we were direct dispatching earlier changed.

libgpopt/src/translate/CTranslatorExprToDXLUtils.cpp
```diff
@@ -2620,4 +2626,28 @@ CTranslatorExprToDXLUtils::FMotionHazardSafeOp
...
+
+	// a filter can lead to direct dispatch if it is over a scan
+	// or partition selector. If it is over a window, it shouldn't
+	// lead to a direct dispatch
+	return (CUtils::FPhysicalScan(pop) ||
+			COperator::EopPhysicalPartitionSelector == pop->Eopid() ||
+			COperator::EopPhysicalFilter == pop->Eopid() || // nested filters allowed
+			COperator::EopPhysicalComputeScalar == pop->Eopid() // project list allowed
+			);
+
```
zellerh: Shouldn't we go down to children and grandchildren, allowing partition selectors, filters and projects, until we either find an unsupported node and a scan node? The current code would allow a project as a child without checking that there isn't an unsupported node below it.

libgpopt/src/translate/CTranslatorExprToDXL.cpp
```diff
@@ -1486,6 +1486,11 @@ CTranslatorExprToDXL::PdxlnFromFilter
 	CExpression *pexprRelational = (*pexprFilter)[0];
 	CExpression *pexprScalar = (*pexprFilter)[1];
 
+	if (!CTranslatorExprToDXLUtils::FDirectDispatchableFilter(pexprFilter))
+	{
+		COptCtxt::PoctxtFromTLS()->ResetHasDirectDispatchableFilter();
+	}
```
zellerh: We might come across an unsupported filter first and then see a supported filter. Wouldn't we allow direct dispatch if that would happen? Overall, I don't really like the current code that uses the global constraint of the query to set up direct dispatch. It might be better to use the constraints of the direct-dispatchable filter instead. If we would do that, we would not have to check for any other filters. So, maybe store a pointer to the direct-dispatchable filter in the optimization context, instead of a boolean?

sambitesh:
This PR fixes this issue

https://github.com/greenplum-db/gporca/issues/505

libgpopt/include/gpopt/base/COptCtxt.h
```diff
@@ -163,9 +163,10 @@ namespace gpopt
 				m_has_replicated_tables = true;
 			}
 
-			void ResetHasDirectDispatchableFilter()
+			void SetHasDirectDispatchableFilter(CExpression *filter_expression)
```
zellerh: A nit: Maybe rename the method to something like `AddDirectDispatchableFilterCandidate` or something a bit shorter :-)

libgpopt/include/gpopt/base/COptCtxt.h
```diff
@@ -183,9 +184,9 @@ namespace gpopt
 				return m_has_replicated_tables;
 			}
 
-			BOOL HasDirectDispatchableFilter() const
+			CExpressionArray* HasDirectDispatchableFilter() const
```
zellerh: A similar nit: Since this no longer returns a BOOL, maybe rename to something that doesn't start with "Has"?

libgpopt/src/translate/CTranslatorExprToDXLUtils.cpp
```diff
@@ -2029,31 +2029,64 @@ CTranslatorExprToDXLUtils::SetDirectDispatchInfo
...
-	if (CDistributionSpec::EdtHashed != pds->Edt())
-	{
-		// direct dispatch only supported for scans over hash distributed tables 
-		return;
-	}
-	
-	CPropConstraint *ppc = pexpr->DerivePropertyConstraint();
-	if (NULL == ppc->Pcnstr())
+
+	// check CTranslatorExprToDXLUtils::FDirectDispatchableFilter
```
zellerh: This comment doesn't apply to all of the expressions in the array (it's not involved for bitmap and index scans), so maybe remove it?

sambitesh: Done, thanks

libgpopt/src/translate/CTranslatorExprToDXLUtils.cpp
```diff
@@ -2620,4 +2652,33 @@ CTranslatorExprToDXLUtils::FMotionHazardSafeOp
...
+	 CExpression *pexprFilter
+	)
+{
+	GPOS_ASSERT(NULL != pexprFilter);
+
+	CExpression *pexprChild = (*pexprFilter)[0];
+	COperator *pop = pexprChild->Pop();
+
+	// find the first child or grandchild of filter which is not
+	// a Project, Filter or PhysicalComputeScalar (result node)
```
chrishajas: is this supposed to include PS in the comment here?

libgpopt/src/translate/CTranslatorExprToDXL.cpp
```diff
@@ -536,6 +535,8 @@ CTranslatorExprToDXL::PdxlnIndexScan
 	GPOS_ASSERT(NULL != pexprIndexScan);
 	CDXLPhysicalProperties *dxl_properties = GetProperties(pexprIndexScan);
 
+	COptCtxt::PoctxtFromTLS()->AddDirectDispatchableFilterCandidate(pexprIndexScan);
```
chrishajas: Just for clarity, the intention here is to add filter candidates from all scans?

libgpopt/src/translate/CTranslatorExprToDXLUtils.cpp
```diff
@@ -2029,31 +2029,63 @@ CTranslatorExprToDXLUtils::SetDirectDispatchInfo
...
+	{
+		// direct dispatch only supported for scans over hash distributed tables
+		for (ULONG i = 0; i < size; i++)
+		{
+			CExpression *pexprFilter = (*pexprFilterArray)[i];
+			CDrvdPropRelational *prop_relational = pexprFilter->GetDrvdPropRelational();
+
+			if (NULL != prop_relational)
+			{
+				CPropConstraint *ppc = prop_relational->GetPropertyConstraint();
```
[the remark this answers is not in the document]
sambitesh: I added this because otherwise the the unit test `CTranslatorDXLToExprTest` was giving seg fault. The normal mdp test were okay with it.

libgpopt/src/translate/CTranslatorExprToDXL.cpp
sambitesh: Yes, add all valid filters and find the first filter which gives you a direct dispatch

libgpopt/src/base/COptCtxt.cpp
```diff
@@ -87,6 +88,7 @@ COptCtxt::~COptCtxt()
 	m_pcteinfo->Release();
 	m_optimizer_config->Release();
 	CRefCount::SafeRelease(m_pdrgpcrSystemCols);
+	CRefCount::SafeRelease(m_direct_dispatchable_filters);
```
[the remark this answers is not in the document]
sambitesh: This cannot be NULL, true, we always instantiate it in constructor.  I was just being extra cautious.

sambitesh:
@chrishajas I do mention it in commit message that filter over window should not be considered for direct dispatch. I will reword it

> Query 'a' gives a plan like below with a filter over Window. So in spite of
having a filter on distribution key, we shouldn't direct dispatch.

libgpopt/src/translate/CTranslatorExprToDXLUtils.cpp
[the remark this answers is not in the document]
chrishajas: Interesting, `GetDrvdPropRelational()` should only be called if all properties have been derived, otherwise bad things can happen (and we've put a retail assert to ensure we don't hit those). We don't seem to explicitly make that assumption here. I'm curious why `pexprFilter->DerivePropertyConstraint()` is failing. What was the seg fault?

libgpopt/src/translate/CTranslatorExprToDXL.cpp
ashuka24: could we add this as a comment?

libgpopt/src/translate/CTranslatorExprToDXLUtils.cpp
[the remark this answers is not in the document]
sambitesh: The unit tests works in a way that a single `COptCtxt` spans across multiple test cases. The leaves behind incomplete `CExpressions` in the array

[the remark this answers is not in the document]
chrishajas: I'm concerned that we'll hit the retail assert in real queries since I'm not sure if all properties have been derived at this point. If you want you can make sure that they're derived by using `PdpDerive()`, but that's a pretty heavy hammer. If it's a unit test issue it's probably better to fix the unit test.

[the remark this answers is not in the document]
sambitesh: I don't think we will. As I said, the way the unit tests runs, it reuses a single`COptCtxt` for multiple queries. So our filter array is left with CExpression with some children released and some not. That's what gives us segfault. 

We need to fix the unit test but that should be a separate chore.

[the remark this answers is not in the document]
hardikar: I agree with Chris here. We should use `pexprFilter->DerivePropertyConstraint()` & then fix the unit test. If you don't want to fix the unittest now, the next option is using `PdpDerive()` and then access `GetDrvdPropRelational()`. That RTL_ASSERT risk can be troublesome down the line.

PR #582, opened 2020-04-13
Do not push Volatile funcs below aggs

sambitesh:
Consider the scenario below

```
create table tenk1 (c1 int, ten int);
create temp sequence ts1;
explain select * from (select distinct ten from tenk1) ss where ten < 10 + nextval('ts1') order by 1;
```

The filter outside the subquery is a candidate to be pushed below the
'distinct' in the sub-query.  But since 'nextval' is a volatile function, we
should not push it.


libgpopt/src/operators/CNormalizer.cpp
```diff
@@ -1154,7 +1163,9 @@ CNormalizer::PushThru
 	{
 		CExpression *pexprConj = (*pdrgpexprConjuncts)[ul];
 		pexprConj->AddRef();
-		if (FPushable(pexprLogical, pexprConj))
+
```
vraghavan78, line 1166: can u add a comment why this is incorrect.

vraghavan78: @hardikar thinks we need to push this logic to fpushable

sambitesh: `fpushable` is also used by joins. Volatile function filters are okay to be join filters.

hardikar: Ok, so based on our discussion during standup, for joins we're OK with pushing down volatile functions, because otherwise we'd a cross join. Somehow I'm still a bit uncomfortable with that. Anyway focussing on GROUP BY case. What happens for this query: `explain select * from (select distinct ten from tenk1) ss where ten < 10 + nextval('ts1') and ten > 0 order by 1;` Is it OK to push `ten > 0` but not `ten < 10 + nextval('ts1')`?

hardikar: Just checked with postgres:10:
```
postgres=# explain select * from (select distinct ten from tenk1) ss where ten < 10 + nextval('ts1') and ten > 0 order by 1;
                                QUERY PLAN
--------------------------------------------------------------------------
 Sort  (cost=47.57..47.74 rows=66 width=4)
   Sort Key: ss.ten
   ->  Subquery Scan on ss  (cost=40.13..45.58 rows=66 width=4)
         Filter: (ss.ten < (10 + nextval('ts1'::regclass)))
         ->  HashAggregate  (cost=40.13..42.11 rows=198 width=4)
               Group Key: tenk1.ten
               ->  Seq Scan on tenk1  (cost=0.00..38.25 rows=753 width=4)
                     Filter: (ten > 0)
(8 rows)
```

hardikar: Also, postgres:10 has no problem pushing the volatile function down in case of joins:
```
postgres=# explain select * from tenk1, tenk2 where tenk1.c1 = tenk2.c1 and tenk1.ten < 10 + nextval('ts1');
                             QUERY PLAN
---------------------------------------------------------------------
 Merge Join  (cost=244.04..375.44 rows=8509 width=16)
   Merge Cond: (tenk1.c1 = tenk2.c1)
   ->  Sort  (cost=85.53..87.41 rows=753 width=8)
         Sort Key: tenk1.c1
         ->  Seq Scan on tenk1  (cost=0.00..49.55 rows=753 width=8)
               Filter: (ten < (10 + nextval('ts1'::regclass)))
   ->  Sort  (cost=158.51..164.16 rows=2260 width=8)
         Sort Key: tenk2.c1
         ->  Seq Scan on tenk2  (cost=0.00..32.60 rows=2260 width=8)
(9 rows)
```

hardikar: FWIW same behavior on postgres:12.

hardikar:
I still feel we can move this logic in FPushable. It already checks for `!CUtils::FHasCTEAnchor(pexprPred);`. So just check if pexprLogical is a COperator::EopLogicalGbAgg?

PR #593, opened 2020-06-16
Adding CONTRIBUTING.md file

zellerh:
Moved some of the contents of the README file that apply to a new file, to be more consistent with other open source projects.


zellerh:
Is it ok if we don't bump any version, given that there is absolutely no change in behavior?

chrishajas:
I think the pipeline expects every commit to have an associated version.

zellerh:
Ok, thanks for reminding me and sorry I forgot since you told me last time. Will bump the version.

PR #598, opened 2020-07-10
Support aarch64 platform

japinli:
On aarch64, the ASM instructions are different from x86_64, so I add a GPOS_aarch64 macro to build gporca on aarch64.


chrishajas:
Hi @japinli  , we've now merged this to the orca repo, thanks for your contribution ! I saw that you PR'd (and we merged) a commit fixing compilation on master (https://github.com/greenplum-db/gpdb/pull/10482), but it isn't in 6X. If you'd like it also in gpdb 6X, please create a backport PR and we'll get that in too.

d:
Thanks for contributing to ORCA @japinli.
That said, IMHO this patch is going in the wrong direction. Instead of piling on with more inline asm, we should just use a widely available intrinsic like `__builtin_frame_address`. I'd be more in favor of backporting https://github.com/greenplum-db/gpdb/pull/10482 than this patch

japinli:
@chrishajas Thanks for reviewing. I think @d is right, we should use a widely available intrinsic. However, if you think this patch can be applied on 6X,  I am glad to create a backport PR on 6X-stable branch.

chrishajas:
Ah yes, that'd be ideal. I agree we should use https://github.com/greenplum-db/gpdb/pull/10482 for 6X rather than this PR.

PR #599, opened 2020-07-15
Fixes for optimizer_join_order exhaustive2

zellerh:
This PR contains various fixes and refactors, mostly related to the new join enumeration method "exhaustive2".

For more details, see the individual commits. Here is a high-level summary:

* Fixes in the decorrelator to handle nary LOJs (CLogicalNAryJoin operators that contain at least one LOJ). These fixes are in the first three commits.
* Fix in the CJoinStatProcessor for nary LOJs.
* Fix for removing implied predicates from nary LOJs.
* Refactoring the way we cache scalar expressions in MEMO groups (needed for the next item).
* Fixes for handling nary LOJs in transforms, particularly `CXformSubqNAryJoin2Apply`.
* Refactor to replace the use of function pointers with a switch statement.
* Fix the DbgPrint method for CColRefSet and friends that I broke in a previous PR.

I would recommend reviewing the commits individually.

The corresponding GPDB PR is https://github.com/greenplum-db/gpdb/pull/10512.


libgpopt/include/gpopt/operators/CExpressionHandle.h
```diff
@@ -271,11 +271,17 @@ namespace gpopt
 			// check whether an expression's children have a volatile function
 			BOOL FChildrenHaveVolatileFuncScan();
 
-			// return the scalar child at given index
-			CExpression *PexprScalarChild(ULONG child_index) const;
+			// return a representative (inexact) scalar child at given index
```
ashuka24, line 274: could you add here that a representative scalar child is the scalar child where any existing subqueries have been converted into `CScalarConst` or NULL

libgpopt/include/gpopt/search/CGroup.h
```diff
@@ -202,8 +202,11 @@ namespace gpopt
 			// group stats
 			IStatistics *m_pstats;
 
-			// scalar expression for stat derivation
-			CExpression *m_pexprScalar;
+			// scalar expression for stat derivation (subqueries substituted with a dummy)
+			CExpression *m_pexprScalarRep;
+
+			// scalar expression above is an exact version of the scalar expr in the group
+			BOOL m_pexprScalarRepIsExact;
```
ashuka24, line 209: maybe add a comment that this is true if and only if the representative scalar expr is the same as the original incoming scalar expression. (or maybe iterate here that it means the scalar expression has no subqueries)

libgpopt/src/xforms/CDecorrelator.cpp
```diff
@@ -704,14 +783,14 @@ CDecorrelator::FProcessProject
...
-		// 2. if the relational child of LogicalSequenceProject node does not have any aggregate window function
+		// 2. if the scalar child of LogicalSequenceProject node does not have any aggregate window function
 
 		// if the project list contains aggregrate on window function, then
 		// we can decorrelate it as the aggregate is performed over a column or count(*).
 		// The IN condition will be translated to a join instead of a correlated plan.
 		// ex: select C.j from C where C.i in (select avg(i) over (partition by B.i) from B where B.i=C.i);
 		// ===> (resulting join condition) b.i = c.i and c.i = avg(i)
-		if (CLogicalSequenceProject::PopConvert(pexpr->Pop())->FHasLocalOuterRefs(exprhdl)
+		if (CLogicalSequenceProject::PopConvert(pexpr->Pop())->FHasLocalReferencesTo(outerRefsToRemove)
```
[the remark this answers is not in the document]
zellerh: The check on line 771 tests whether the project list, which is a child expression, has any outer refs. The check on line 793 tests whether the operator itself (the order by, partition by and other window-spec items stored in the operator) has outer refs.

libgpopt/src/base/CUtils.cpp
```diff
@@ -1904,6 +1910,58 @@ CUtils::PexprScalarConstOid
...
+CUtils::PexprScalarConstNull(CMemoryPool *mp, const IMDType *typ, INT type_modifier)
+{
+	IDatum *datum = NULL;
+	IMDId *mdid = typ->MDId();
+	mdid->AddRef();
+
+	switch (typ->GetDatumType())
+	{
+		case IMDType::EtiInt2:
+			datum = GPOS_NEW(mp) CDatumInt2GPDB(mdid, 0, true);
```
ashuka24, line 1924: It looks like in other locations we create IDatum's using something like `pmdtype->CreateInt2Datum(mp, 0, true)` method rather than bringing in the CDatum class. Should we make this consistent here?

libgpopt/src/operators/CExpressionHandle.cpp
```diff
@@ -1468,13 +1468,46 @@ CExpressionHandle::PexprScalar() const
...
+	if (NULL != m_pexpr &&
+		NULL != (*m_pexpr)[child_index]->Pgexpr() &&
+		!((*m_pexpr)[child_index]->Pgexpr()->Pgroup()->FScalarRepIsExact()))
+	{
+		// the expression does not come from a group, but its child does and
+		// the child group does not have an exact expression
+		return NULL;
+	}
+
+	return PexprScalarRepChild(child_index);
```
ashuka24, line 1496: is there a way to add an assert here to confirm that the return  that the `PexprScalarRepChild` will always be the exact child?

libgpopt/src/operators/CExpressionHandle.cpp
```diff
@@ -1468,13 +1468,46 @@ CExpressionHandle::PexprScalar() const
...
+CExpression *
+CExpressionHandle::PexprScalarExact() const
+{
+	if (NULL != m_pgexpr && !m_pgexpr->Pgroup()->FScalarRepIsExact())
+	{
+		return NULL;
+	}
+
+	return PexprScalarRep();
+}
```
ashuka24, line 1509: same question as above

libgpopt/src/operators/CLogicalLimit.cpp
```diff
@@ -223,8 +223,9 @@ CLogicalLimit::DeriveMaxCard
 	)
 	const
 {
-	CExpression *pexprCount = exprhdl.PexprScalarChild(2 /*child_index*/);
-	if (CUtils::FScalarConstInt<IMDTypeInt8>(pexprCount))
+	CExpression *pexprCount = exprhdl.PexprScalarExactChild(2 /*child_index*/);
```
ashuka24, line 226: can you add a comment why we need the exact child here and cannot just make do with the rep child?

libgpopt/src/operators/CLogicalNAryJoin.cpp
```diff
@@ -87,7 +87,7 @@ CLogicalNAryJoin::DeriveMaxCard
 		return CMaxCard(0 /*ull*/);
 	}
 
-	CExpression *pexprScalar = exprhdl.PexprScalarChild(arity-1);
+	CExpression *pexprScalar = exprhdl.PexprScalarExactChild(arity-1);
```
ashuka24, line 90: same question as above, can you make a comment why we need the exact child here and cannot use the rep child?

libgpopt/src/base/CUtils.cpp
zellerh: Done.

libgpopt/src/operators/CLogicalLimit.cpp
zellerh: Sure. It would probably work either way, since the following check also excludes any expressions with subqueries.

libgpopt/src/operators/CLogicalNAryJoin.cpp
zellerh: Updated the comment in the if check below, explaining why we have to return NULL if we can't get an exact expression.

libgpopt/src/operators/CExpressionHandle.cpp
zellerh: Take that back, when we deal with a CExpression instead of a CGroupExpression, the exact child may contain a subquery...

libnaucrates/src/statistics/CJoinStatsProcessor.cpp
```diff
@@ -559,18 +559,9 @@ CJoinStatsProcessor::DeriveJoinStats
 		statistics_array->Append(child_stats);
 	}
 
-	CExpression *join_pred_expr = exprhdl.PexprScalarExactChild(arity - 1);
+	CExpression *join_pred_expr = exprhdl.PexprScalarRepChild(arity - 1);
 
-	if (NULL == join_pred_expr)
-	{
-		// in case of subquery in join predicate, assume join condition is True
-		join_pred_expr = CUtils::PexprScalarConstBool(mp, true /*value*/);
```
chrishajas, line 566: Do we have any other tests that show the cardinality difference?

libgpopt/src/operators/CLogicalInnerJoin.cpp
```diff
@@ -137,8 +137,8 @@ CLogicalInnerJoin::FFewerConj
 	GPOS_ASSERT(pgroupScalarFst->FScalar());
 	GPOS_ASSERT(pgroupScalarSnd->FScalar());
 
-	CExpressionArray *pdrgpexprConjFst = CPredicateUtils::PdrgpexprConjuncts(mp, pgroupScalarFst->PexprScalar());
-	CExpressionArray *pdrgpexprConjSnd = CPredicateUtils::PdrgpexprConjuncts(mp, pgroupScalarSnd->PexprScalar());
+	CExpressionArray *pdrgpexprConjFst = CPredicateUtils::PdrgpexprConjuncts(mp, pgroupScalarFst->PexprScalarRep());
```
chrishajas, line 140: Is this not exact because we're just getting the size?

libgpopt/src/operators/CExpressionHandle.cpp
```diff
@@ -1399,15 +1399,14 @@ CExpressionHandle::DeriveProducerStats
...
 
 //---------------------------------------------------------------------------
-//	@function:
-//		CExpressionHandle::PexprScalarChild
+// CExpressionHandle::PexprScalarRepChild
 //
-//	@doc:
-//		Get the scalar child at given index
+// Get a representative (inexact) scalar child at given index. Subqueries
+// in the child are replaced by a TRUE or NULL constant.
```
chrishajas, line 1405: Can we also list where we should be using representative vs exact scalars, I think they're mentioned in one location, but it might be good to have it in each of the PexprScalar* functions

libgpopt/src/xforms/CJoinOrderDPv2.cpp
```diff
@@ -652,39 +655,47 @@ CJoinOrderDPv2::AddExprToGroupIfNecessary
 void
 CJoinOrderDPv2::PopulateExpressionToEdgeMapIfNeeded()
```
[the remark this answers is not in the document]
zellerh: Done.

libnaucrates/src/statistics/CJoinStatsProcessor.cpp
zellerh: Yes, as far as I remember this caused some of the cardinality diffs seen in the MDPs. Sorry, I didn't note down exactly which diffs these were. Note that the cardinality changes can go both ways.

a = 10 and b in (subquery): This used to be interpreted as "true", now interpreted as "a = 10".

a = 10 or b in (subquery): This used to be interpreted as "true", now interpreted as "a = 10".

a = (scalar subquery): This used to be interpreted as "true", now interpreted as a = NULL ("false").

zellerh:
Forgot to reply to one more of Chris's suggestions:

> Should we also enable exhaustive2 on the master branch with this so it's a first class citizen and we get more test coverage?

Sounds good to me. Enabling it for 5X and 6X might have to happen later, but when I port this to master I'll make it the default and mention it in the PR as an additional change to the forward port.

libgpopt/src/operators/CExpressionHandle.cpp
ashuka24: I think I may be a bit confused. Does this mean that it is possible here to treat a representation of the child (which contains no subqueries) as the exact child if we have a CExpression instead of a CGroupExpression?

zellerh: Hope I understand the question correctly. If so, I think the answer is yes. But, things may be more complicated yet.  Looking at this some more, we may have to consider three cases:

* A CGroupExpression, which may or may not be exact
* A CExpression tree (no node attached to a CGroupExpression), which will always be exact, even with subqueries
* A CExpression tree that was extracted from a CBinding for an xform, where a CExpression will point to a CGroupExpression (the CExpression may be a CPatternLeaf). In that case, the method will return not the "leaf" node, but the expression from the group, which might again be inexact.

But for this particular line, we simply check in the code above whether we have an exact expression. If not, we return null. Therefore, when we reach here, we can guarantee that `PexprScalarRepChild` will return an exact expression.

libgpopt/src/xforms/CDecorrelator.cpp
```diff
@@ -489,9 +510,59 @@ CDecorrelator::FProcessJoin
...
+			if (isNaryLOJ)
+			{
+				// we need to fish our ON clause out of the scalar argument of the NAry join
+				onPred = naryJoin->GetOnPredicateForLOJChild(pexpr, ul);
+			}
+
+			// the ON predicate of the LOJ can refer to any columns produced by
+			// (decorrelated) children we already visited or by the right child of the LOJ
+			CColRefSet *availableColsForOnPred = GPOS_NEW(mp) CColRefSet(mp, *pcrsOutput);
+			availableColsForOnPred->Union((*pexpr)[ul]->DeriveOutputColumns());
```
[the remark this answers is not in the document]
zellerh: Would you agree with the comment 4 lines above? This should check whether the ON predicate refers only to columns that come from the left or right child, not any others. Actually, the next commit ea3b84f85c7bfb45fec77458d5fa490b50a1a3bc changes this code again and checks the opposite, whether the ON predicate refers to any of the outer references to be removed.

libgpopt/src/xforms/CXformUtils.cpp
```diff
@@ -118,7 +118,13 @@ CXformUtils::ExfpExpandJoinOrder
 	CExpressionHandle &exprhdl
 	)
 {
-	if (exprhdl.DeriveHasSubquery(exprhdl.Arity() - 1) || exprhdl.HasOuterRefs())
+	// Note that there is a potential problem here: If we have outer refs that
+	// can't be removed by the decorrelator, then we should expand this join
+	// despite the outer refs. This is something we could investigate. For now,
```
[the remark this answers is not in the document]
zellerh: The reason is that in exhaustive, we will get a `CLogicalLeftOuterJoin` and that operator will be implemented, even if it has outer refs. We need to make sure we also allow an NAry join with outer refs to be implemented in exhaustive2. This comment got changed in another commit in this PR to the following:

```
	// With optimizer_join_order set to 'query' or 'exhaustive', the
	// 'query' join order will expand the join even if it contains
	// outer refs, using another method to get the promise.
	// Therefore we also allow expansion for 'exhaustive2'
	// when we have outer refs.
```

Is this better?

libgpopt/src/operators/CPredicateUtils.cpp
```diff
@@ -2416,6 +2416,23 @@ CPredicateUtils::PexprRemoveImpliedConjuncts
...
+		// for a ScalarNAryJoinPredList we have to preserve that operator and
+		// remove implied preds from each child individually
+		CExpressionArray *newChildren = GPOS_NEW(mp) CExpressionArray(mp);
+
+		for (ULONG c=0; c<pexprScalar->Arity(); c++)
+		{
+			newChildren->Append(PexprRemoveImpliedConjuncts(mp, (*pexprScalar)[c], exprhdl));
+		}
+
+		// reassemble the CScalarNAryJoinPredList with its new children without outer refs
```
zellerh, line 2430: Fix comment

libgpopt/src/operators/CPredicateUtils.cpp
```diff
@@ -2200,6 +2200,34 @@ CPredicateUtils::SeparateOuterRefs
...
+			CExpression *childOuterRefExpr = NULL;
+
+			SeparateOuterRefs(mp, (*pexprScalar)[c], outer_refs, &childLocalExpr, &childOuterRefExpr);
+			localChildren->Append(childLocalExpr);
+			outerRefChildren->Append(childOuterRefExpr);
+		}
+
+		// reassemble the CScalarNAryJoinPredList with its new children without outer refs
+		pexprScalar->Pop()->AddRef();
+		*ppexprLocal = GPOS_NEW(mp) CExpression(mp, pexprScalar->Pop(), localChildren);
```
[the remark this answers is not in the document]
zellerh: We are not actually mixing inner and outer joins preds. This creates two `CLogicalNAryJoinPredicates` trees, one with the regular predicates and one with the outer refs. Both preserve the information what's an inner and an outer join when we call `CJoinStatsProcessor::CalcAllJoinStats`. Also, I ran an example query and the TPC-DS queries and did not see any difference in cardinality estimates between exhaustive and exhaustive2. Here is the example query:

```
drop table if exists foo, bar;

create table foo(a int, b int);
create table bar(a int, b int);

insert into foo select i, i from generate_series(1,100000) i;
insert into bar select i, i from generate_series(1,100000) i;

set optimizer_join_order to exhaustive2;

explain
select *
from foo out where exists (select * from foo i1 join bar i2 on i1.a=i2.a left outer join bar i3 on i1.a=i3.a and i3.b=out.b);
```

libgpopt/src/xforms/CXformSubqJoin2Apply.cpp
```diff
@@ -206,15 +206,30 @@ CXformSubqJoin2Apply::PexprSubqueryPushDown
...
+			// inner join child
+			pcrsOutput = pexprChild->DeriveOutputColumns();
+			pcrsOutput->AddRef();
+		}
+		else
+		{
+			// use an empty set for right children of LOJs, because we don't want to
+			// push any subqueries down to those children (note that non-correlated
+			// subqueries will be pushed to the leftmost child, which is never the
+			// right child of an LOJ)
```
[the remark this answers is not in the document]
zellerh: Good question. I don't have a complete answer. I think the reason we stay away from LOJ children so far is that in the existing logic for binary LOJs, we don't have an xform that pushes subqueries down. It might be possible to do this in at least some cases, given that the xform just pushes down a project that computes the subquery result, not any of the predicates. But, that would be outside the scope of this PR.

libgpopt/src/operators/CLogicalNAryJoin.cpp
```diff
@@ -212,6 +212,51 @@ CLogicalNAryJoin::GetTrueInnerJoinPreds(CMemoryPool *mp, CExpressionHandle &expr
...
+			new_children->Append(existing_child);
+		}
+
+		pop->AddRef();
+
+		return GPOS_NEW(mp) CExpression(mp, pop, new_children);
+	}
+
+	// with all inner joins it's a total replacement, just return the inner join preds
+	// (caller should have passed us a ref count which they now get back from us)
```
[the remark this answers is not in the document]
zellerh: Yes, I can add some asserts that the scalar child and the array of child pred indexes are consistent.

libgpos/include/gpos/common/OsPrintTrait.h
```diff
@@ -0,0 +1,90 @@
...
+			// the following occurs:
+			// 1. T doesn't have a non-static member OsPrint
+			// 2. said member is not a function that matches the signature we expect
+			//
+			// N.B. for exposition purposes this requirement is overly strict to achieve
+			// simplicity: we _really_ only care about contravariant parameter types and
+			// covariant return types. In plain English, the parameter types can be broader,
+			// and the return type can be narrower. The C++14 version achieves this.
+			template <class E>
+			struct COPImpl<E, val_type<0 * sizeof(static_cast<IOstream& (E::*)(IOstream&) const>(&E::OsPrint))> >
```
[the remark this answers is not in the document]
zellerh: I couldn't figure that out, sorry. I tried to simplify the code and wasn't able to - didn't get the desired behavior or got a compile error. @d would have the answer.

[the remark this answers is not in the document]
d: @dgkimura the zero multiplication is to simplify it to always be `val_type<0>` when the member existence check passes. We don't actually want to know the size of the member pointer (plus it's actually implementation dependent), just want to know it's not SFINAE'd away. This is similar to how in C++14 you would do a `decltype(void(... actual expression to check for SFINAE...))` where the `void` simplifies the type.

PR #605, opened 2020-09-16
Fall back for selects of outer refs in scalar subquery in ORCA

zellerh:
This is a backport of PR https://github.com/greenplum-db/gpdb/pull/10799.

When we have an outer ref in a subquery, like this

select * from foo where foo.a is null or foo.a = (select foo.b from bar)

then we can't simply use the outer reference for the condition when
we unnest the subquery into an apply. This is because if the subquery
returns no rows, then we must be using a NULL instead of the outer
reference.

We have code to handle this for quantified subqueries, but not for
scalar subqueries.

When we translate the generated DXL to a plan, we assert when we find
an outer reference in the project list of a subquery. In rare cases,
we might also crash, when the subquery contained a project with multiple
values below the outer reference (see added test in gporca.sql).

The "fix" (more a workaround) is to force a fallback when we detect
this situation during unnesting of a scalar subquery.


chrishajas (approved):

libgpopt/src/xforms/CSubqueryHandler.cpp
```diff
@@ -393,7 +400,20 @@ CSubqueryHandler::FRemoveScalarSubquery
...
+	SSubqueryDesc *psd = Psd(pmp, pexprSubquery, pexprOuter, pcrSubquery, esqctxt);
+
+	if (psd->m_fReturnedPcrIsOuterRef)
+	{
+		// The subquery returns an outer reference. We can't simply replace the subquery with that
+		// expression, because we would miss the case where the subquery is an empty table and we
+		// would have to substitute the outer ref with a NULL.
+		// We could use a dummy expression from the subquery to perform a check, but for now we'll
+		// just give up.
+		// Example: select * from foo where foo.a = (select foo.b from bar);
```
hardikar, line 412: Change this comment as well?

hardikar (approved):

zellerh:
Holding off with this PR, since we will probably include another change: https://github.com/greenplum-db/gpdb/pull/10863, which allows us to handle some of these queries in ORCA.

chrishajas (approved):

PR #606, opened 2020-10-08
Disallow physical index scans in Orca to have predicate condition with subquery

chrishajas:
This is a backport of the first commit of https://github.com/greenplum-db/gpdb/pull/10944.
The second commit in the above PR didn't trivially apply cleanly and isn't necessary to prevent a crash.
If we think that commit should also be backported to 5X I can do that.

Physical operators should not have subqueries, by this point they should
have been converted into correlated joins or expressions. While we would
consider these invalid plans this wouldn't ever be executed, we assumed
in one case that we could retrieve an exact expression from a group that
contained a subquery when deriving the distribution spec, which caused a
crash.

Now, we'll be more defensive and disallow creating physical index scans
if the predicate contains a subquery. This isn't a problem for generic
gets, as they don't have a scalar predicate.


zellerh (approved):
LGTM.

PR #608, opened 2020-11-12
Fix handling of empty constraint array in Orca (#11139)

chrishajas:
During preprocessing, Orca simplifies the query by
merging/deduplicating constraints. However, Orca did not consider
the case where the passed in constraint was a column compared with
an empty array. This assumption caused Orca to crash when dealing
with predicates such as `a = ANY('{}')`.

Instead, we now explicitly return an empty constraint when dealing
with an ANY( '{}') clause. For ALL('{}'), we won't process the
constraint as there's no simplification to do.


zellerh (approved):
LGTM.

PR #609, opened 2020-11-16
Enforce formatting in CI

d:
Similar to greenplum-db/gpdb#10921 and greenplum-db/gpdb#10926, this PR adds format checking to GPORCA main pipeline and its PR pipeline

### DIFF
--- a/concourse/check_format.yml
+++ b/concourse/check_format.yml
@@ -1,0 +1,14 @@
+platform: linux
+image_resource:
+  type: registry-image
+  source:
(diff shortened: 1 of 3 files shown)


chrishajas (approved):
LGTM, also remember to bump the orca patch version.

d:
> LGTM, also remember to bump the orca patch version.

Thanks for the review! Yes I bumped the patch version, pushed, and `fly set-pipeline`d both pipelines.

PR #610, opened 2020-12-15
Infer predicates from constraints on outer refs (#11175)

zellerh:
* Pass outer ref constraints to several methods

To generate inferred predicates for outer refs, we need to pass the
constraints for outer refs to methods
and CExpressionPreprocessor::PexprFromConstraints(),
CExpressionPreprocessor::PexprFromConstraintsScalar(), and
CExpressionPreprocessor::PexprScalarPredicates().

Initially NULL, the new argument is set to the constraints of the
current expression when we enter a subquery through a call to
CExpressionPreprocessor::PexprFromConstraintsScalar(). Please
note that this commit has a limitation, it does not handle skip-level
outer refs.

* Infer predicates from constraints on outer refs

When generating inferred predicates, combine the local equivalence class
for the colref with that of the outer ref at the outer scope. With this
combined equivalence class, search both the current expression's constraints
as those of the outer refs for predicates to infer.

This will help queries such as

select *
from   foo
where  foo.a in (select max(bar.a) from bar where foo.b=bar.b)
   and foo.b < 2;

We can now infer a predicate bar.b < 2 inside the subquery. The outer scope
contributes the constraint foo.b < 2 and the inner scope contributes the
equivalence class { foo.b, bar.b }.

(cherry picked from commit 7df2d57e1ffae3ce80119e2eac0607aa842ba5ba)


hardikar (approved):
Only took a brief look; I assume it ported cleanly.

PR #612, opened 2021-01-13
5X: Ban plans that demand a rescan of a dynamic PartitionSelector

soumyadeep2007:
We should not generate a plan that involves a rescan of a
PartitionSelector for DPE(Dynamic Partition Elimination).

Generating such a plan => the assertion in ExecRescanPartitionSelector()
is tripped:

```c
	/* If the PartitionSelector is in the inner side of a nest loop join,
	 * it should be constant partition elimination and thus has no child node.*/
	PlanState  *outerPlan = outerPlanState(node);
	Assert (NULL == outerPlan);
```

Without asserts, we will trip:

ERROR:  no hash_seq_search scan for hash table "Dynamic Table Scan Pid Index"

This commit ensures that such a plan is NOT generated. For example,
before this commit, a plan such as the following is possible.

```sql
repro=# explain select 1 from dpe_bar join (select * from dpe_part join dpe_foo on i = a) t on i + a < c;
                                                        QUERY PLAN
--------------------------------------------------------------------------------------------------------------------------
 Result  (cost=0.00..1724.01 rows=3 width=4)
   ->  Gather Motion 3:1  (slice2; segments: 3)  (cost=0.00..1724.01 rows=8 width=1)
         ->  Nested Loop  (cost=0.00..1724.01 rows=3 width=1)
               Join Filter: (dpe_part.i + dpe_foo.a) < dpe_bar.c
               ->  Broadcast Motion 3:3  (slice1; segments: 3)  (cost=0.00..431.00 rows=2 width=4)
                     ->  Table Scan on dpe_bar  (cost=0.00..431.00 rows=1 width=4)
               ->  Hash Join  (cost=0.00..862.01 rows=3 width=8)
                     Hash Cond: dpe_part.i = dpe_foo.a
                     ->  Dynamic Table Scan on dpe_part (dynamic scan id: 1)  (cost=0.00..431.00 rows=27 width=4)
                     ->  Hash  (cost=100.00..100.00 rows=34 width=4)
                           ->  Partition Selector for dpe_part (dynamic scan id: 1)  (cost=10.00..100.00 rows=34 width=4)
                                 ->  Table Scan on dpe_foo  (cost=0.00..431.00 rows=3 width=4)
 Settings:  optimizer_nestloop_factor=1
```

After this commit, since DPE PartitionSelectors are no longer
rewindable, the plan generated will now have a Spool on top of the
HashJoin or will have the PartitionSelector moved to the left hand side
of the NestedLoopJoin, among other alternatives.

[...]

PR #613, opened 2021-01-27
[Orca] sort includes

d:
I find myself doing it so often that it's time to upstream this: let
clang-format sort and regroup our #include blocks.

We actually tried doing this during the original #10357 but one
logistical hurdle of trying to sort #includes was that our headers are
not exactly order-independent, hence it retreated to bing a TODO
comment.

About 70% of this PR is automated by just running clang-format with the
new configuration, and 30% was manually fixing any order dependency that
the sorting surfaces.

This is manually remade, but still marking as a spiritual cherry-pick of
commit f81aaafd25206b8c.
(cherry picked from commit f81aaafd25206b8c52205835ba19f6bcb0a66297)


dgkimura (approved):

d:
Merged as part of #612

PR #617, opened 2021-03-26
Disallow subquery in predicate of implementation xforms

chrishajas:
This PR is for 5X, but will be forward-ported to 6X/master after review. We haven't been able to repro the issue on 6X/master, but it's still possible on those branches.

This generates an invalid plan for most cases. While we had this logic
in some xforms, we missed a few cases.

This manifested as a crash in some cases as we assumed subqueries had
been handled by time we get to optimization stage.

Additionally, add error handling to PexprScalarExactChild()

This function is only called in places where the exact child must exist,
and returns null otherwise. 

However, there have been some cases where we assume the child must
exist, but due to another bug the child does not and we don't handle the
null. This is a defensive change; if we get into this situation we'll
throw an exception and fall back to planner instead of crashing.


libgpopt/src/xforms/CXformImplementBitmapTableGet.cpp
```diff
@@ -43,6 +43,19 @@ CXformImplementBitmapTableGet::CXformImplementBitmapTableGet(CMemoryPool *mp)
 {
 }
 
+CXform::EXformPromise
+CXformImplementBitmapTableGet::Exfp(CExpressionHandle &exprhdl) const
+{
+	if (exprhdl.DeriveHasSubquery(0) || exprhdl.DeriveHasSubquery(1))
+	{
+		// xform is inapplicable if scalar is a subquery
```
d, line 51: This comment doesn't look correct, aren't you checking _both_ the scalar bitmap index condition _and_ the relational child?

libgpopt/src/xforms/CXformImplementFullOuterMergeJoin.cpp
```diff
@@ -31,9 +31,13 @@ CXformImplementFullOuterMergeJoin::CXformImplementFullOuterMergeJoin(
 }
 
 CXform::EXformPromise
-CXformImplementFullOuterMergeJoin::Exfp(CExpressionHandle &	 //exprhdl
-) const
+CXformImplementFullOuterMergeJoin::Exfp(CExpressionHandle &exprhdl) const
 {
+	if (exprhdl.DeriveHasSubquery(exprhdl.Arity() - 1))
```
d, line 36: Hmm, this check suggests that only the RHS matters in a `FULL JOIN`, but is that true? Aren't we equally busted if the LHS of the full join has a residual subquery?

libgpopt/src/xforms/CXformImplementLimit.cpp
```diff
@@ -42,6 +42,15 @@ CXformImplementLimit::CXformImplementLimit(CMemoryPool *mp)
 {
 }
 
+CXform::EXformPromise
+CXformImplementLimit::Exfp(CExpressionHandle &exprhdl) const
+{
+	if (exprhdl.DeriveHasSubquery(1) || exprhdl.DeriveHasSubquery(2))
```
d, line 48: Again, the first child (scalar count) _can_ be a subquery, no? I haven't tried, TBH, but from a SQL perspective it can be a subquery, like

```
SELECT a FROM grandma LIMIT (SELECT max(c) FROM bar) OFFSET (SELECT min(d) FROM bar);
```

chrishajas: It can from a SQL perspective, but Orca doesn't handle subqueries in the limit/offset currently so I thought it's best to prevent them here.

libgpopt/src/xforms/CXformImplementFullOuterMergeJoin.cpp
chrishajas: This is checking the scalar child

libgpopt/src/xforms/CXformImplementBitmapTableGet.cpp
chrishajas: I'm just planning to remove the comment, since it seems unnecessary (the above lines show the children that we're checking).

chrishajas:
Talked with @d a bit offline, if I understood correctly, we should be banning implementation xforms with scalar children. I think I found all the relevant xforms by grepping for `CXformImplementation(GPOS_NEW`.

libgpopt/include/gpopt/xforms/CXformImplementFullOuterMergeJoin.h
```diff
@@ -7,13 +7,13 @@
...
 
-#include "gpopt/xforms/CXformExploration.h"
+#include "gpopt/xforms/CXformImplementation.h"
 
 namespace gpopt
 {
 using namespace gpos;
 
-class CXformImplementFullOuterMergeJoin : public CXformExploration
+class CXformImplementFullOuterMergeJoin : public CXformImplementation
```
zellerh, line 16: Nice catch :-)
Interesting, we must have done the FOJ implementation in the exploration phase until now. Looks like doing that doesn't cause any problems.

zellerh (approved):
LGTM, thank you for fixing this problem! I took a look at the implementation xforms. There is one that doesn't yet have the subquery checks: `CXformImplementSplit`. Do you want to add that one as well?

chrishajas:
> LGTM, thank you for fixing this problem! I took a look at the implementation xforms. There is one that doesn't yet have the subquery checks: `CXformImplementSplit`. Do you want to add that one as well?

Thanks, added! The newline threw off my grep.

d (changes requested):
Thanks for picking this up! I took a quick look and here's some feedback:

## Stylistic nitpick
The following pattern occurred 3 times:

```c++
	if (exprhdl.DeriveHasSubquery(exprhdl.Arity() - 1))
	{
		return CXform::ExfpNone;
	}
```

I find that a little hard to read. Given all three instances are for binary joins, I'd recommend substituting a constant 2 (bonus point for giving it a name) for `exprhdl.Arity() - 1`.

## Implementation transforms that are missed in this PR / deliberately not changed:

1. `CXformDynamicIndexGet2DynamicIndexScan`
1. `CXformDynamicGet2DynamicTableScan`
1. `CXformExternalGet2ExternalScan`
1. `CXformGet2TableScan`
1. `CXformImplementCTEConsumer`
1. `CXformImplementCTEProducer`
1. `CXformImplementConstTableGet`
1. `CXformImplementDML`
1. `CXformImplementPartitionSelector`
1. `CXformImplementRowTrigger`
1. `CXformImplementSequence`
1. `CXformImplementUnionAll`

Most of the above (everything except for the first one) makes sense: in ORCA's canonical model these xform produce physical operators that don't have scalar sub groups. However the first one doesn't make that cut, surely a `CLogicalDynamicIndexGet` has a scalar condition?

Upon closer examination, it turns out that we get away with the current "always transform" implementation in `Exfp` for `CXformDynamicIndexGet2DynamicIndexScan` because we have the following code in the actual `Transform` method:

```c++
	if (pexprIndexCond->DeriveHasSubquery())
	{
		return;
	}
```

This is unfortunate, because it gives us more than one way of _not_ performing a transform:

1. returning `ExfpNone` when overriding `Exfp()`
1. by adding nothing into the `CXformResult` object during `Transform()` (i.e. assuming we returned `ExfpHigh` in `Exfp()`

Do you think there's an opportunity to move the sub query check into `Exfp`?


## P.S.

If you're wondering how I found these, the answer is clang-query:

```
match cxxMethodDecl(isDefinition(), hasName("Exfp"), ofClass(isDerivedFrom(cxxRecordDecl(hasName("CXformImplementation")))), unless(hasDescendant(callExpr(callee(functionDecl(hasName("DeriveHasSubquery"))))))).bind("Exfp")
```

d (approved):
The above should read "LGTM with a few minor feedback"

zellerh (approved):
Still LGTM. Thank you again for the fix!

libgpopt/include/gpopt/xforms/CXformImplementFullOuterMergeJoin.h
chrishajas: Surprisingly, this did cause a change. In `ImplementMergeJoin()`, we check if we've already computed join keys for the scalar child, and if so, we reuse those. If we derive the class from `CXformExploration`, we haven't computed the scalar keys yet and go through a different code path that checks if the merge join is compatible, this check fails (which is fine), and we don't generate a full join alternative. However, if we derive from `CXformImplementation`, we don't do that check, and generate the full join alternative that is lower cost, we choose it, it fails during ExprToDXL and we fall back to planner.

So there seems to be a bug that this exposes within merge join. I'm planning to leave this change out for now and we can address it separately.

zellerh: Maybe add a FIXME so we remember...

d:
LGTM, :shipit:

PR #618, opened 2021-05-04
Remove uploading of releases to Conan

chrishajas:
The JFrog Bintray service we were using to host ORCA release source
code was recently shut down. With our Bintray repository completely
wiped and the service ended, our Conan pull was unable to find the
ORCA release source code package. As a workaround, just pull the
release source code directly from the Github release page instead and
compile accordingly. Note that this only applies to our Ubuntu CI
platform because the CentOS and SLES compilation jobs pulls the
precompiled binary package (bin_orca_centos5_release.tar.gz) from the
Github release page instead of using Conan.

We therefore no longer can upload to conan and will consume the orca
source from the github release directly.

Corresponding 5X PR: https://github.com/greenplum-db/gpdb/pull/11906


zellerh (approved):
LGTM

PR #619, opened 2021-05-06
Pin clang-toolchain image to specific version

chrishajas:


scripts/fmt
```diff
@@ -28,7 +28,7 @@ set -e -u -o pipefail
 # against it.
 if ((BASH_VERSINFO[0] < 4)); then false; fi
 
-readonly INTENT=src/backend/gporca/clang-format.intent.yaml
+readonly INTENT=clang-format.intent.yaml
```
d, line 31: Ouch, good catch

d (approved):

PR #621, opened 2021-06-11
Migrate to gpc

dgkimura:
Followed migration steps as per gp-image-baking README [1].

[1] https://github.com/pivotal/gp-image-baking#migrating-from-dockerhub-to-gcr


concourse/clone_remote_repo.yml
```diff
@@ -1,6 +1,6 @@
 platform: linux
 image_resource:
-  type: docker-image
+  type: registry-image
   source:
     repository: yolo/orcadev
```
chrishajas, line 5: are we changing this to gcr at some point too?

chrishajas (approved):
Looks good, thanks!

chrishajas:
Oh also, needs an orca patch bump.

concourse/clone_remote_repo.yml
dgkimura: @khuddlefish had offline mentioned we could, but also that it's not necessary since it's not in pivotaldata account.  Because it wasn't necessary, I didn't do it.  But maybe it makes sense to do so that all our images are in one place in GCR.  I'm okay either way.  Thoughts?

chrishajas: Ah, I'm good with keeping it as-is then, especially since this is legacy and isn't pulled from often.

chrishajas (approved):

dgkimura:
Merged. Thanks for review!

PR #623, opened 2021-06-28
Fix crash in ORCA if root and leaf dropped column layouts mismatch

khannaekta:
ORCA for GPDB6 and older versions assumes that the root and leaf
partitions have the same underlying column structure. This assumption
can be broken when an exchange partition with or without same dropped
columns as root is inserted into the partition table. Further
complicating the matter is that ORCA always uses the root partition to
construct index metadata. If ORCA detects a mismatch in the index and
relation metadata, then it will not consider an index plan.

(backported from https://github.com/greenplum-db/gpdb/pull/12201)
GPDB PR that ports tests for 5X_STABLE: https://github.com/greenplum-db/gpdb/pull/12232


chrishajas (approved):

CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ project(gpopt LANGUAGES CXX C)
 set(CMAKE_CXX_STANDARD 98)
 set(GPORCA_VERSION_MAJOR 3)
 set(GPORCA_VERSION_MINOR 116)
-set(GPORCA_VERSION_PATCH 3)
+set(GPORCA_VERSION_PATCH 4)
```
chrishajas, line 9: Bump minor, not patch

PR #625, opened 2021-07-14
[Backport] Fix Cardinality estimation for not well defined histograms

l-wang:
Consider the below example:
```
EXPLAIN WITH  CTE AS
 (SELECT c::int8, b  FROM badestimate1)
SELECT 1 FROM CTE WHERE c <> 1 OR c = 1 AND b <> 101;
                                                      QUERY PLAN
----------------------------------------------------------------------------------------------------------------------
 Gather Motion 3:1  (slice1; segments: 3)  (cost=0.00..431.13 rows=2 width=4)
   ->  Result  (cost=0.00..431.13 rows=1 width=4)
         ->  Result  (cost=0.00..431.13 rows=1 width=1)
               Filter: ((c::bigint) <> 1 OR ((c::bigint) = 1 AND b <> 101)) AND ((c::bigint) <> 1 OR (c::bigint) = 1)
               ->  Result  (cost=0.00..431.13 rows=1 width=12)
                     ->  Table Scan on badestimate1  (cost=0.00..431.08 rows=3334 width=8)
 Optimizer status: PQO version 3.116.0
```
This histogram for c::int8 is not well defined, so while merging the
histograms due to other predicates on c::int8, we cannot create a
correctly defined histogram. But currently, we still merge 2 not well
defined histograms and generate a well defined histogram in
CHistogram::MakeUnionHistogramNormalize.
This commits adds the condition to not merge the histograms if the
histograms are not well defined.

Co-authored-by: Bhuvnesh Chaudhary <bchaudhary@vmware.com>
(cherry picked from commit 8e7b0fc7095902e4f02a06a4ee31a2d9f383dc1a)

No plan diff in tpc-ds pipeline: https://qpa.ci.gpdb.pivotal.io/teams/main/pipelines/ds_explain_5X/jobs/diff_explain_results_with_baseline/builds/117


khannaekta (commented):
We need to bump the ORCA version, else LGTM!

PR #626, opened 2021-07-23
Fix distribution spec used for CTE producer requirement

hardikar:
After optimizing with the "natural" distribution spec of a CTE
producer, we try to translate the query's distribution spec to
the column of the producer and do another round of optimization.

This could lead to incorrect results when the query had a distribution
spec (including equivalent distribution specs) that came from multiple
CTE consumers, with some of these columns being equivalent because
of "=" predicates between them.

For example:

   with cte as (select a,b from foo where b<10)
   select * from cte x1 join cte x2 on x1.a=x2.b

On the query side, columns x1.a and x2.b are equivalent, but we should
NOT treat columns a and b of the producer as equivalent.


l-wang (approved):

l-wang:
LGTM. It would be better if you could include the original back-ported commit number in the commit message to give us more context when look at this in the future.

hardikar:
@l-wang Thanks for the review. I updated the commit message.

PR #637, opened 2022-04-02
Execute on segment Function Scan JOIN External Scan

dgkimura:
Following query gave unexpected results:
  ```
  EXPLAIN SELECT * FROM gp_toolkit.__gp_localid, gp_toolkit.__gp_param_local_setting('gp_role');
                         QUERY PLAN
  -------------------------------------------------------------
   Nested Loop
     Join Filter: true
     ->  Function Scan on __gp_param_local_setting
     ->  Materialize
           ->  Gather Motion 3:1
                 ->  External Scan on __gp_localid
   Optimizer status: PQO version 3.122.0
  ```

Notice that the Function Scan runs on corredinator node. However, the
expectation is to run on the segment.

Let's consider following tree (r=requires, d=derives):

                 InnerNLJoin
                  /       \        .
 r:ANY           /         \   r:SINGLETON
 d:UNIVERSAL    /           \  d:EXTERNAL
               /             \     .
       FunctionScan        ExternalScan

On the outer side UNIVERSAL satisifies ANY, thus no enforcer is needed.
However, on the inner side EXTERNAL doesn't satisfy SINGLETON. But we
need to carefully consider how to implement the enforcer. If we apply a
GATHER MOTION, then the function scan may end up executing on the
coordinator node. Instead, we apply a BROADCAST MOTION to ensure that
the function scan executes on the segment nodes.


libgpopt/src/base/CDistributionSpecSingleton.cpp
```diff
@@ -120,21 +126,48 @@ CDistributionSpecSingleton::AppendEnforcers(CMemoryPool *mp,
...
-		COrderSpec *pos = prpp->Peo()->PosRequired();
-		pos->AddRef();
 		pexpr->AddRef();
-
-		CExpression *pexprGatherMerge = GPOS_NEW(mp) CExpression(
-			mp, GPOS_NEW(mp) CPhysicalMotionGather(mp, m_est, pos), pexpr);
-		pdrgpexpr->Append(pexprGatherMerge);
+		CExpression *pexprMotion = GPOS_NEW(mp)
+			CExpression(mp, GPOS_NEW(mp) CPhysicalMotionBroadcast(mp), pexpr);
+		pdrgpexpr->Append(pexprMotion);
```
l-wang, line 151: Do we need to also consider the following case inside this block?
```cpp
if (!prpp->Peo()->PosRequired()->IsEmpty() &&
			CDistributionSpecSingleton::EstMaster == m_est)
```

l-wang:
Awesome work! Nice diagram in the commit message and comment! There is a typo in the commit message `s/corredinator/coordinator/`. Can we also add a test for what you described?

l-wang (commented):
Awesome work! Nice diagram in the commit message and comment! There is a typo in the commit message s/corredinator/coordinator/. Can we also add a test for what you described?

